PR #1, opened 2021-01-26
Split SciMLBase from DiffEqBase

ChrisRackauckas:
There are many reasons for this. 

1. DiffEqBase.jl has grown to be a base between the differential equation solvers, with common tools for nonlinear solvers and callbacks adding a ton of dependencies. Users like @mauro3 noticed this in https://github.com/SciML/DiffEqBase.jl/issues/618
2. A lot of DiffEqBase was becoming unrelated to differential equations. OptimizationProblem, NonlinearProblem, all of those functions, etc. It's a full interface for SciML, not DiffEq which is by now a subset of the full ecosystem. But it needed to play the Base role, which meant that nonlinear problems were defined in the repo carrying all of the dependencies for the ODE solver's callback system.
3. It started inviting circular dependencies. Some of these problems are pervasive. We need to get the simplest form to GalacticOptim.jl, Quadrature.jl, NonlinearSolve.jl, etc. which are becoming the DiffEq's of their respective regimes (though there's not as much to do in those areas). But this meant NonlinearSolve.jl needed NonlinearProblem, but NonlinearSolve.jl was used in the callbacks, so... it had its own problem type. This gives an issue to a system like MTK which then has no unique type to target.
4. There's a lot of churn on the DiffEq solver portions of DiffEqBase, but relatively little on the actual interface portions. So it makes sense to let them be more stable.
5. There's a ton of packages which are @requires in DiffEqBase, which adds to the compile time. This was done because the library was supposed to avoid dependencies, but as the central piece of the differential equation solvers it really couldn't, and so it used required to fake lower dependencies but some of those like ForwardDiff.jl should really be promoted to being a real dependency now that it doesn't have the "idea" of being no-dependency.

For backwards compatibility, DiffEqBase reexports SciMLBase and imports a ton of the abstract types it used to have. This should allow for the migration to take place over time, while still fixing some of the MTK and NonlinearSolve issues that need a fix ASAP.

DiffEqBase will have a companion PR to match it.

There are some remaining issues here though:

1. I would like to get rid of the LabelledArrays dependency. Is there a good generic to use for the names? `nameof`? 
2. Is the split at the right level?
3. I want to drop the Queryverse stuff like TableTraits and IteratorInterfaceExtensions since we have Tables.jl
4. What should we do about our linear solver interface? Is there a way to fold it into these other problems? I think not, but there's a few things to think about.

Downstream changes to setup:

- [ ] GalacticOptim should use SciMLBase
- [ ] NonlinearSolve should use SciMLBase
- [ ] ModelingToolkit should use SciMLBase
- [ ] Quadrature should use SciMLBase


ChrisRackauckas:
Tested https://github.com/JuliaComputing/NonlinearSolve.jl/pull/25 works.

YingboMa (commented):
Could you move the some of the DiffEqBase tests here, too? It's probably pretty tricky to do, though.

ChrisRackauckas:
I plan to add a bunch of downstream tests.

YingboMa:
Yeah, makes sense. I like this a lot. This should cut down a lot of complication time. BTW, do we want to move OrdinaryDiffEq's generic dense output to DiffEqBase? Then other solvers and callbacks can use that interface.

devmotion:
I agree, it's great to separate the more basic non-DiffEq stuff. Is the plan also to move common functionality (e.g., generic interpolation, initialization, and keyword arguments) that appears in OrdinaryDiffEq and StochasticDiffEq to DiffEqBase then? I assume this would make it a bit easier to synchronize changes in OrdinaryDiffEq, StochasticDiffEq, DelayDiffEq etc. and maybe would allow us to drop the dependency on OrdinaryDiffEq in StochasticDiffEq.

Project.toml
```diff
@@ -3,7 +3,27 @@ uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
...
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+LabelledArrays = "2ee39098-c373-598a-b85f-a56591580800"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
+TreeViews = "a2a6695c-b41b-5b7d-aed9-dbfdeacea5d7"
```
devmotion, line 17: I was a bit surprised to see that the latest release is from July 2018 and the repo contains no Project.toml but only a REQUIRE file. But I assume it just indicates that the package is very stable and no changes are expected?

src/operators.jl
```diff
@@ -0,0 +1,55 @@
...
+1. Function call and multiplication: L(du, u, p, t) for inplace and du = L(u, p, t) for
+   out-of-place, meaning L*u and mul!(du, L, u).
+2. If the operator is not a constant, update it with (u,p,t). A mutating form, i.e.
+   update_coefficients!(A,u,p,t) that changes the internal coefficients, and a
+   out-of-place form B = update_coefficients(A,u,p,t).
+3. isconstant(A) trait for whether the operator is constant or not.
+4. islinear(A) trait for whether the operator is linear or not.
+=#
+
+Base.eltype(L::AbstractSciMLOperator{T}) where T = T
```
devmotion, line 13:
```suggestion
Base.eltype(::Type{AbstractSciMLOperator{T}}) where T = T
```

src/tabletraits.jl
```diff
@@ -0,0 +1,42 @@
...
+    types::Vector{Type}
+    lookup::Dict{Symbol, Int}
+    t::T
+    u::U
+end
+
+DESolutionRows(names, types, t, u) = DESolutionRows(names, types, Dict(nm => i for (i, nm) in enumerate(names)), t, u)
+
+Base.length(x::DESolutionRows) = length(x.u)
+Base.eltype(x::DESolutionRows{T, U}) where {T, U} = DESolutionRow{eltype(T), eltype(U)}
```
devmotion, line 15:
```suggestion
Base.eltype(::Type{DESolutionRows{T, U}}) where {T, U} = DESolutionRow{eltype(T), eltype(U)}
```

devmotion (commented):
Just noticed that we defined `eltype` for instances of `DESolutionRow` and `AbstractSciMLOperator` although the Julia docs recommend to define it for types.

devmotion:
> 1. I would like to get rid of the LabelledArrays dependency. Is there a good generic to use for the names? `nameof`?

It seems https://github.com/SciML/DiffEqBase.jl/blob/b5ac95fedb065af3b9aeb95a09eb2fa043b7424f/src/solutions/solution_interface.jl#L158 is equivalent to
```julia
    return keys(sol.u[1])
```
- so maybe use `keys`?

src/operators.jl
ChrisRackauckas: `eltype(zeros(2))`

src/tabletraits.jl
ChrisRackauckas: @quinnj does this work for the Tables.jl interface?

ChrisRackauckas:
>Did you check if the PRs improve the compile time of DiffEqBase?

No, that's a secondary concern here. Right now it's a big shuffle so I'm more concerned about correctness.

Project.toml
ChrisRackauckas: It's super stable. @pfizseb

src/operators.jl
devmotion: Sure, you can call `eltype` for instances since the generic fallback is `eltype(x) = eltype(typeof(x))`. "However the form that accepts a type argument should be defined for new types." (https://docs.julialang.org/en/v1/base/collections/#Base.eltype).

ChrisRackauckas:
>Is the plan also to move common functionality (e.g., generic interpolation, initialization, and keyword arguments) that appears in OrdinaryDiffEq and StochasticDiffEq (and to some extent also in DelayDiffEq and StochasticDelayDiffEq) to DiffEqBase then? I assume this would make it a bit easier to synchronize changes in OrdinaryDiffEq, StochasticDiffEq, DelayDiffEq etc. and maybe would allow us to drop the dependency on OrdinaryDiffEq in StochasticDiffEq.

Yes, precisely. And we'll want to PR to upstream packages telling them about SciMLBase.jl.

>BTW, do we want to move OrdinaryDiffEq's generic dense output to DiffEqBase?

Oh buddy, it's even worse than that:

https://github.com/SciML/DiffEqBase.jl/blob/master/src/interpolation.jl

That's the kind of thing I hope we can start cleaning up.

>so maybe use keys?

Yes that looks like it works.

src/operators.jl
ChrisRackauckas: Let's get an issue open on this one. I don't know if breaking this at the same time is a good idea. We'll need to make sure it's not breaking to DiffEqOperators.jl

src/tabletraits.jl
devmotion: The generic fallback ensures that there are no problems, regardless if you call it with types or instances, doesn't it?

ChrisRackauckas: I didn't know there's a generic fallback there.

src/operators.jl
YingboMa: This is not breaking though. It's a better definition.
```julia
julia> struct A end; Base.eltype(::A) = Float64

julia> struct B end; Base.eltype(::Type{B}) = Float64

julia> eltype(A())
Float64

julia> eltype(B())
Float64

julia> eltype(A)
Any

julia> eltype(B)
Float64
```

Project.toml
pfitzseb: Yep. I don't think there are any actively maintained consumers of TreeViews's tree views anymore though ;)

ChrisRackauckas: What do we define for VSCode and Juno nowadays then? TreeViews still seems to work for Juno?

pfitzseb: Yes. VSCode's tree views (in the workspace) aren't customizable atm.

src/tabletraits.jl
```diff
@@ -0,0 +1,42 @@
+Tables.isrowtable(::Type{<:DESolution}) = true
+Tables.columns(x::DESolution) = Tables.columntable(Tables.rows(x))
```
quinnj, line 2: Did we have this definition previously? At first glance, it doesn't seem necessary (the fallback in Tables.jl should do the same thing)

quinnj: Otherwise, all the Tables.jl stuff here looks great

ChrisRackauckas:
tests

Churning through those downstream tests 😃

ChrisRackauckas:
DelayDiffEq passed.

ChrisRackauckas:
LSODA passed.

ChrisRackauckas:
Sundials passed.

ChrisRackauckas:
DiffEqCallbacks passed.

ChrisRackauckas:
DiffEqJump passed

ChrisRackauckas:
DiffEqBase and ModelingToolkit passed

ChrisRackauckas:
DiffEqSensitivity passed

ChrisRackauckas:
StochasticDelayDiffEq passed

ChrisRackauckas:
DiffEqUncertainty passed

src/operators.jl
ChrisRackauckas: Had to revert because this caused a downstream failure in DiffEqOperators.jl. Can't say I know why, but this isn't the right time to look at that.

PR #6, opened 2021-01-28
Add helper functions for AbstractSciMLFunction

utkarsh530:
@ChrisRackauckas


src/scimlfunctions.jl
```diff
@@ -1110,6 +1110,18 @@ has_paramjac(f::AbstractDiffEqFunction) = __has_paramjac(f) && f.paramjac !== no
 has_syms(f::AbstractDiffEqFunction) = __has_syms(f) && f.syms !== nothing
 has_colorvec(f::AbstractDiffEqFunction) = __has_colorvec(f) && f.colorvec !== nothing
```
ChrisRackauckas, line 1111: delete these ones since AbstractDiffEqFunction <: AbstractSciMLFunction

devmotion:
I guess it would make sense to remove the definitions for `AbstractDiffEqFunction` above then since they would be covered as well?

PR #14, opened 2021-01-31
SecondOrderDDEProblem (and underlying DynamicalDDEProblem)

wi11dey:
Hey @ChrisRackauckas,

Following up on last semester, this PR has the plumbing necessary for `SecondOrderDDEProblem` that seems to be working well in Photron the last few weeks. It mirrors the structure of `SecondOrderODEProblem`, with the necessary changes in order to support a history function as well as 2 initial conditions. [DelayDiffEq.jl](https://github.com/SciML/DelayDiffEq.jl) requires a one-line change for this to work, because before it would only wrap the concrete `DDEFunction` rather than any `AbstractDDEFunction`: https://github.com/SciML/DelayDiffEq.jl/pull/203

Once Photron reaches stability, I can add its Lienard-Wiechert solver to [DiffEqProblemLibrary.jl/src/dde](https://github.com/SciML/DiffEqProblemLibrary.jl/tree/master/src/dde) as an example of how to use this problem to solve relativistic physics.

Companion PRs:
 - DelayDiffEq.jl: https://github.com/SciML/DelayDiffEq.jl/pull/203
 - Docs: https://github.com/SciML/DiffEqDocs.jl/pull/432

### DIFF
--- a/src/problems/dde_problems.jl
+++ b/src/problems/dde_problems.jl
@@ -1,7 +1,12 @@
 """
 $(TYPEDEF)
 """
-struct DDEProblem{uType,tType,lType,lType2,isinplace,P,F,H,K} <:
+struct StandardDDEProblem end
+
+"""
+$(TYPEDEF)
+"""
+struct DDEProblem{uType,tType,lType,lType2,isinplace,P,F,H,K,PT} <:
                           AbstractDDEProblem{uType,tType,lType,isinplace}
   f::F
   u0::uType
@@ -13,18 +18,20 @@ struct DDEProblem{uType,tType,lType,lType2,isinplace,P,F,H,K} <:
   kwargs::K
   neutral::Bool
   order_discontinuity_t0::Int
+  problem_type::PT
 
   @add_kwonly function DDEProblem{iip}(f::AbstractDDEFunction{iip}, u0, h, tspan, p=NullParameters();
                                        constant_lags = (),
                                        dependent_lags = (),
                                        neutral = f.mass_matrix !== I && det(f.mass_matrix) != 1,
                                        order_discontinuity_t0 = 0,
+                                       problem_type = StandardDDEProblem(),
                                        kwargs...) where {iip}
     _tspan = promote_tspan(tspan)
     new{typeof(u0),typeof(_tspan),typeof(constant_lags),typeof(dependent_lags),isinplace(f),
-        typeof(p),typeof(f),typeof(h),typeof(kwargs)}(
+        typeof(p),typeof(f),typeof(h),typeof(kwargs),typeof(problem_type)}(
           f, u0, h, _tspan, p, constant_lags, dependent_lags, kwargs, neutral,
-          order_discontinuity_t0)
+          order_discontinuity_t0, problem_type)
   end
 
   function DDEProblem{iip}(f::AbstractDDEFunction{iip}, h, tspan::Tuple, p=NullParameters();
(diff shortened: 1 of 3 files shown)


src/problems/dde_problems.jl
```diff
@@ -43,3 +50,112 @@ DDEProblem(f, args...; kwargs...) =
...
+function DynamicalDDEProblem(f::DynamicalDDEFunction,v0,u0,h::Function,tspan,p=NullParameters();dependent_lags=(),kwargs...)
+  DDEProblem(f,ArrayPartition(v0,u0),h,tspan,p;
+             problem_type=DynamicalDDEProblem{isinplace(f)}(),
+             dependent_lags=ntuple(i->(u,p,t)->dependent_lags[i](u[1],u[2],p,t),length(dependent_lags)),
+             kwargs...)
+end
+function DynamicalDDEProblem(f::DynamicalDDEFunction,h::Function,tspan,p=NullParameters();kwargs...)
+  DynamicalDDEProblem(f,h(p,first(tspan))...,h,tspan,p;kwargs...)
+end
+function DynamicalDDEProblem(f1::Function,f2::Function,args...;kwargs...)
```
wi11dey, line 80: Varargs are used here so the optional `v0`, `u0` initial conditions can be handled by dispatch by the two constructors above, and the code doesn't have to be repeated for any of the following constructors. `f1` and `f2`'s types are specified to be `Function` so that the constructor doesn't just pull out the first 2 arguments of and try to pass on the rest, which would result in more confusing dispatch errors down the line if an incorrect call is made.

src/scimlfunctions.jl
```diff
@@ -322,6 +343,21 @@ end
...
+function (f::DynamicalDDEFunction)(u,h,p,t)
+  ArrayPartition(f.f1(u.x[1],u.x[2],h,p,t),f.f2(u.x[1],u.x[2],h,p,t))
+end
+function (f::DynamicalDDEFunction)(du,u,h,p,t)
+  f.f1(du.x[1],u.x[1],u.x[2],h,p,t)
+  f.f2(du.x[2],u.x[1],u.x[2],h,p,t)
+end
+function Base.getproperty(f::DynamicalDDEFunction, name::Symbol)
+  if name === :f
+    # Use the f property as an alias for calling the function itself, so DynamicalDDEFunction fits the same interface as DDEFunction as expected by the ODEFunctionWrapper in DelayDiffEq.jl.
```
wi11dey, line 355: See https://github.com/SciML/DelayDiffEq.jl/pull/203: this is so that `ODEFunctionWrapper` can expect all `AbstractDDEFunction`s to define `f.f` with the entire DDE.

src/problems/dde_problems.jl
```diff
@@ -43,3 +50,112 @@ DDEProblem(f, args...; kwargs...) =
...
+struct DynamicalDDEProblem{iip} <: AbstractDynamicalDDEProblem end
+
+# u' = f1(v,h)
+# v' = f2(t,u,h)
+"""
+    DynamicalDDEProblem(f::DynamicalDDEFunction,v0,u0,tspan,p=NullParameters(),callback=CallbackSet())
+
+Define a dynamical DDE problem from a [`DynamicalDDEFunction`](@ref).
+"""
+function DynamicalDDEProblem(f::DynamicalDDEFunction,v0,u0,h::Function,tspan,p=NullParameters();dependent_lags=(),kwargs...)
```
devmotion, line 71: More generally:
```suggestion
function DynamicalDDEProblem(f::DynamicalDDEFunction,v0,u0,h,tspan,p=NullParameters();dependent_lags=(),kwargs...)
```

src/problems/dde_problems.jl
```diff
@@ -43,3 +50,112 @@ DDEProblem(f, args...; kwargs...) =
...
+  DDEProblem(f,ArrayPartition(v0,u0),h,tspan,p;
+             problem_type=DynamicalDDEProblem{isinplace(f)}(),
+             dependent_lags=ntuple(i->(u,p,t)->dependent_lags[i](u[1],u[2],p,t),length(dependent_lags)),
+             kwargs...)
+end
+function DynamicalDDEProblem(f::DynamicalDDEFunction,h::Function,tspan,p=NullParameters();kwargs...)
```
devmotion, line 77:
```suggestion
function DynamicalDDEProblem(f::DynamicalDDEFunction,h,tspan,p=NullParameters();kwargs...)
```

ChrisRackauckas (approved):
This looks good thanks.

PR #16, opened 2021-02-01
Add left & right params for Bracketing methods

utkarsh530:
I have kept all `u`, `left` & `right` for compatibility issues. I will set `u` to `left` in case of bracketing methods. Should I also set `left = u` implicitly?
Ref: https://github.com/JuliaComputing/NonlinearSolve.jl/pull/28
@ChrisRackauckas @YingboMa , please review.

### DIFF
--- a/src/solutions/nonlinear_solutions.jl
+++ b/src/solutions/nonlinear_solutions.jl
@@ -1,15 +1,15 @@
 """
 $(TYPEDEF)
 """
-struct NonlinearSolution{T,N,uType,R,P,A,O} <: AbstractNonlinearSolution{T,N}
+struct NonlinearSolution{T,N,uType,R,P,A,O,uType2} <: AbstractNonlinearSolution{T,N}
   u::uType
   resid::R
   prob::P
   alg::A
   retcode::Symbol
   original::O
-  left::uType
-  right::uType
+  left::uType2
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I don't think it makes sense to double up. We might as well have a canonical left or right choice, since it will be required that `u` is set to something anyways.

utkarsh530:
Yes, I was thinking of that. I guess `left = nothing` makes sense when we do not have bracketing methods.

src/solutions/nonlinear_solutions.jl
```diff
@@ -16,22 +18,24 @@ function build_solution(prob::AbstractNonlinearProblem,
...
 
   NonlinearSolution{T,N,typeof(u),typeof(resid),
-                    typeof(prob),typeof(alg),typeof(original)}(
-                    u,resid,prob,alg,retcode,original)
+                    typeof(prob),typeof(alg),typeof(original),typeof(left)}(
+                    u,resid,prob,alg,retcode,original,left,right)
 end
 
-function sensitivity_solution(sol::AbstractNonlinearProblem,u)
+function sensitivity_solution(sol::AbstractNonlinearSolution,u)
```
utkarsh530, line 33: Probably there was a typo here.

PR #17, opened 2021-02-03
fix N param for tuple in bracketing solution

utkarsh530:


src/solutions/nonlinear_solutions.jl
```diff
@@ -23,7 +23,7 @@ function build_solution(prob::AbstractNonlinearProblem,
                         kwargs...)
 
   T = eltype(eltype(u))
-  N = length((size(prob.u0)...,))
+  N = length((size(u)...,))
```
YingboMa, line 26:
```suggestion
  N = ndims(u)
```

ChrisRackauckas:
in olden times that didn't infer, but it likely does now.

PR #20, opened 2021-02-05
Update remake.jl

KirillZubov:


src/remake.jl
```diff
@@ -3,8 +3,8 @@
   Expr(:tuple, A...)
 end
 
-Base.@pure remaker_of(prob::T) where {T} = parameterless_type(T){isinplace(prob)}
-Base.@pure remaker_of(alg::T) where {T<:SciMLAlgorithm} = parameterless_type(T)
+Base.@pure remaker_of(prob::T) where {T<:SciMLProblem} = parameterless_type(T){isinplace(prob)}
```
YingboMa, line 6: These functions are not pure. Could be a generated function though

ChrisRackauckas: Is that a reason to be a blocker?

PR #25, opened 2021-02-07
Disable patch test for codecov

giordano:
This is mostly a test, Codecov settings are extremely finicky and documentation is often not incredibly clear


ChrisRackauckas:
that completely disabled it haha, which could be fine.

giordano:
I'll never understand Codecov settings....

PR #28, opened 2021-02-21
Start the soft deprecation of minimizer

ChrisRackauckas:
This allows for using the new OptimizationSolution as the underbelly for DiffEqFlux.jl in a non-breaking way. This is going to be a very "soft" change, as in `.minimizer` will still work without throwing an error, and in about a year we can add a depwarn. But we will change the documentation and tests almost immediately. This will let users slowly phase out the use of `.minimizer`, and we can throw a depwarn on code about a year. Naturally it should all be moving from `sciml_train` to direct GalacticOptim use over that time.


ChrisRackauckas:
@Vaibhavdixit02

Vaibhavdixit02:
Sounds good! 👍

PR #30, opened 2021-03-04
Use IO in Base.summary, respect :color setting

fonsp:
Resolves https://github.com/SciML/DifferentialEquations.jl/issues/728#issuecomment-789047602 by cleaning up the display methods.

# Problem

Currently, displaying an `ODEProblem` _in a display that does not support color_ looks glitchy, e.g.:
```
[36mODEProblem[0m with uType [36mArray{Int64,1}[0m and tType [36mFloat64[0m. In-place: [36mtrue[0m
timespan: (0.0, 100.0)
u0: [1, 0]
```

Here, `[36m` etc are "colorize" commands, `[36m` means "cyan from now on", when read by a terminal that supports color.

# Fix

Julia has a mechanism to solve such problems: `IOContext`. The `IO` (supertype of `IOContext`) argument to `Base.show`, `Base.summarize`, etc, contains extra information, including the `:color` flag, which the REPL sets to `true` if you are in a color terminal, and Pluto sets it to `false`, since it is not a (color) terminal.

The show methods in SciMLBase should check this `IOContext` property to switch colored output on or off. This is done by https://github.com/fonsp/SciMLBase.jl/blob/62c97e28276dfcfc398426198a15598cb566674e/src/utils.jl#L60 .

Many show methods in SciMLBase call `summary`, I have modified these calls to pass the `io` through to the `summary` call. I also changed the signatures of `Base.summary` overloads to take `io` as the first argument. This is the intended way to overload `Base.summary`, since the single-argument version is a fallback method built in to Julia. e.g.:

```julia
julia> struct Wow end

julia> Base.summary(io::IO, w::Wow) = println(io, "cool!")

julia> summary(Wow()) # (no io provided)
"cool!\n"
```

# Testing

Besides code review, I have only added one test, for an `ODEProblem`. To test fully test this PR, we should either:
- Add more tests like the one I wrote. This means constructing an instance for each type that has a show method defined, I wasn't sure how to do this easily.
- Use this branch of SciMLBase inside a REPL/Jupyter/Pluto session where all of these types occur, just to check that the show methods don't error.
- Be satisfied with code review.

FYI I changed the show methods for:
```
DEIntegrator

AbstractSciMLFunction

AbstractEnsembleSolution

DEProblem
AbstractLinearProblem
AbstractNonlinearProblem
AbstractOptimizationProblem
AbstractQuadratureProblem
AbstractNoiseProblem
AbstractEnsembleProblem
AbstractPDEProblem

AbstractOptimizationSolution

AbstractNoTimeSolution
```

### DIFF
--- a/src/ensemble/ensemble_solutions.jl
+++ b/src/ensemble/ensemble_solutions.jl
@@ -91,12 +91,13 @@ end
 
 ### Displays
 
-Base.summary(A::AbstractEnsembleSolution) = string("EnsembleSolution Solution of length ",length(A.u)," with uType:\n",eltype(A.u))
+Base.summary(io::IO, A::AbstractEnsembleSolution) =
+  print(io,"EnsembleSolution Solution of length ",length(A.u)," with uType:\n",eltype(A.u))
 function Base.show(io::IO, A::AbstractEnsembleSolution)
-  print(io,summary(A))
+  summary(io,A)
 end
 function Base.show(io::IO, m::MIME"text/plain", A::AbstractEnsembleSolution)
-  print(io,summary(A))
+  summary(io,A)
 end
 
 ### Plot Recipes
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -281,13 +281,16 @@ has_reinit(i::DEIntegrator) = false
 
 ### Display
 
-Base.summary(I::DEIntegrator) = string(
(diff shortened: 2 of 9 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Fantastic. I never know the show mechanism so it's good you figured it out! The new summary dispatch is great!

PR #32, opened 2021-03-10
fix mean and cov for integer solution values in Ensembles

isaacsas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/731 locally at least. CC @augustinas1

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -20,12 +20,13 @@ jobs:
           - {user: SciML, repo: DelayDiffEq.jl, group: All}
           - {user: SciML, repo: DiffEqBase.jl, group: All}
           - {user: SciML, repo: DiffEqFlux.jl, group: DiffEqFlux}
+          - {user: SciML, repo: DiffEqJump.jl, group: All}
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
add a test?

isaacsas:
Here or DiffEqJump?

ChrisRackauckas:
Either. Probably here, since I can see that `/1` being deleted by someone so there should be an immediate test that says no.

BTW, we should add DiffEqJump to the list of downstream tests.

isaacsas:
Ok. I’ll update with a test here and update the GitHub action tomorrow morning (away from my computer now).

ChrisRackauckas:
There's now a setup for easier downstream tests directly from SciMLBase for adding this kind of thing,.

isaacsas:
@ChrisRackauckas if tests pass this should be good.

isaacsas:
Decided to make it a bit more general since this is closer to the original issue. (i.e. integers actually worked fine in some cases, just not with broadcasting which wasn't used for flat array inputs like I previously had...)

PR #34, opened 2021-03-10
Remove `@show`

devmotion:
Was introduced in https://github.com/SciML/SciMLBase.jl/pull/33.


ChrisRackauckas:
yeah umm, who would do such a thing?

PR #38, opened 2021-03-20
Add `IncompleteProblemArgument`

YingboMa:
We can use

```julia
ODEProblem(fun, IncompleteProblemArgument(), (0, 10.0), IncompleteProblemArgument())
```

to explicitly hint to the libraries that incomplete initialization of a problem is intended.


ChrisRackauckas:
And upstream PRs? I don't see how using `nothing` isn't effectively the same thing.

PR #39, opened 2021-03-22
Add `observed` in `NonlinearFunction`

YingboMa:


ChrisRackauckas:
https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/solution_interface.jl#L81

I think you need `AbstractNoTimeSolution` dispatches too?

ChrisRackauckas (approved):

PR #40, opened 2021-03-26
Add 5-arg `mul!`

briochemc:
This PR simply adds a 5-arg `mul!` for `AbstractDiffEqLinearOperator`.

I'd like this functionality to be able to compose more things together and facilitate in place computations of sums of ODE-like functions and linear operators, i.e. something like `f(u) + L * u`:

```julia
f(du, u, p, t) # some in-place function writes du <- f(u)
mul!(du, L, u, 1, 1) # du <- L * u + f(u)
```

I'm not sure this is the right way to add this, please let me know if I need to think more about it! 😅 My current thinking is that I could use that, but maybe there's just no need for it?  🤷


ChrisRackauckas:
Yeah I think there might be some downstream piracy to delete?

ChrisRackauckas:
Can you rebase?

briochemc:
Done! (I think 😅)

PR #41, opened 2021-03-26
Remove fallback for `op(α::Number, L::Operator)`

briochemc:
This PR removes (unintended?) fallbacks for `op(α, L)` and `op(L, α)` where `op ∈ (:*, :/, :\)` when `α` is a `Number` and `L` an `AbstractDiffEqLinearOperator`. 

This is because, currently, DiffEqOperators overwrites (pirates) at least the `α * L` method to create a lazy `DiffEqScaledOperator`. My understanding from Slack discussions with Chris is that these operations with scalars should be removed here to avoid downstream piracy from DiffEqOperators which provides more efficient lazy operator constructions.

### DIFF
--- a/src/operators/common_defaults.jl
+++ b/src/operators/common_defaults.jl
@@ -10,9 +10,9 @@ LinearAlgebra.opnorm(L::AbstractDiffEqLinearOperator, p::Real=2) = opnorm(conver
 Base.@propagate_inbounds Base.getindex(L::AbstractDiffEqLinearOperator, I::Vararg{Any,N}) where {N} = convert(AbstractMatrix,L)[I...]
 Base.getindex(L::AbstractDiffEqLinearOperator, I::Vararg{Int, N}) where {N} =
   convert(AbstractMatrix,L)[I...]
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I think it would make sense to move the basics like ScaledOperator to the base libraries.

SebastianM-C:
In order to move `DiffEqScaledOperator` you would also need to move `AbstractDiffEqCompositeOperator` since `DiffEqScaledOperator <: AbstractDiffEqCompositeOperator`.
In this case, should other abstract operators be moved to `SciMLBase`?

I can make a PR with this if you want.

ChrisRackauckas:
That would be good.

SebastianM-C:
Should the operators be moved to `SciMLBase` or `DiffEqBase`?

ChrisRackauckas:
SciMLBase

briochemc:
Should I do anything here or should I let you guys take over?

ChrisRackauckas:
Looks like @SebastianM-C solved it, but there needs to be a downstream PR

Issue #42, opened 2021-03-28
Ambiguity error for `DynamicalDDEProblem`


SebastianM-C:
Running
```julia
using Test
using SciMLBase

problems = detect_ambiguities(SciMLBase)
```
gives
```julia
8-element Vector{Tuple{Method, Method}}:
 (DynamicalDDEProblem(f::DynamicalDDEFunction, h, tspan) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:77, DynamicalDDEProblem(f1::Function, f2::Function, args...; kwargs...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:80)
 (DynamicalDDEProblem(f::DynamicalDDEFunction, h, tspan, p; kwargs...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:77, DynamicalDDEProblem(f1::Function, f2::Function, args...; kwargs...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:80)
 ((var"#s255"::Core.var"#Type##kw")(::Any, ::Type{DynamicalDDEProblem}, f::DynamicalDDEFunction, h, tspan, p) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:77, (::Core.var"#Type##kw")(::Any, ::Type{DynamicalDDEProblem}, f1::Function, f2::Function, args...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:80)
 (DynamicalDDEProblem(f::DynamicalDDEFunction, v0, u0, h, tspan, p; dependent_lags, kwargs...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:71, DynamicalDDEProblem(f1::Function, f2::Function, args...; kwargs...) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\problems\dde_problems.jl:80)
[...]

PR #43, opened 2021-03-28
Fix ambiguities

SebastianM-C:
I removed the `::Function` restriction as mentioned in https://github.com/SciML/SciMLBase.jl/issues/42#issuecomment-808923055

### DIFF
--- a/src/problems/dde_problems.jl
+++ b/src/problems/dde_problems.jl
@@ -77,7 +77,7 @@ end
 function DynamicalDDEProblem(f::DynamicalDDEFunction,h,tspan,p=NullParameters();kwargs...)
   DynamicalDDEProblem(f,h(p,first(tspan))...,h,tspan,p;kwargs...)
 end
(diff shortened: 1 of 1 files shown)


devmotion:
Maybe you could add a test of `detect_ambiguities` to avoid regressions and to ensure that no new ambiguities are added?

SebastianM-C:
There already is one (it's the first test actually)
https://github.com/SciML/SciMLBase.jl/blob/15df15622631a58a7dc943515a3902e65859cb6c/test/runtests.jl#L6

devmotion:
Oh OK, I didn't notice this, somehow I just assumed all tests pass on the master branch :smile:

devmotion (approved):
LGTM :+1:

isaacsas:
FWIW the DiffEqJump downstream failure is a dumb test bug (there is apparently a RNG-dependent outcome for a test...).

SebastianM-C:
@isaacsas Thanks, I was not sure what was the problem with that.

PR #44, opened 2021-03-29
Move `DiffEqScaledOperator`

SebastianM-C:
As a continuation from https://github.com/SciML/SciMLBase.jl/pull/41#issuecomment-808957928

This moves the following from DiffEqOperators
- DiffEqScaledOperator
- AbstractDiffEqCompositeOperator (since it was required by subtyping)
- update_coefficients!
- getops (since it was needed for other functions)
- Base and LinearAlgebra overloads for DiffEqScaledOperator

This should be tested together with https://github.com/SciML/DiffEqOperators.jl/pull/363

### DIFF
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -444,6 +444,11 @@ abstract type AbstractDiffEqLinearOperator{T} <: AbstractDiffEqOperator{T} end
 
 """
 $(TYPEDEF)
+"""
+abstract type AbstractDiffEqCompositeOperator{T} <: AbstractDiffEqLinearOperator{T} end
+
+"""
+$(TYPEDEF)
 
 Base for types defining SciML functions.
 """
@@ -488,9 +493,9 @@ include("scimlfunctions.jl")
 include("alg_traits.jl")
 
 include("operators/operators.jl")
+include("operators/basic_operators.jl")
 include("operators/diffeq_operator.jl")
 include("operators/common_defaults.jl")
-include("operators/basic_operators.jl")
 
 include("problems/problem_utils.jl")
 include("problems/discrete_problems.jl")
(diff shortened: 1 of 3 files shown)


SebastianM-C:
There seems to be an ambiguity after the move with `ldiv!`
```julia
problems = detect_ambiguities(SciMLBase)
1-element Vector{Tuple{Method, Method}}:
 (ldiv!(Y::AbstractVecOrMat{T} where T, L::SciMLBase.AbstractDiffEqLinearOperator, B::AbstractVecOrMat{T} where T) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\operators\common_defaults.jl:19, ldiv!(Y::AbstractArray, L::DiffEqScaledOperator, B::AbstractArray) in SciMLBase at C:\Users\sebastian\.julia\dev\SciMLBase\src\operators\diffeq_operator.jl:116)
```
In the case of the default fallback we have
```julia
LinearAlgebra.ldiv!(Y::AbstractVecOrMat, L::AbstractDiffEqLinearOperator, B::AbstractVecOrMat) =
  ldiv!(Y, convert(AbstractMatrix,L), B)
```
while for the specialized operator we have
```julia
LinearAlgebra.ldiv!(Y::AbstractArray, L::DiffEqScaledOperator, B::AbstractArray) =
  lmul!(1/L.coeff, ldiv!(Y, L.op, B))
```

The problem comes from the fact that the default fallback is more specific on the `Y` and `B` arguments than the specialized `ldiv!` on `DiffEqScaledOperator`.

Should I restrict the definition in the specialized method to `AbstractVecOrMat` or should the generic fallback accept `AbstractArray`s?

SebastianM-C:
The `OrdinaryDiffEq` integration test fails due to the removal of the `op(α::Number, L::Operator)` fallback.
https://github.com/SciML/SciMLBase.jl/pull/44/checks?check_run_id=2223664193#step:6:660

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L506-L507

This should probably be explicitly concretizing

SebastianM-C:
@ChrisRackauckas I added the default fallback for `DiffEqArrayOperator` and removed the `ldiv!` fallback that caused the ambiguities. Locally tests seem to pass, but I'm not sure if removing the `ldiv!` fallback doesn't cause any downstream problems (CI should tell us about that).

PR #45, opened 2021-03-31
Change map for a for loop

andreasnoack:
to work around https://github.com/JuliaLang/julia/issues/35800

which can cause very slow compile times in Julia 1.6. In Pumas, this change alone had the following effect on a single function call:

### Before:
```julia
40.194758 seconds (83.73 M allocations: 6.623 GiB, 2.70% gc time, 99.99% compilation time)
```

### After
```julia
2.916390 seconds (7.31 M allocations: 446.180 MiB, 3.71% gc time, 99.96% compilation time)
```

I'm erroring out in the empty case since it avoids the problem of determining the return type. I'd think that it's expected that the number of trajectories is positive. I actually think that part of the problem with `map` is that it has to consider the empty case but that is mostly speculation.

### DIFF
--- a/src/ensemble/basic_ensemble_solve.jl
+++ b/src/ensemble/basic_ensemble_solve.jl
@@ -187,6 +187,7 @@ function solve_batch(prob, alg, ::EnsembleSerial, II, pmap_batch_size; kwargs...
     throw(ArgumentError("number of trajectories must be positive"))
   end
   batch_data = [batch_func(first(II), prob, alg; kwargs...)]
+  sizehint!(batch_data,length(II))
   for i in 2:length(II)
     @inbounds push!(batch_data, batch_func(II[i], prob, alg; kwargs...))
   end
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Seriously 😱?

andreasnoack:
Yes. It's really sad.

korsbo:
Ickie ickie! You're even growing an array and it's still much faster - before 1.6 I would have been on the hunt to do the exact opposite changes to my code 😢

src/ensemble/basic_ensemble_solve.jl
```diff
@@ -182,11 +182,15 @@ function solve_batch(prob,alg,ensemblealg::EnsembleDistributed,II,pmap_batch_siz
...
 
-function solve_batch(prob,alg,::EnsembleSerial,II,pmap_batch_size;kwargs...)
-  batch_data = map(II) do i
-    batch_func(i,prob,alg;kwargs...)
+function solve_batch(prob, alg, ::EnsembleSerial, II, pmap_batch_size; kwargs...)
+  if isempty(II)
+    throw(ArgumentError("number of trajectories must be positive"))
   end
-  tighten_container_eltype(batch_data)
+  batch_data = [batch_func(first(II), prob, alg; kwargs...)]
```
ChrisRackauckas, line 189: should we add a `sizehint!` here?

andreasnoack: Yes. I'll add that.

ChrisRackauckas (approved):

PR #66, opened 2021-06-18
Add symbol based indexing for interpolated solutions

sharanry:
This PR requires https://github.com/SciML/RecursiveArrayTools.jl/pull/143 and does the following:

- [x] Fills extra fields of DiffEqArray offered by https://github.com/SciML/RecursiveArrayTools.jl/pull/143 after interpolating to enable symbol based indexing of eliminated and reduced states.
- [x] Add tests for the same.


devmotion:
I always thought at some point one could unify the standard interpolations in SciMLBase and OrdinaryDiffEq etc. Maybe the additional arguments might make this more difficult? I guess it doesn't matter though since there is no push in this direction currently AFAIK.

ChrisRackauckas:
The lazy interpolations of the Verner methods make this a bit hard. Keeping the SciMLBase ones light makes sense IMO since the OrdinaryDiffEq.jl ones do a lot of very special stuff that nothing else needs.

devmotion:
Yes, I had only the Hermite and linear interpolations in mind.

ChrisRackauckas:
That would make `dense=false` a type unstable change because it would change the underlying type for the interpolation, instead of just checking for the bool and doing the linear behavior. Constant prop may handle it, but 🤷

test/downstream/observables.jl
```diff
@@ -53,3 +56,25 @@ plot(sol,vars=(lorenz2.x,α))
...
+@test sol[x] isa Vector
+@test sol[x,2] isa Float64
+@test sol[x,:] isa Vector
+@test length(sol[x,1:5]) == 5
+@test sol[RHS] isa Vector
+@test sol[RHS,3] isa Float64
+@test length(sol[RHS,5:10]) == 6
+
+interpolated_sol = sol(0.0:1.0:10.0)
+@test interpolated_sol[RHS] ≈ (1 .- interpolated_sol[x])./3.0
```
sharanry, line 80: The previous Lorenz system-based example found in `master` fails for some reason outside of tests in `structural_simplify` step. I have created an issue #68 .

For now, I have replaced it with a simpler example that can be used for all old and new checks.

src/solutions/solution_interface.jl
```diff
@@ -1,7 +1,3 @@
-### Abstract Interface
-struct AllObserved
-end
```
ChrisRackauckas, line 3: What's the replacement for this? And this would be technically breaking.

ChrisRackauckas: Oh, moved to https://github.com/SciML/RecursiveArrayTools.jl/pull/143 ?

sharanry: https://github.com/SciML/RecursiveArrayTools.jl/pull/143 moves this upstream to  https://github.com/SciML/RecursiveArrayTools.jl/blob/5f02cb8c603c0d854cbcf8d148720ea503383a8d/src/vector_of_array.jl#L19. This is needed for `issymbollike(x)` function which is in turn needed for symbol resolution in `Base.getindex(...)`. 

(This is reexported by SciMLBase.jl)

ChrisRackauckas: Yeah, we really need a SymbolicsBase.jl which just defines a few functions and types to use instead of that `issymbollike(x)` hack. @YingboMa is now the time?

test/downstream/observables.jl
sharanry: https://github.com/SciML/SciMLBase.jl/issues/68

src/solutions/solution_interface.jl
YingboMa: We can use https://github.com/JuliaSymbolics/TermInterface.jl

src/utils.jl
```diff
@@ -72,10 +72,6 @@ end
 
 using Base: typename
 
-Base.@pure __parameterless_type(T) = typename(T).wrapper
-parameterless_type(x) = parameterless_type(typeof(x))
-parameterless_type(x::Type) = __parameterless_type(x)
-
```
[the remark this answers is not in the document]
YingboMa: Maybe just don't move those functions?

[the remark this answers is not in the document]
sharanry: This is needed for `issymbollike(x)` function which is in turn needed for symbol resolution in `Base.getindex(::DiffEqArray, ...)` https://github.com/SciML/SciMLBase.jl/pull/66#discussion_r655297501 .

Are you suggesting keeping a copy in both packages? Wouldn't that would create issues?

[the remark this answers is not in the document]
YingboMa: It won't if it's not exported.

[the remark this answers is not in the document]
sharanry: But is it advisable to create two types with the same name (`AllObserved`) in different but related packages?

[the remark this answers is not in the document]
YingboMa: We don't need AllObserved in SciMLBase. Just keep `parameterless_type` here.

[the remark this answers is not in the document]
sharanry: But that would still be breaking?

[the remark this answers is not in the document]
YingboMa: No one is using AllObserved, so that's fine.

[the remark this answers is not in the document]
YingboMa: Reminder to release RecursiveArrayTools.jl first, and export AllObserved in SciML and add an appropriate bound on RecursiveArrayTools.

ChrisRackauckas:
Seems like most tests are failing?

test/downstream/observables.jl
```diff
@@ -45,6 +46,68 @@ sol = solve(prob,Rodas4())
...
+@test sol9.u isa Vector
+@test first(sol9.u) isa Real
+@test length(sol9.u) == 11
+@test length(sol9.t) == 11
+
+sol10 = sol(0.1, idxs=2)
+@test sol6 isa Real
+
+@test_throws ErrorException sol(0.1, Val{1})
+
```
sharanry, line 110: @YingboMa All these scenarios should work now. CC: @ChrisRackauckas

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
   destats::DE
   retcode::Symbol
 end
-(sol::ODESolution)(t,deriv::Type=Val{0};idxs=nothing,continuity=:left) = sol.interp(t,idxs,deriv,sol.prob.p,continuity)
+
+function (sol::ODESolution)(t,deriv::Type=Val{0};idxs=nothing,continuity=:left)
+  if deriv != Val{0}
```
devmotion, line 20: Why do we need to restrict it to `Val{0}`? Isn't this just passed through to the calls of `sol.interp`?

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
+    end
+  elseif idxs isa Vector{<:Integer} || idxs isa Integer
+      sol.interp(t,idxs,deriv,sol.prob.p,continuity)
+  elseif issymbollike(idxs)
+    if t isa Real
+      interp_sol = augment(sol.interp([t],nothing,deriv,sol.prob.p,continuity), sol)
+      interp_sol[idxs][1]
+    else
+      interp_sol = augment(sol.interp(t,nothing,deriv,sol.prob.p,continuity), sol)
+      @info interp_sol[idxs]
```
devmotion, line 38:
```suggestion
```

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
   retcode::Symbol
 end
-(sol::ODESolution)(t,deriv::Type=Val{0};idxs=nothing,continuity=:left) = sol.interp(t,idxs,deriv,sol.prob.p,continuity)
+
+function (sol::ODESolution)(t,deriv::Type=Val{0};idxs=nothing,continuity=:left)
+  if deriv != Val{0}
+    error("Not Implemented")
+  end
+
+  if idxs === nothing
```
devmotion, line 24: It seems it would be easier to extend if we would dispatch on `idxs` and `t`.

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
+    if t isa Real
+      sol.interp(t,idxs,deriv,sol.prob.p,continuity)
+    else
+      augment(sol.interp(t,idxs,deriv,sol.prob.p,continuity), sol)
+    end
+  elseif idxs isa Vector{<:Integer} || idxs isa Integer
```
devmotion, line 30: Maybe better
```suggestion
  elseif idxs isa AbstractVector{<:Integer} || idxs isa Integer
```
?

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
+      sol.interp(t,idxs,deriv,sol.prob.p,continuity)
+  elseif issymbollike(idxs)
+    if t isa Real
+      interp_sol = augment(sol.interp([t],nothing,deriv,sol.prob.p,continuity), sol)
```
devmotion, line 34: Can't we use

```suggestion
      interp_sol = augment(sol.interp(t,nothing,deriv,sol.prob.p,continuity), sol)
```

src/solutions/ode_solutions.jl
```diff
@@ -15,7 +15,40 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
+      interp_sol[idxs][1]
+    else
+      interp_sol = augment(sol.interp(t,nothing,deriv,sol.prob.p,continuity), sol)
+      @info interp_sol[idxs]
+      DiffEqArray(interp_sol[idxs], t)
+    end
+  else
+    if t isa Real
+      interp_sol = augment(sol.interp([t],nothing,deriv,sol.prob.p,continuity), sol)
```
devmotion, line 43: Same here, can't we use

```suggestion
      interp_sol = augment(sol.interp(t,nothing,deriv,sol.prob.p,continuity), sol)
```

src/solutions/ode_solutions.jl
sharanry: AFAIK, higher-order interpolation is currently not supported yet. https://github.com/SciML/ModelingToolkit.jl/issues/1014 This is just an effort to to indicate to the user that higher-order interpolation is not implemented yet. @YingboMa suggested that we could implement a ForwardDiff based higher-order interpolation -- but it might warrant a separate PR as it doesn't seem straightforward.

src/solutions/ode_solutions.jl
sharanry: Yes. That seems more practical. Thanks! Will change it.

src/solutions/ode_solutions.jl
sharanry: I guess I will make it more clear with -- "Higher-order interpolation is not implemented."

devmotion: IMO it is more correct and easier to fix if it fails when calling `sol.interp(....)` in case it is missing. Alternatively, if one really wants to exclude this case on a higher level, it would be more idiomatic to just restrict the function signature to `Val{0}` instead of throwing a custom error.

src/solutions/ode_solutions.jl
sharanry: We will have to add an extra constructor to `DiffEqArray` in RecursiveArrayTools.jl to support this. https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/vector_of_array.jl#L42-L46 doesn't cover the case where `A.u` is Vector{Real}.

src/solutions/ode_solutions.jl
```diff
@@ -15,9 +15,68 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
...
+  interp_sol = augment(sol.interp(t,nothing,Val{0},sol.prob.p,continuity), sol)
+  u = [[interp_sol[idx][i] for idx in idxs] for i in 1:length(t)]
+  DiffEqArray(u, t)
+end
+
+for T in 1:3
+  function (sol::ODESolution)(t,::Type{Val{T}},idxs,continuity)
+    error("Higher-order interpolation is not implemented.")
+  end
+end
```
sharanry, line 78: Is this approach okay to handle higher-order interpolation? @devmotion

devmotion: Just don't define these functions?

sharanry: I guess the idea was to have a user-readable error. @YingboMa Any thoughts?

src/solutions/ode_solutions.jl
sharanry: Will be added in https://github.com/SciML/RecursiveArrayTools.jl/pull/148

src/solutions/ode_solutions.jl
YingboMa: How about
```julia
function (sol::ODESolution)(t::AbstractVector{<:Real},::Type{Val{N}},idxs::AbstractVector,continuity) where N
    N == 0 || error("Higher-order interpolation is not implemented.")
```

src/solutions/ode_solutions.jl
sharanry: The current API of interpolation outputs `Vector{<:Real}` when the input `t isa Real`. DiffEqArray expects a vector of times `t` (`Vector{<:Real}`)  and vector of corresponding states `u` (`Vector{Vector{<:Real}}`). 

These two factors together make it difficult to accommodate what you are suggesting without making substantial changes to interpolation and DiffEqArray indexing.

src/solutions/ode_solutions.jl
sharanry: @YingboMa I think erroring out at `sol(t, idxs=[...])` stage for higher-order interpolation is wrong (as @devmotion already mentioned). I just realized that it has been already implemented in many scenarios.

https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L274-L314
https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L335-L369
https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L387
https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L414
https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L439
https://github.com/SciML/SciMLBase.jl/blob/master/src/interpolation.jl#L465

sharanry: I have set a default at `interpolant` stage https://github.com/SciML/SciMLBase.jl/pull/66/commits/b97346ced028c6113082c0cb83ae493047c43e6a .

sharanry:
Weird. The downstream tests seem to pass locally.

ChrisRackauckas:
https://github.com/SciML/SciMLBase.jl/pull/66/checks?check_run_id=2961581350#step:6:881 that one is slightly stochastic (we need to make it take more samples), so that's fine and would almost certainly pass if re-tested.

https://github.com/SciML/SciMLBase.jl/pull/66/checks?check_run_id=2961581076 is worrisome: do your symbolic versions match?

sharanry:
@ChrisRackauckas I updated it to v1.2.1 and reconfirmed SciMLBase's downstream tests. Could you restart just that particular workflow?

sharanry:
Okay, it errored again. I will take a closer look.

sharanry:
@YingboMa @ChrisRackauckas It looks like [`issymbollike()`](https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/vector_of_array.jl#L22)  is erroneously returning false on `Symbolics.Num` type object only in a sandboxed downstream test environment.  It correctly returns true when I normally import ModellingToolkit/RecursiveArrayTools. Any idea what could possibly lead in such behavior?

ChrisRackauckas:
We need to just get rid of that hack. @YingboMa @shashi can we just add `istree` to https://github.com/JuliaSymbolics/TermInterface.jl/blob/master/src/TermInterface.jl and depend on it here for this?

shashi:
We are renaming istree i today isterm so yes it’s already in TermInterface. We just need to use it in SU :p

ChrisRackauckas:
Okay yes, so try and prioritize that and we'll change this from the hacky thing to some traits and proper dispatches.

sharanry:
@ChrisRackauckas The tests for this PR will probably pass once https://github.com/SciML/RecursiveArrayTools.jl/pull/152 is merged and released.

ChrisRackauckas:
Shit that was a mistake. Reverting.

PR #67, opened 2021-06-18
Begin documenting the common interface at a high level

ChrisRackauckas:


ChrisRackauckas:
@frankschae I'll pull you in for the differentiation section later. First I want to get something merged quickly so we can all collab on different parts (and open bounties)

frankschae:
sounds good.  Is this in addition to existing docs or is it intended to replace them?

ChrisRackauckas:
I think just in addition. The current docs are all differential equation focused. My plan is that every equation package (differential equations, nonlinear systems, optimization, etc.) is documented in such a way that they are complete: a user feels fine using that package in isolation (so NonlinearSolve.jl needs help). But we need to also document at a high level exactly how things relate between the packages. It's more from a design point of view to explain it all to both deep users and potential developers. For example, what is the relationship between `sensealg` on differential equations and nonlinear solves? That doesn't have a place, but it's important to define what exactly a `sensealg` is, what it does, that it's universal, and how the arguments should always be defined.

PR #72, opened 2021-06-29
fix serial ensembles for jumps with prob_funcs

isaacsas:
Seems to fix the serial case that was resetting the seed for jump problems in https://github.com/SciML/DiffEqJump.jl/issues/184, while multithreading still works as before.

I'm not sure if this will work properly for distributed or GPU problems (but I don't know if what we have now seeds properly in those cases). 

Happy to modify if there is a better solution.

### DIFF
--- a/src/ensemble/basic_ensemble_solve.jl
+++ b/src/ensemble/basic_ensemble_solve.jl
@@ -135,13 +135,12 @@ function __solve(prob::AbstractEnsembleProblem,
 
 end
 
-function batch_func(i,prob,alg;kwargs...)
+function batch_func(i,prob,alg,ensemblealg;kwargs...)
   iter = 1
 
-  # if no user specified safetycopy and one thread we don't copy, otherwise
-  # we copy when given a user specified prob_func
+  # if running in serial or on one thread we disable safetycopy as a default
   if prob.safetycopy === nothing
-    safetycopy = (Threads.nthreads() == 1) ? false : DEFAULT_SAFETYCOPY(prob.prob_func)
+    safetycopy = ((Threads.nthreads() == 1) || (ensemblealg isa EnsembleSerial)) ? false : DEFAULT_SAFETYCOPY(prob.prob_func)
   else 
     safetycopy = prob.safetycopy
   end
@@ -185,7 +184,7 @@ function solve_batch(prob,alg,ensemblealg::EnsembleDistributed,II,pmap_batch_siz
   =#
 
   batch_data = pmap(wp,II,batch_size=pmap_batch_size) do i
-    batch_func(i,prob,alg;kwargs...)
+    batch_func(i,prob,alg,ensemblealg;kwargs...)
   end
 
   tighten_container_eltype(batch_data)
@@ -199,9 +198,9 @@ function responsible_map(f,II...)
   batch_data
 end
 
-function SciMLBase.solve_batch(prob,alg,::EnsembleSerial,II,pmap_batch_size;kwargs...)
+function SciMLBase.solve_batch(prob,alg,ensemblealg::EnsembleSerial,II,pmap_batch_size;kwargs...)
   batch_data = responsible_map(II) do i
-    SciMLBase.batch_func(i,prob,alg;kwargs...)
+    SciMLBase.batch_func(i,prob,alg,ensemblealg;kwargs...)
   end
   SciMLBase.tighten_container_eltype(batch_data)
 end
@@ -219,7 +218,7 @@ function solve_batch(prob,alg,ensemblealg::EnsembleThreads,II,pmap_batch_size;kw
   end
   
   batch_data = tmap(II) do i
-    batch_func(i,prob,alg;kwargs...)
+    batch_func(i,prob,alg,ensemblealg;kwargs...)
   end
   tighten_container_eltype(batch_data)
 end
--- a/src/ensemble/basic_ensemble_solve.jl
+++ b/src/ensemble/basic_ensemble_solve.jl
@@ -137,7 +137,16 @@ end
 
 function batch_func(i,prob,alg;kwargs...)
   iter = 1
-  _prob = prob.safetycopy ? deepcopy(prob.prob) : prob.prob
+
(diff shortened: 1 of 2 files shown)


src/ensemble/basic_ensemble_solve.jl
```diff
@@ -137,7 +137,16 @@ end
 
 function batch_func(i,prob,alg;kwargs...)
   iter = 1
-  _prob = prob.safetycopy ? deepcopy(prob.prob) : prob.prob
+
+  # if no user specified safetycopy and one thread we don't copy, otherwise
+  # we copy when given a user specified prob_func
+  if prob.safetycopy === nothing
+    safetycopy = (Threads.nthreads() == 1) ? false : DEFAULT_SAFETYCOPY(prob.prob_func)
```
devmotion, line 144: Maybe it would be cleaner to use
```suggestion
_safetycopy = if safetycopy === nothing
    Threads.nthreads() == 1 ? false : DEFAULT_SAFETYCOPY(prob.prob_func)
else
    safetycopy
end
```
in the keyword argument constructors of EnsembleProblem and use `safetycopy = nothing` as default keyword argument (as done in this PR)? Then one would not have to change the type of `safetycopy::Bool` to `safetycopy::Union{Nothing,Bool}`.

isaacsas: Yeah, though I wonder if I should be checking if it would be better to just check if the user is using EnsembleSerial? Even if the number of threads is two, if they are using the serial algorithm we'd still want this fix right? In that case we can't make the decision at the time of constructing the JumpProblem based on `nthreads` since the choice of what algorithm to use occurs during the call to solve.

ChrisRackauckas: Interesting. Is there a case where that is safer? Since you cannot change numbers of threads while Julia is running, I'm not sure if it makes a difference?

ChrisRackauckas: EnsembleSerial already won't safetycopy IIRC.

isaacsas: No it does if `prob_func` is not the default. That was @TorkelE problem.

isaacsas: `safetycopy` is (currently) set before the choice of EnsembleAlg type.

isaacsas:
Yeah, so this doesn't work if the user selects EnsembleSerial when running Julia with more than one thread. Checking if the ensemblealg is EnsembleSerial should fix that case. I can change to that, but I think that would require keeping the type union in the EnsembleProblem. Alternatively, we can just decide that if you are running Julia multithreaded you need to use a threaded ensemble algorithm (which is the current default I think). Thoughts?

src/ensemble/basic_ensemble_solve.jl
devmotion: Shouldn't this be fixed in the implementation of EnsembleSerial (i.e., shouldn't it just ignore `safetycopy`)? It seems this is orthogonal to the default value of `safetycopy` of EnsembleProblem.

isaacsas: Could there be contexts where a user might want to always force a copy even in serial, so we wouldn't want to override their default? If we want to say that it is only used with threading though that is fine and I can update to handle both cases.

ChrisRackauckas: Yes. If a user has callbacks which mutates the parameters, you need to copy or the second solve will not match the firsr

src/ensemble/basic_ensemble_solve.jl
```diff
@@ -135,13 +135,12 @@ function __solve(prob::AbstractEnsembleProblem,
 
 end
 
-function batch_func(i,prob,alg;kwargs...)
+function batch_func(i,prob,alg,ensemblealg;kwargs...)
```
ChrisRackauckas, line 138: I think we may have to update DiffEqGPU that is using this internal.

isaacsas: Would making it a kwarg avoid that?

ChrisRackauckas:
It's fine, just make sure to do the downstream

isaacsas:
OK, I'll add a PR there later tonight.

ChrisRackauckas:
>Yes. If a user has callbacks which mutates the parameters, you need to copy or the second solve will not match the firsr

Yeah this worries me because it means it's unsafe. If you do `remake` to change `u0`, and have a callback modifying parameters, then your code is now broken with this while it was safe before. I am now thinking this might not be the right solution. It might be better to ask why that case was failing for you, since it seems to work on my computer with threads > 1, so something might be happening in the threads == 1 branch of EnsembleThreads. 

That said, yesterday it came up that safetycopy is too strong, as in, look at the big slack thread showing how it slowed down a case by 3x where it wasn't even needed. I think we need to do a few things like:

- Define `copy` for problem types, and use a cheap `copy` instead of `deepcopy` (which is super expensive).
- Have a better solve time heuristic. Parameters only need to be made safe if there are callbacks, so check for callbacks. And check if the `prob` that comes out has `newprob.u0 === prob.u0 && newprob.p === prob.p`: only copy what you need to via `remake` and `copy` on the pieces.

the latter is more effort but might make the heuristic both safe and usually no cost.

isaacsas:
You really don’t want to copy at all for pure JumpProblems when running on one thread if possible, especially in the most common use case of just running many simulations serially with the same parameters. (But it shouldn’t be needed in the serial case of just changing a parameter or initial condition either.) 

I don’t know the non-jump code-bases as well, but for jumps, if you are calling `remake` in `prob_func` your already making an aliasing copy, so that should be getting handled right?

The specific issue here, I think, was the continual safety copying of the original problem was resetting the rng state each time is the serial case. The multithreading case worked fine as it goes through a separate code-path in SSAStepper that avoids aliasing at the expense of (potentially expensive) copies each time.

ChrisRackauckas:
It would be the same with JumpProblems, ignoring the jump callback.

PR #80, opened 2021-07-07
Fix issymbollike special case not covered by downstream tests

sharanry:
Match `issymbollike`  to [RecursiveArrayTools's implementation](https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/vector_of_array.jl#L22). 
Just spotted this bug. Didn't trigger any downstream tests in #66 unexpectedly.


ChrisRackauckas:
We should just delete this one and alias it.

Issue #83, opened 2021-07-09
potential logic error in solution interface


lamorton:
It looks to me like [this](https://github.com/SciML/SciMLBase.jl/blob/9ac2da300e34bb84b4167aa3250d5aa6fe55876b/src/solutions/solution_interface.jl#L52-L53) is a mistake:
```julia
  if i == nothing
      if issymbollike(i) && Symbol(i) == getindepsym(A)
```

If `i` is `nothing` then it makes no sense to test whether it is symbol-like or is equal to the independent variable -- I think it's probably meant to be `sym` here instead of `i`.

ChrisRackauckas:
@sharanry

sharanry:
This might be a wider issue with RecursiveArrayTools too. https://github.com/SciML/RecursiveArrayTools.jl/blob/58d89af893261490dde31d1faf042b89b2270289/src/vector_of_array.jl#L75

Will look into it.

sharanry:
I think we can close this? @ChrisRackauckas

PR #84, opened 2021-07-12
Remove old unreachable code

sharanry:
https://github.com/SciML/SciMLBase.jl/issues/83

@ChrisRackauckas

### DIFF
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -50,7 +50,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
   end
 
     indepsym = getindepsym(A)
-    if i === nothing     
+    if i == nothing     
       if issymbollike(sym) && indepsym !== nothing && Symbol(sym) == indepsym
           A.t
       else
@@ -69,7 +69,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
     end
 
     indepsym = getindepsym(A)
-    if i === nothing
+    if i == nothing
         if issymbollike(sym) && indepsym !== nothing && Symbol(sym) == indepsym
             A.t[args...]
         else
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -49,12 +49,8 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
       i = sym
   end
 
-  if i == nothing
-      if issymbollike(i) && Symbol(i) == getindepsym(A)
-        A.t
-      else
-        observed(A,sym,:)
-      end
+  if i === nothing     
+      observed(A,sym,:)
   else
       A[i,:]
   end
@@ -67,12 +63,8 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
         i = sym
     end
 
-    if i == nothing
-      if issymbollike(i) && Symbol(i) == getindepsym(A)
-        A.t[args...]
-      else
+    if i === nothing
         observed(A,sym,args...)
-      end
     else
         A[i,args...]
     end
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -50,7 +50,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
   end
 
     indepsym = getindepsym(A)
-    if i == nothing     
+    if i === nothing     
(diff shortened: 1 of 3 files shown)


src/solutions/solution_interface.jl
```diff
@@ -49,11 +49,12 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
       i = sym
   end
 
-  if i == nothing
-      if issymbollike(i) && Symbol(i) == getindepsym(A)
-        A.t
+  indepsym = getindepsym(A)
```
ChrisRackauckas, line 52:
```suggestion
    indepsym = getindepsym(A)
```

src/solutions/solution_interface.jl
```diff
@@ -49,11 +49,12 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution,sy
       i = sym
   end
 
-  if i == nothing
-      if issymbollike(i) && Symbol(i) == getindepsym(A)
-        A.t
+    indepsym = getindepsym(A)
+    if i === nothing
```
ChrisRackauckas, line 53: why `===`? I don't think this could be inferred.

sharanry: I have used `===` whenever I compare against `nothing`. My understanding is that it does a type comparison in case of `nothing` -- am I correct to assume what you say implies that this only helps when the type of `i` is inferrable leads to the if the statement being compiled away? Does it cause any overhead if the type of `i` is not inferrable?

ChrisRackauckas: It does a type-comparison, yes, which can be helpful for branch elimination. But `i` here is not known to be `nothing`, so we don't want to compare types I would think.

sharanry: https://docs.julialang.org/en/v1/manual/performance-tips/#Checking-for-equality-with-a-singleton This what I was intending to say.

sharanry:
The downstream tests for this probably require https://github.com/SciML/RecursiveArrayTools.jl/pull/157.

src/solutions/solution_interface.jl
sharanry: image
I think the julia extension has also led to me subconsciously  always using `===` with `nothing`. 😅

sharanry: I have replace is it with normal euality.

YingboMa: It's always recommended to use `===` to compare with nothing.

YingboMa:
Don't replace `===` with `==`.

ChrisRackauckas:
It needed to merge to general first. 🎖️

ChrisRackauckas:
Tests failing.

sharanry:
Looking into it.

sharanry:
@ChrisRackauckas I think the leftover errors are unrelated?

YingboMa:
Do you mind to squash some commits? Other than that, I think we are good to merge.

sharanry:
@YingboMa I don't really have permissions for "squash and merge". Could you just do that?

YingboMa:
Also, I think we shouldn't convert symbolic variables to `Symbol` to compare. It should just use `isequal`. `Symbol` is both terribly slow and prone to errors (i.e. many terms cannot be converted to `Symbol`s).

sharanry:
> Also, I think we shouldn't convert symbolic variables to Symbol to compare. It should just use isequal. Symbol is both terribly slow and prone to errors (i.e. many terms cannot be converted to Symbols).

I am not sure what exactly you are referring to. Are you talking about `issymbollike`?

YingboMa:
I am referring to patterns like `Symbol(sym) == indepsym` and `sym_to_index`.

YingboMa:
> I don't really have permissions for "squash and merge". Could you just do that?

I meant to squash the `===` -> `==` -> `===` commits. For those minor changes, I often just use `git commit --amend`, so the commits don't become messy.

sharanry:
> I am referring to patterns like Symbol(sym) == indepsym and sym_to_index.

But isn't this required for Term to Symbol conversion? Are you suggesting overloading `isqual` or `==` for this?

ChrisRackauckas:
Yingbo is saying we shouldn't be doing that, and yes we shouldn't. It was a hack from ParameterizedFunctions.jl times, so some of that code existed since... 2016(?) In various forms. With a real symbolic engine that has metadata, we shouldn't be using symbol checking and should change it all to using `isequal` directly: that is more correct with MTK, and it would be safe for symbols for backwards compatibility.

YingboMa:
And of course, ideally, `syms` should be a `Dict{Any,Int}` that maps symbolic variables to indices, so the look-up is O(1) or sublinear in practice.

ChrisRackauckas:
We need to fix those two downstream tests...

PR #85, opened 2021-07-18
Zygote rules for solution_interface

lamorton:
Trying to solve [DifferentialEquations#746](https://github.com/SciML/DifferentialEquations.jl/issues/746). This is work in progress. At the moment, there are method ambiguities introduced by the lack of type specificity for the Zygote rules. In testing, I had been using `Num` from `Symbolics.jl` to specialize the adjoints, but inside `SciMLBase` that causes a circular dependency, and it's not completely general. I'm not sure if a Holy trait would work with Zygote here. 

Also, for `observed` quantities, I haven't figured out how to implement the adjoint for `sol[sym]` yet, but the adjoint of `sol[sym,i]` works for an individual time slice.


lamorton:
I'm puzzled by the fact that some of the method ambiguities are for the underlying `getindex` function itself, rather than for the adjoint.  For instance, the first ambiguity in the list is:
```julia
getindex(x::AbstractArray{T, N}, mask::FillArrays.Ones{Bool, N, Tuple{Vararg{Base.OneTo{Int64}, N}}}) where {T, N} in FillArrays at /home/runner/.julia/packages/FillArrays/5CbY2/src/trues.jl:39
```
vs.
```julia
getindex(A::SciMLBase.AbstractNoTimeSolution, sym) in SciMLBase at /home/runner/work/SciMLBase.jl/SciMLBase.jl/src/solutions/solution_interface.jl:95
```

ChrisRackauckas:
@shahsi @YingboMa @sharanry is there a way to not leave these getindexes completely open?

sharanry:
@ChrisRackauckas 
I think the circular dependency @lamorton is talking about and the open getindex can be easily fixed for good if we have access to all the Terms from symbolics through a base symbolics dependency instead of the current `issymbollike` hack.

As for fixing the ambiguity, for now, dispatching `getindex(A::SciMLBase.AbstractNoTimeSolution, mask::FillArrays.Ones)` should do the trick? This is however helpful only until we come across the next such ambiguity.

YingboMa:
Can we add those methods in Symbolics rather than in SciMLBase?

ChrisRackauckas:
Yes, let's move Symbolic to TermInterface.jl and just dispatch on that. @shashi

shashi:
The right way to check this is `Symbolics.unwrap(x) isa Symbolic`.... This is what I would suggest if you're adding the dependency on Symbolics.jl here

TermInterface deals with traits only, there are no types in there.

shashi:
We didn't plan on having abstract types in TermInterface.... That would be confusing. Maybe we can make a SymbolicsBase with `Symbolic`, `wrap` and `unwrap`.

cc @0x0f0f0f

lamorton:
I've found that using the scalar indexing rule results in very slow loss function evaluations, probably b/c the multitude of copies of the system that are being created. In fact, it's faster to use a gradient-free solver.

ChrisRackauckas:
That's a general Zygote thing. It's really unoptimized for that kind of scalar stuff. We should be using the Zygote.OneHot here.

Project.toml
```diff
@@ -18,6 +18,8 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 TreeViews = "a2a6695c-b41b-5b7d-aed9-dbfdeacea5d7"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
ChrisRackauckas, line 21: Zygote cannot be a dependency here.

src/solutions/zygote.jl
```diff
@@ -0,0 +1,40 @@
+@adjoint function getindex(VA::ODESolution, sym, j::Int)
```
ChrisRackauckas, line 1: Define these as ChainRules v1.0 rules?

ChrisRackauckas:
Seems to have gone stale.

PR #93, opened 2021-07-26
Add residual to the SciMLBase.ODESolution.

NAThompson:
Still leaves a bit to be desired on the count of elegance, and tests still need to go in, but maybe we can run the CI and at least see if I've broke the universe.

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -78,6 +78,40 @@ function interpolation_residual(sol::AbstractODESolution, t)
   sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t)
 end
 
+function max_residual_estimate(sol::AbstractODESolution)
+  max_res = 0
+  for i in 1:length(sol.t)
+    tᵢ = sol.t[i]
+    uᵢ = sol.u[i]
+    abs_res = abs(sol(tᵢ,Val{1}) - sol.prob.f(uᵢ, sol.prob.p, tᵢ))
+    if abs_res > max_res
+      max_res = abs_res
+    end
+  end
+
+  for i in 1:length(sol.t)-1
+    tᵢ = sol.t[i]
+    tᵢ₊₁ = sol.t[i+1]
+    Δᵢ = tᵢ₊₁ - tᵢ
+    t = tᵢ + Δᵢ/4
+    abs_res = abs(sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t))
+    if abs_res > max_res
+      max_res = abs_res
+    end
+    t = tᵢ + Δᵢ/2
+    abs_res = abs(sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t))
+    if abs_res > max_res
+      max_res = abs_res
+    end
+    t = tᵢ + 3Δᵢ/4
+    abs_res = abs(sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t))
+    if abs_res > max_res
+      max_res = abs_res
+    end
+  end
+  max_res
+end
+
 function build_solution(
         prob::Union{AbstractODEProblem,AbstractDDEProblem},
         alg,t,u;timeseries_errors=length(u)>2,
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -14,6 +14,7 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
   tslocation::Int
   destats::DE
   retcode::Symbol
+  residual::Function
 end
 (sol::ODESolution)(t,deriv::Type=Val{0};idxs=nothing,continuity=:left) = sol(t,deriv,idxs,continuity)
 (sol::ODESolution)(v,t,deriv::Type=Val{0};idxs=nothing,continuity=:left) = sol.interp(v,t,idxs,deriv,sol.prob.p,continuity)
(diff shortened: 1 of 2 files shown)


src/solutions/ode_solutions.jl
```diff
@@ -14,6 +14,7 @@ struct ODESolution{T,N,uType,uType2,DType,tType,rateType,P,A,IType,DE} <: Abstra
   tslocation::Int
   destats::DE
   retcode::Symbol
+  residual::Function
```
ChrisRackauckas, line 17: instead, it would probably better to add a `interpolation_residual(sol::ODESolution,t)` function.

NAThompson: That the diff you were thinking of?

ChrisRackauckas:
Yeah, and with this I'm not quite sure it really needs to be in the package if it simplifies just down to this 😅 . I guess the question is, where does it go from here? Is there a bunch of cool uncertainty quantification stuff you have planned? Plot recipes or something? If it's just one function it's not worth it, but if this then ties into other functionality then I will see it.

NAThompson:
Yeah, I get where you're coming from.

I had envisioned adding a `destat` which printed ‖u̇-f‖∞ along with all the other stats, so it would've been a maximization over the residual, which is . . . slightly less trivial?

In some sense adding this would be a popularization of the idea that you don't need the analytic solution to decide if your solve was good; the residual is measurable, and *should* be measured as a QC check after every solve. I think it would certainly aid the unit test suite, as it would allow more extensive testing of problems without analytic solutions.

ChrisRackauckas:
>I had envisioned adding a destat which printed ‖u̇-f‖∞ along with all the other stats, so it would've been a maximization over the residual, which is . . . slightly less trivial?

That is interesting. What the plan for cheaply getting the maximum?

ChrisRackauckas:
>For chaotic problems, forward error cannot be controlled, so you could ask the user to print the backward error in the bug report.

That is an interesting point.

NAThompson:
> That is interesting. What the plan for cheaply getting the maximum?

1 or 2 digits of accuracy would be fine right? That could be done by sampling (say) all the nodes, and 2 or 3 nodes in between. Let's estimate 50ns for evaluating the interpolator, 20 nodes, 3 samples in between each node, giving us about 3 microseconds.

src/solutions/ode_solutions.jl
```diff
@@ -78,6 +78,40 @@ function interpolation_residual(sol::AbstractODESolution, t)
   sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t)
 end
 
+function max_residual_estimate(sol::AbstractODESolution)
```
ChrisRackauckas, line 81: it should probably check `sol.dense` first?

NAThompson: Just read the [docs](https://diffeq.sciml.ai/stable/tutorials/ode_example/#Step-3:-Analyzing-the-Solution) for the dense output. It appears that if `dense=false` then we have linear interpolation, which should give a very poor backward error, moreover the derivatives at the nodes will be undefined. So I guess what you're getting at is that we shouldn't check the nodes if `dense=false`?

ChrisRackauckas: I guess it could just do it, but it won't be the most accurate error estimate in that case.

NAThompson: So there are some philosophical issues here. . . IMO the *interpolator is the numerical solution*. Some people think the solution skeleton ((t_i, u_i) pairs) is the solution.

If you adopt my viewpoint, then maybe you'd say we need to do a symmetric derivative at the node? If you adopt the solution skeleton viewpoint, then you might say "only compute the residual between the nodes".

NAThompson: Corless gives a definition that might help here:

> Definition 12.1. If the computed solution (typically a piecewise polynomial interpolant) is continuously differentiable, we will call the residual a defect. If the computed solution is only piecewise continuously differentiable, so that the residual is only defined piecewise and may have jump discontinuities, we will call it a deviation.

src/solutions/ode_solutions.jl
```diff
@@ -74,6 +74,44 @@ function (sol::ODESolution)(t::AbstractVector{<:Real},deriv,idxs::AbstractVector
   DiffEqArray([[interp_sol[idx][i] for idx in idxs] for i in 1:length(t)], t, idxs, getindepsym(sol), observed, p)
 end
 
+function interpolation_residual(sol::AbstractODESolution, t)
+  sol(t,Val{1}) - sol.prob.f(sol(t), sol.prob.p, t)
```
NAThompson, line 78: Will this syntax work for the "inplace" syntax exemplified by the [Lorenz ODE example](https://diffeq.sciml.ai/stable/tutorials/ode_example/#Example-2:-Solving-Systems-of-Equations)?

ChrisRackauckas: no, you'll want to check whether `f` isinplace and then write into a cache vector via `similar(u)`.

NAThompson:
Ok, managed to successfully export the `max_residual_estimate`; should this be tested in another repo?

ChrisRackauckas:
>Quick note: The backward error from the Lorenz example is a bit larger than I expected:

Interesting.

>but wouldn't it be easier to just:

This definitely isn't something to compute by default though, given that it adds `f` evaluations and not everyone would access it.

I just realized that the right place for this to go would be DiffEqUncertainty.jl so that we can document it next to the other UQ tooling. Would you mind migrating it?

NAThompson:
> I just realized that the right place for this to go would be DiffEqUncertainty.jl so that we can document it next to the other UQ tooling. Would you mind migrating it?

I wouldn't mind at all.

PR #103, opened 2021-09-07
add missing Any in constructor

jd-lara:
The constructor with no recompile was missing a couple of `Any` to match the number of fields in the struct.


ChrisRackauckas:
@YingboMa @shashi looks like something in Symbolics caused a downstream breakage.

https://github.com/SciML/SciMLBase.jl/pull/103/checks?check_run_id=3537092699#step:6:490

PR #113, opened 2021-10-29
Add the ability to remake an `AbstractEnsembleProblem`

dannys4:
This should enable the user to modify either an `AbstractEnsembleProblem` or the problem stored therein. There could be an issue if one tries to modify a field in the inner problem but the keyword is also in the `AbstractEnsembleProblem`, but I'll leave that up to your discretion if that's a real issue.


chriselrod:
Should be fixed.

PR #118, opened 2021-11-13
Fix tests

YingboMa:


YingboMa:
I cannot reproduce the test failure locally. Might have to wait the upgrade rollout.

PR #128, opened 2021-12-24
Allow direct setting of kwargs in remake

ChrisRackauckas:


src/remake.jl
```diff
@@ -38,7 +42,8 @@ function remake(prob::ODEProblem; f=missing,
                                   u0=missing,
                                   tspan=missing,
                                   p=missing,
-                                  kwargs...)
+                                  kwargs=missing,
```
YingboMa, line 45: How is `kwargs` used?

ChrisRackauckas: See https://github.com/SciML/DiffEqSensitivity.jl/pull/529. The problem is that if we don't have a `kwargs` in remake, we cannot force `prob.kwargs` to be something that we want. Instead, it always merges with the previous kwargs, so there's no way to use remake to delete things from `kwargs`.

YingboMa: Ohhh, I see.

YingboMa (approved):

PR #133, opened 2022-01-28
Prevent inference pessimization

Keno:
The `let` block here, together with the constprop heuristics drops the type from `Const(true)` to `Bool`,
with progressively worse type information resulting from there. The result is that inference ends up
inferring a bunch of unnecessary code, which slows down total inference time. We saw about a 20%
inference time improvement for ODE initialization with this change.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -2158,9 +2158,7 @@ for S in [
         Base.convert(::Type{$S}, x::$S) = x
         Base.convert(::Type{$S{iip}}, x::T) where {T<:$S{iip}} where iip = x
         function ConstructionBase.constructorof(::Type{<:$S{iip}}) where iip
-            let IIP = iip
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
@chriselrod looks like you have a current release failure? https://github.com/SciML/SciMLBase.jl/runs/4974672446?check_suite_focus=true#step:6:593

chriselrod:
> https://github.com/SciML/SciMLBase.jl/runs/4974672446?check_suite_focus=true#step:6:593

That error comes from Static.jl 0.5 replacing `nothing` with `missing`.
No released version of LoopVectorization is listed as compatible with Static 0.5.

ChrisRackauckas:
Interesting CI is allowing it.

chriselrod:
Pkg's resolver works in mysterious ways.

PR #134, opened 2022-01-28
Add the median to EnsembleSummary and adjust the plot recipe

### LINKED ISSUE #132: Add median to `EnsembleSummary`
Right now, EnsembleSummary contains the mean, variance, and quantiles. The corresponding plot recipe plots (by default) the mean, and the quantiles as ribbons. This can lead to unexpected plots for cases with strong outliers, where the line lies outside of the ribbon.

**Proposal:** Add the median to the ensemble summary. Then plot either the median (line) and the quantiles (ribbon) for `ci_type=:quantile` or the mean and variance as currently done when `ci_type=:SEM`.

nathanaelbosch:
Fixes #132

The following script can be used for a before/after comparison:
```julia
using OrdinaryDiffEq, Plots

logistic(du, u, p, t) = @. du = p * u * (1 - u)
prob = ODEProblem(logistic, [1e-2], (0.0, 2.5), [3.0])
prob_func(prob, i, repeat) = i < 100 ? remake(prob, u0=[0.9]) : remake(prob, u0=[1e-2 + 0.001*randn()])
eprob = EnsembleProblem(prob, prob_func=prob_func)
esol = solve(eprob, Tsit5(), trajectories=1000)
esum = EnsembleSummary(esol, 0.0:0.01:2.5, quantiles=[0.1, 0.9])
plot(esum)
```
Now, the plot recipe plots the median and quantiles and the resulting plot becomes more easily interpretable.

I did not add any new tests or adjust version numbers, so please let me know what should be done in these regards.

### DIFF
--- a/src/ensemble/ensemble_solutions.jl
+++ b/src/ensemble/ensemble_solutions.jl
@@ -41,9 +41,9 @@ struct EnsembleSummary{T,N,Tt,S,S2,S3,S4,S5} <: AbstractEnsembleSolution{T,N,S}
   t::Tt
   u::S
   v::S2
-  med::S5
-  qlow::S3
-  qhigh::S4
+  med::S3
+  qlow::S4
+  qhigh::S5
   num_monte::Int
   elapsedTime::Float64
   converged::Bool
--- a/src/ensemble/ensemble_solutions.jl
+++ b/src/ensemble/ensemble_solutions.jl
@@ -37,10 +37,11 @@ end
 """
 $(TYPEDEF)
 """
-struct EnsembleSummary{T,N,Tt,S,S2,S3,S4} <: AbstractEnsembleSolution{T,N,S}
+struct EnsembleSummary{T,N,Tt,S,S2,S3,S4,S5} <: AbstractEnsembleSolution{T,N,S}
   t::Tt
   u::S
   v::S2
+  med::S5
   qlow::S3
   qhigh::S4
   num_monte::Int
@@ -120,12 +121,12 @@ end
 @recipe function f(sim::EnsembleSummary;
                    idxs= typeof(sim.u[1])<:AbstractArray ? eachindex(sim.u[1]) : 1,
                    error_style=:ribbon,ci_type=:quantile)
-  if typeof(sim.u[1])<:AbstractArray
-    u = vecarr_to_vectors(sim.u)
-  else
-    u = [sim.u.u]
-  end
   if ci_type == :SEM
+    if typeof(sim.u[1])<:AbstractArray
+        u = vecarr_to_vectors(sim.u)
+    else
+        u = [sim.u.u]
+    end
     if typeof(sim.u[1])<:AbstractArray
       ci_low = vecarr_to_vectors(VectorOfArray([sqrt.(sim.v[i]/sim.num_monte).*1.96 for i in 1:length(sim.v)]))
       ci_high = ci_low
@@ -134,6 +135,11 @@ end
       ci_high = ci_low
     end
   elseif ci_type == :quantile
+    if typeof(sim.med[1])<:AbstractArray
+        u = vecarr_to_vectors(sim.med)
+    else
+        u = [sim.med.u]
+    end
     if typeof(sim.u[1])<:AbstractArray
       ci_low = u - vecarr_to_vectors(sim.qlow)
       ci_high = vecarr_to_vectors(sim.qhigh) - u
--- a/src/ensemble/ensemble_analysis.jl
+++ b/src/ensemble/ensemble_analysis.jl
@@ -108,16 +108,18 @@ function SciMLBase.EnsembleSummary(sim::SciMLBase.AbstractEnsembleSolution{T,N},
                          t=sim[1].t;quantiles=[0.05,0.95]) where {T,N}
   if typeof(sim[1]) <: SciMLSolution
     m,v = timeseries_point_meanvar(sim,t)
+    med = timeseries_point_median(sim,t)
     qlow = timeseries_point_quantile(sim,quantiles[1],t)
     qhigh = timeseries_point_quantile(sim,quantiles[2],t)
   else
     m,v = timeseries_steps_meanvar(sim)
+    med = timeseries_steps_median(sim)
     qlow = timeseries_steps_quantile(sim,quantiles[1])
     qhigh = timeseries_steps_quantile(sim,quantiles[2])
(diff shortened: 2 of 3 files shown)


src/ensemble/ensemble_solutions.jl
```diff
@@ -37,10 +37,11 @@ end
 """
 $(TYPEDEF)
 """
-struct EnsembleSummary{T,N,Tt,S,S2,S3,S4} <: AbstractEnsembleSolution{T,N,S}
+struct EnsembleSummary{T,N,Tt,S,S2,S3,S4,S5} <: AbstractEnsembleSolution{T,N,S}
   t::Tt
   u::S
   v::S2
+  med::S5
```
nathanaelbosch, line 44: This was the minimal change but the result is a bit weird. Maybe some types should be reordered, but I was not 100% sure if anything else might rely on the current type signature of EnsembleSummary.

ChrisRackauckas: Nothing else should rely on it.

ChrisRackauckas:
Add a test for the median call?

nathanaelbosch:
> Add a test for the median call?
I'm not quite sure what test you're suggesting. Checking that `timestep_median` and `timepoint_median` return the correct values for a simple example? Which would probably mean adding an actual EnsembleProblem solve to the tests?

ChrisRackauckas:
@chriselrod Looks like LabelledArrays conversions may have led to a downstream issue?

chriselrod:
Another LoopVectorization + ArrayInterface 4 here: https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:56
@KristofferC any idea what could be going on here, or how to diagnose, why we seemingly keep getting ArrayInterface 4 + LoopVectorization, even though LoopVectorization's compat bounds exclude ArrayInterface 4? https://github.com/JuliaSIMD/LoopVectorization.jl/blob/b0ba0370926a7288a270b4a0a19ab2990be83c1c/Project.toml#L28
This has been causing test failures for weeks; LoopVectorization is incompatible with ArrayInterface 4, and using them together will result in the error messages listed there.

I'm not sure if it's a Pkg bug, because we've never really gotten a reproducer/had someone share a problematic environment. Whenever anyone reports the error, as soon as they add one of the packages to report the status, Pkg re-resolves, installs a compatible version, and then the problem goes away.

chriselrod:
@ChrisRackauckas, I guess you were talking about this
```julia
LabelledArrays Tests: Error During Test at /home/runner/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  LoadError: type Array has no field x
  Stacktrace:
    [1] getproperty(x::Vector{Float64}, f::Symbol)
      @ Base ./Base.jl:33
    [2] f1(du::Vector{Float64}, u::LabelledArrays.LArray{Float64, 1, Vector{Float64}, (x = 1:4, y = 5:8)}, p::Vector{Float64}, t::Float64)
      @ Main.##282 ~/work/SciMLBase.jl/SciMLBase.jl/downstream/test/downstream/labelledarrays.jl:20
    [3] (::SciMLBase.ODEFunction{true, typeof(Main.##282.f1), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing})(::Vector{Float64}, ::Vararg{Any, N} where N)
      @ SciMLBase ~/work/SciMLBase.jl/SciMLBase.jl/src/scimlfunctions.jl:334
```
`du` is a `Vector` and `u` is a `LabelledArray`.

The PR added a method:
```julia
julia> using LabelledArrays

julia> A = @LArray [1,2,3] (:a,:b,:c)
3-element LArray{Int64, 1, Vector{Int64}, (:a, :b, :c)}:
 :a => 1
 :b => 2
 :c => 3

julia> convert(AbstractVector{Float64}, A)
3-element LArray{Float64, 1, Vector{Float64}, (:a, :b, :c)}:
 :a => 1.0
 :b => 2.0
 :c => 3.0
```
Even though this is in `LabelledArray`s tests, perhaps `u` was supposed to be a vector?
But it'd take more digging to find out what went wrong. I may be able to take a look tomorrow.

KristofferC:
> Another LoopVectorization + ArrayInterface 4 here: https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:56 

Not really. In the status output here: https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:36 there is no LoopVectorization present which means that ArrayInterface is free to be whatever it wants. It is only after the test start running and the test specific dependencies are installed that LoopVectorization is present: https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:336 but then ArrayInterface is downgraded: https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:265.

So at no point do you have both LoopVectorization and ArrayInterface v4 in a project.



In the link I don't see LoopVectorization being a part of the project at all so therefore it's compat won't be in effect.

chriselrod:
> So at no point do you have both LoopVectorization and ArrayInterface v4 in a project.

The error LoopVectorization throws is one we'd get if it were hypothetically paired with ArrayInterface 4, but not if it were paired with ArrayInterface 3.
ArrayInterface 3 returned `nothing`s where ArrayInterface 4 returns `missing`.

It is possible that another package is guilty of having made the switch from returning `nothing` to returning `missing`, while allowing compatibility with both 3 or 4.

KristofferC:
https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:525

> 4 dependencies precompiled but different versions are currently loaded. Restart julia to access the new versions
[591](https://github.com/SciML/SciMLBase.jl/runs/5083436845?check_suite_focus=true#step:6:591)

So, it seems that one version of ArrayInterface is loaded already. But `Pkg.test` should make a new process so I don't see why that would matter...

KristofferC:
I guess it is this stuff: https://github.com/SciML/SciMLBase.jl/blob/176e987ff8bea5d6e1acbf3dd3e92a3e242e5434/test/runtests.jl#L12-L16. You probably need to do that in a separate process.

PR #139, opened 2022-02-21
Add the `DynamicalODEProblem` problem_type

nathanaelbosch:
All other problems in this file set the `problem_type` of the resulting `ODEProblem`, e.g. the SecondorderODEProblem here: 
https://github.com/SciML/SciMLBase.jl/blob/8f709a06f208bf24900a93e5384bab95fade6b2d/src/problems/ode_problems.jl#L154-L155

It therefore seemed to me that this should also be the case for `DynamicalODEProblem`s.


ChrisRackauckas:
Yes thanks, this should have the trait.

ChrisRackauckas (approved):

PR #148, opened 2022-04-18
adjoint

vpuri3:
https://github.com/SciML/LinearSolve.jl/issues/92

### DIFF
--- a/src/operators/basic_operators.jl
+++ b/src/operators/basic_operators.jl
@@ -118,9 +118,9 @@ Like DiffEqArrayOperator, but stores a Factorization instead.
 
 Supports left division and `ldiv!` when applied to an array.
 """
-struct FactorizedDiffEqArrayOperator{T<:Number, FType <: Union{Factorization{T},
-                                                               Diagonal{T},
-                                                               Bidiagonal{T},
+struct FactorizedDiffEqArrayOperator{T<:Number, FType <: Union{
+                                                               Factorization{T}, Diagonal{T}, Bidiagonal{T},
+                                                               Adjoint{T,<:Factorization{T}},
                                                               }
                                     } <: AbstractDiffEqLinearOperator{T}
   F::FType
@@ -133,6 +133,12 @@ LinearAlgebra.ldiv!(Y::AbstractVecOrMat, L::FactorizedDiffEqArrayOperator, B::Ab
 LinearAlgebra.ldiv!(L::FactorizedDiffEqArrayOperator, B::AbstractVecOrMat) = ldiv!(L.F, B)
 Base.:\(L::FactorizedDiffEqArrayOperator, x::AbstractVecOrMat) = L.F \ x
 LinearAlgebra.issuccess(L::FactorizedDiffEqArrayOperator) = issuccess(L.F)
+Base.adjoint(L::FactorizedDiffEqArrayOperator) = FactorizedDiffEqArrayOperator(L.F')
+
+LinearAlgebra.ldiv!(y, L::FactorizedDiffEqArrayOperator, x) = ldiv!(y, L.F, x)
+#isconstant(::FactorizedDiffEqArrayOperator) = true
+has_ldiv(::FactorizedDiffEqArrayOperator) = true
+has_ldiv!(::FactorizedDiffEqArrayOperator) = true
 
 # The (u,p,t) and (du,u,p,t) interface
 for T in [DiffEqScalar, DiffEqArrayOperator, FactorizedDiffEqArrayOperator, DiffEqIdentity]
--- a/src/operators/basic_operators.jl
+++ b/src/operators/basic_operators.jl
@@ -126,14 +126,26 @@ struct FactorizedDiffEqArrayOperator{T<:Number, FType <: Union{
   F::FType
 end
 
-Base.Matrix(L::FactorizedDiffEqArrayOperator) = Matrix(L.F)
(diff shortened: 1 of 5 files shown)


src/operators/basic_operators.jl
```diff
@@ -89,6 +89,7 @@ end
 update_coefficients!(L::DiffEqArrayOperator,u,p,t) = (L.update_func(L.A,u,p,t); L)
 isconstant(L::DiffEqArrayOperator) = L.update_func == DEFAULT_UPDATE_FUNC
 Base.similar(L::DiffEqArrayOperator, ::Type{T}, dims::Dims) where T = similar(L.A, T, dims)
+Base.adjoint(L::DiffEqArrayOperator) = DiffEqArrayOperator(L.A'; update_func = (A,u,p,t) -> L.update_func(L.A,u,p,t)')
```
ChrisRackauckas, line 92: and a `has_adjoint` overload

vpuri3:
> No. What needs to be done is the operator interface documentation should get a note about adjoint(::SciMLOperator) as being a part of the (optional) interface, required for reverse mode automatic differentiation. Then, adjoint(::DiffEqArrayOperator) should be added to SciMLBase by just taking the adjoint of the internal array. Many other operators would have to be handled though, but that can be done over time.

not sure where the documentation note needs to be added. could you do it?

vpuri3:
should we wrap `A` in `DiffEqArrayOperator` so that it has `has_adjoint` defined?

https://github.com/SciML/SciMLBase.jl/blob/4656c0f0d09f46bbbcd93bcb750a3d7902efffbd/src/problems/basic_problems.jl#L18-L26

vpuri3:
@ChrisRackauckas if not ^ then good to go

src/operators/basic_operators.jl
```diff
@@ -86,9 +86,11 @@ struct DiffEqArrayOperator{T,AType<:AbstractMatrix{T},F} <: AbstractDiffEqLinear
     new{eltype(A),AType,typeof(update_func)}(A, update_func)
 end
 
+has_adjoint(::DiffEqArrayOperator) = true
```
ChrisRackauckas, line 89: add `has_adjoint(::AbstractMatrix) = true` as well. That's defined in the abstract matrix interface that it can be done via index swapping.

vpuri3:
maybe we can add overloads for `FactorizedDiffEqArrayOperator`

vpuri3:
@ChrisRackauckas good to go otherwise

src/operators/basic_operators.jl
```diff
@@ -116,22 +118,40 @@ Like DiffEqArrayOperator, but stores a Factorization instead.
...
                                                               }
                                     } <: AbstractDiffEqLinearOperator{T}
   F::FType
 end
 
-Base.Matrix(L::FactorizedDiffEqArrayOperator) = Matrix(L.F)
-Base.convert(::Type{AbstractMatrix}, L::FactorizedDiffEqArrayOperator) = convert(AbstractMatrix, L.F)
+function Base.convert(::Type{AbstractMatrix}, L::FactorizedDiffEqArrayOperator)
+    if L.F isa Adjoint
+        convert(AbstractMatrix,L.F')'
```
ChrisRackauckas, line 131: Why is this overload necessary? Does it not have a fallback?

ChrisRackauckas:
Good to go sans that one comment. Let's clear that up and it's ready.

vpuri3:
^no apparently not.
```julia
julia> using LinearAlgebra; A=lu!(rand(2,2))'                                   
Adjoint of LU{Float64, Matrix{Float64}, Vector{Int64}}                          
L factor:                                                                       
2×2 Matrix{Float64}:                                                            
 1.0       0.0                          
 0.449258  1.0                                                                  
U factor:                                                                       
2×2 Matrix{Float64}:                                                            
 0.837142  0.439897                                                             
 0.0       0.0989137                                                                                                                                            
julia> convert(AbstractMatrix, A) * rand(2)                                     
ERROR: CanonicalIndexError: getindex not defined for Adjoint{Float64, LU{Float64, Matrix{Float64}, Vector{Int64}}}                                              
Stacktrace:                             
 [1] error_if_canonical_getindex(::IndexCartesian, ::Adjoint{Float64, LU{Float64, Matrix{Float64}, Vector{Int64}}}, ::Int64, ::Int64)                           
   @ Base ./abstractarray.jl:1255
```

ChrisRackauckas:
I don't think the adjoint of a factorization is the factorization of the adjoint:

```julia
julia> using LinearAlgebra

julia> A = rand(2,2)
2×2 Matrix{Float64}:
 0.123926  0.948724
 0.471701  0.279714

julia> lu(A)'
Adjoint of LU{Float64, Matrix{Float64}}
L factor:
2×2 Matrix{Float64}:
 1.0       0.0
 0.262722  1.0
U factor:
2×2 Matrix{Float64}:
 0.471701  0.279714
 0.0       0.875237

julia> lu(A')
LU{Float64, Matrix{Float64}}
L factor:
2×2 Matrix{Float64}:
 1.0       0.0
 0.130624  1.0
U factor:
2×2 Matrix{Float64}:
 0.948724  0.279714
 0.0       0.435163
```

Let's just error on this one for now and merge.

vpuri3:
that is correct, so for
```julia
julia> FFt = rand(n,n) |> DiffEqArrayOperator |> factorize |> adjoint;

julia> FFt.F |> typeof
Adjoint{Float64, LU{Float64, Matrix{Float64}, Vector{Int64}}}
```

in the fwd application `FFt(u,p,t)`, we convert back to the original `A` and take its adjoint. and the inverse `FFt \ u` just works.

PR #149, opened 2022-04-21
display :minimizer property with tab completion

IlyaOrson:
ref https://github.com/SciML/DiffEqFlux.jl/issues/702


src/solutions/optimization_solutions.jl
```diff
@@ -39,6 +39,10 @@ Base.@propagate_inbounds function Base.getproperty(x::AbstractOptimizationSoluti
     return getfield(x,s)
 end
 
+function Base.propertynames(x::AbstractOptimizationSolution)
+    return (:minimizer, fieldnames(typeof(x))...)
+end
+
```
ChrisRackauckas, line 45: we're phasing this one out, since it's not always the minimizer.

PR #151, opened 2022-04-26
Add docstrings for all of the problem types

ChrisRackauckas:


src/problems/bvp_problems.jl
```diff
@@ -3,8 +3,73 @@ $(TYPEDEF)
...
+### Fields
+
+* `f`: The function for the ODE.
+* `bc`: The boundary condition function.
+* `u0`: The initial condition. Either the initial condition for the ODE as an
+  initial value problem, or a `Vector` of values for ``u(t_i)`` for collocation
+  methods
+* `tspan`: The timespan for the problem.
+* `p`: The parameters for the problem. Defaults to `NullParameters`
+* `kwargs`: The keyword arguments passed onto the solves.
```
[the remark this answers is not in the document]
ChrisRackauckas: Some of them use it, some of them don't. But I think that would be "perfect is the enemy of good" here, since I won't have the time to split them all up so I think I'll plan to merge and open issue about splitting out the fields docs.

src/problems/bvp_problems.jl
```diff
@@ -3,8 +3,73 @@ $(TYPEDEF)
...
+@doc doc"""
 $(TYPEDEF)
+
+Defines an BVP problem.
+Documentation Page: https://diffeq.sciml.ai/stable/types/bvp_types/
+
+## Mathematical Specification of a BVP Problem
+
+To define a BVP Problem, you simply need to give the function ``f`` and the initial
+condition ``u₀`` which define an ODE:
```
odow, line 15: Potentially should be ``` ``u_0`` ```.

src/problems/dae_problems.jl
```diff
@@ -1,6 +1,70 @@
...
 $(TYPEDEF)
+
+Defines an implicit ordinary differential equation (ODE) or 
+differential-algebraic equation (DAE) problem.
+Documentation Page: https://diffeq.sciml.ai/stable/types/dae_types/
+
+## Mathematical Specification of an DAE Problem
+
+To define a DAE Problem, you simply need to give the function ``f`` and the initial
+condition ``u₀`` which define an ODE:
```
odow, line 11: Ditto

Issue #159, opened 2022-05-20
Finish the QuadratureProblem docstring


ChrisRackauckas:
Left unfinished at the end of https://github.com/SciML/SciMLBase.jl/pull/157 . @utkarsh530 could I enlist you to help get this one finished?

utkarsh530:
Yeah I can look into it

Issue #158, opened 2022-05-20
Finish the OptimizationProblem docstring


ChrisRackauckas:
Left incomplete at the end of https://github.com/SciML/SciMLBase.jl/pull/157 . @Vaibhavdixit02 could I enlist you to help this get finished?

Vaibhavdixit02:
Sure 👍

PR #174, opened 2022-05-28
Improved OptimizationProblem, OptimizationFunction, and Quadrature docs

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLBase.jl/issues/159 and fixes https://github.com/SciML/SciMLBase.jl/issues/158


ChrisRackauckas:
@Vaibhavdixit02 @ValentinKaisermayer notice the full definition of the interface. We don't quite hit it right now (especially with the IIP stuff), but this is where we need to go for full interface conformity. You might want to check out https://scimlbase.sciml.ai/dev/ for the full interface definition as well.

PR #180, opened 2022-06-10
Throw early warning function argument errors

ChrisRackauckas:
Note that there is a broken case where `f` is `iip` and `jac` has `oop` and `iip` dispatches, but it throws a non-conforming error. That will take quite a bit of code to solve and is quite an edge case, so I'm not going to let that block this and will just keep as `@test_broken` with an issue open.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -2629,9 +2629,9 @@ function NonlinearFunction{iip,true}(f;
     _colorvec = colorvec
   end
 
-  jaciip = jac !== nothing ? isinplace(jac,6,"jac") : iip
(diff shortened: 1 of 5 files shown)


src/scimlfunctions.jl
```diff
@@ -1901,6 +1959,22 @@ function SDEFunction{iip,true}(f,g;
                    _colorvec = colorvec
                  end
 
+                 giip = isinplace(g,4,"g")
```
isaacsas, line 1962: @ChrisRackauckas does this work if a user gives both iip and oop versions but sets `iip=false? That is what ModelingToolkit does by default.

PR #182, opened 2022-06-11
Add dispatch function checks to OptimizationFunctions

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02 I plan to extend the dispatch function checking and kwarg checking to Optimization.jl over the next week.

PR #184, opened 2022-06-12
Fix overreactive function checking

ChrisRackauckas:


isaacsas:
Would it make sense to add ModelingToolkit as an integration test at some point? I think that would have caught this one.

ChrisRackauckas:
I did

Issue #191, opened 2022-06-23
kwargs field makes `IntegralProblem` type unstable


sharanry:
```julia
function copy_(prob)
    prob_ = IntegralProblem(prob.f, prob.lb, prob.ub, prob.p , nout = prob.nout , batch = prob.batch , prob.kwargs...) 
end
@code_warntype copy_(prob)
```

# Related
- https://github.com/SciML/Integrals.jl/issues/65

CC: @ChrisRackauckas @YingboMa

sharanry:
@ChrisRackauckas is it okay to remove that field? That would enable making `transformation_if_inf` in https://github.com/SciML/Integrals.jl/issues/65 type stable.

ChrisRackauckas:
Can we make it type stable by making it a namedtuple?

It's supposed to be passed on as is normal for the other probs. It's a bug that it isn't.

PR #193, opened 2022-06-23
Make `IntegralProblem` type stable

sharanry:
Solves https://github.com/SciML/SciMLBase.jl/issues/191
Related: https://github.com/SciML/Integrals.jl/issues/65

CC: @ChrisRackauckas Yingbo


devmotion:
> Solves #191

How does it solve the issue? It seems the `kwargs` field is not changed by the PR, and the two fields `batch` and `nout` are already concretely typed.

ChrisRackauckas:
I have the same question.

sharanry:
Yeah. This doesn't make sense. Closing it.

PR #200, opened 2022-07-05
Complete OptimizationFunction and OptimizationProblem docstrings

Vaibhavdixit02:
Fixes https://github.com/SciML/Optimization.jl/issues/291

### DIFF
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -315,7 +315,7 @@ Any extra keyword arguments are captured to be sent to the optimizers.
 
 Both inequality and equality constraints are defined by the `f.cons` function in the `OptimizationFunction`
 description of the problem structure. This `f.cons` is given as a function `f.cons(u,p)` which computes
-the value at `u` of the constraints. For example, take `f.cons(u,p) = u[1] - u[2]` 
+the value of the constraints at `u`. For example, take `f.cons(u,p) = u[1] - u[2]`.
 With these definitions, `lcons` and `ucons` define the bounds on the constraint that the solvers try to satisfy.
 If `lcons` and `ucons` are `nothing`, then there are no constraints bounds, meaning that the constraint is satisfied when `-Inf < f.cons < Inf` (which of course is always!). If `lcons[i] = ucons[i] = 0`, then the constraint is satisfied when `f.cons(u,p)[i] = 0`, and so this implies the equality constraint ``u[1] = u[2]`. If `lcons[i] = ucons[i] = a`, then ``u[1] - u[2] = a`` is the equality constraint. 
 
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -305,11 +305,24 @@ Any extra keyword arguments are captured to be sent to the optimizers.
 * `lb`: the lower bounds for the optimization of `u`.
 * `ub`: the upper bounds for the optimization of `u`.
 * `lcons`: the vector of lower bounds for the constraints passed to [OptimizationFunction](@ref).
-    Defaults to `nothing`, implying no lower bounds for the constraints.
+    Defaults to `nothing`, implying no lower bounds for the constraints (i.e. the constraint bound is `-Inf`)
 * `ucons`: the vector of upper bounds for the constraints passed to [OptimizationFunction](@ref).
-    Defaults to `nothing`, implying no upper bounds for the constraints.
+    Defaults to `nothing`, implying no upper bounds for the constraints (i.e. the constraint bound is `Inf`)
 * `sense`: the objective sense, can take `MaxSense` or `MinSense` from Optimization.jl.
 * `kwargs`: the keyword arguments passed on to the solvers.
+
+## Inequality and Equality Constraints
+
+Both inequality and equality constraints are defined by the `f.cons` function in the `OptimizationFunction`
+description of the problem structure. This `f.cons` is given as a function `f.cons(u,p)` which computes
+the value at `u` of the constraints. For example, take `f.cons(u,p) = u[1] - u[2]` 
+With these definitions, `lcons` and `ucons` define the bounds on the constraint that the solvers try to satisfy.
+If `lcons` and `ucons` are `nothing`, then there are no constraints bounds, meaning that the constraint is satisfied when `-Inf < f.cons < Inf` (which of course is always!). If `lcons[i] = ucons[i] = 0`, then the constraint is satisfied when `f.cons(u,p)[i] = 0`, and so this implies the equality constraint ``u[1] = u[2]`. If `lcons[i] = ucons[i] = a`, then ``u[1] - u[2] = a`` is the equality constraint. 
+
+Inequality constraints are then given by making `lcons[i] != ucons[i]`. For example, `lcons[i] = -Inf` and `ucons[i] = 0` would imply the inequality constraint ``u[1] < u[2]`` since any `f.cons[i] < 0` satisfies the constraint. Similarly, `lcons[i] = -1` and `ucons[i] = 1` would imply that `-1 < f.cons[i] < 1` is required or ``-1 < u[1] - u[2] < 1``.
+
+Note that these vectors must be sized to match the number of constraints, with one set of conditions for each constraint.
+
 """
 struct OptimizationProblem{iip, F, uType, P, B, LC, UC, S, K} <:
        AbstractOptimizationProblem{iip}
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -198,14 +198,14 @@ IntegralProblem{iip}(f,lb,ub,p=NullParameters();
 - ub: Either a number or vector of upper bounds.
 - p: The parameters associated with the problem.
 - nout: The output size of the function f. Defaults to 1, i.e., a scalar integral output.
-- batch: The preferred number of points to batch. This allows user-side parallelization 
-  of the integrand. If batch != 0, then each x[:,i] is a different point of the integral 
-  to calculate, and the output should be nout x batchsize. Note that batch is a suggestion 
-  for the number of points, and it is not necessarily true that batch is the same as 
+- batch: The preferred number of points to batch. This allows user-side parallelization
+  of the integrand. If batch != 0, then each x[:,i] is a different point of the integral
+  to calculate, and the output should be nout x batchsize. Note that batch is a suggestion
+  for the number of points, and it is not necessarily true that batch is the same as
   batchsize in all algorithms.
 - kwargs:: Keyword arguments copied to the solvers.
 
-Additionally, we can supply iip like IntegralProblem{iip}(...) as true or false to declare at 
+Additionally, we can supply iip like IntegralProblem{iip}(...) as true or false to declare at
 compile time whether the integrator function is in-place.
 
 ### Fields
@@ -250,9 +250,9 @@ min_u f(u,p)
 ```
 
 ``u₀`` is an initial guess of the minimum. `f` should be specified as `f(u,p)`
-and `u₀` should be an AbstractArray (or number) whose geometry matches the 
-desired geometry of `u`. Note that we are not limited to numbers or vectors 
-for `u₀`; one is allowed to provide `u₀` as arbitrary matrices / 
+and `u₀` should be an AbstractArray (or number) whose geometry matches the
+desired geometry of `u`. Note that we are not limited to numbers or vectors
+for `u₀`; one is allowed to provide `u₀` as arbitrary matrices /
 higher-dimension tensors as well.
 
 ## Problem Type
@@ -299,15 +299,17 @@ Any extra keyword arguments are captured to be sent to the optimizers.
 
 ### Fields
 
-* `f`: The function in the problem.
-* `u0`: The initial guess for the optima.
-* `p`: The parameters for the problem. Defaults to `NullParameters`.
+* `f`: the function in the problem.
+* `u0`: the initial guess for the optima.
+* `p`: the parameters for the problem. Defaults to `NullParameters`.
 * `lb`: the lower bounds for the optimization of `u`.
 * `ub`: the upper bounds for the optimization of `u`.
-* `lcons`:
-* `ucons`:
-* `sense`:
-* `kwargs`: The keyword arguments passed on to the solvers.
+* `lcons`: the vector of lower bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no lower bounds for the constraints.
+* `ucons`: the vector of upper bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no upper bounds for the constraints.
+* `sense`: the objective sense, can take `MaxSense` or `MinSense` from Optimization.jl.
+* `kwargs`: the keyword arguments passed on to the solvers.
 """
 struct OptimizationProblem{iip, F, uType, P, B, LC, UC, S, K} <:
        AbstractOptimizationProblem{iip}
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -319,7 +319,7 @@ the value at `u` of the constraints. For example, take `f.cons(u,p) = u[1] - u[2
 With these definitions, `lcons` and `ucons` define the bounds on the constraint that the solvers try to satisfy.
 If `lcons` and `ucons` are `nothing`, then there are no constraints bounds, meaning that the constraint is satisfied when `-Inf < f.cons < Inf` (which of course is always!). If `lcons[i] = ucons[i] = 0`, then the constraint is satisfied when `f.cons(u,p)[i] = 0`, and so this implies the equality constraint ``u[1] = u[2]`. If `lcons[i] = ucons[i] = a`, then ``u[1] - u[2] = a`` is the equality constraint. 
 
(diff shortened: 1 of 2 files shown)


gaurav-arya:
Hey, a couple nits regarding the constraints docs:

- it's still not clear to me how to specify an equality constraint
- the docs for `cons_j`, `cons_h`  still mention "equality constraints" rather than constraint functions

Thanks!

gaurav-arya:
Thank you! I'm still a little confused regarding a couple points:

- If both `lcons[i]` and `ucons[i]` are `nothing`, do we have an equality constraint (because they are equal values) or a meaningless inequality constraint (because there is no lower bound or upper bound)?
- Is the intention behind `lcons[i] == ucons[i]` for equality constraints that this can be recognized before calling out to the optimization algorithm? Because I'd imagine that many optimization algorithms may have a special interface for equality constraints, rather than specifying it like this?

DanielVandH:
I have the same thoughts as @gaurav-arya. I think it will end up being a bit messy trying to get it so perfectly clear in the documentation, though. What if this page http://optimization.sciml.ai/stable/tutorials/rosenbrock/ was extended to actually write out the optimisation problems being solved, so that the documentation can refer to it for more detail? i.e. in the section of the linked page starting with

```
# Now derivative-based optimizers with various constraints
```

it would make the page a lot more useful to actually see the equations written out, at least for what constraints are being used, and would probably clear up some of these ambiguities; maybe more examples are needed to cover certain cases. That's something that e.g. https://diffeq.sciml.ai/stable/tutorials/ode_example/ does very well, in that we actually get to see some math rather than a block of code.

Vaibhavdixit02:
>If both lcons[i] and ucons[i] are nothing, do we have an equality constraint (because they are equal values) or a meaningless inequality constraint (because there is no lower bound or upper bound)? (I imagine the latter, but technically speaking the current wording seems ambiguous?)

No, in that case it'd be treated as unbounded, so -Inf to Inf.

>Is the intention behind lcons[i] == ucons[i] for equality constraints that this can be recognized before calling out to the optimization algorithm? Because I'd imagine that many optimization algorithms may have a special interface for equality constraints, rather than specifying it like this?

Yes, but right now we don't need to do it on our end since both MOI and Optim do it internally so we just forward the bounds and they are able to figure it out, later if it needs to be set explicitly for some solver package we should be able to do it easily this way.

>What if this page http://optimization.sciml.ai/stable/tutorials/rosenbrock/ was extended to actually write out the optimisation problems being solved, so that the documentation can refer to it for more detail? i.e. in the section of the linked page starting with

I think that's a very good idea and I do plan to add that, though note that is going to have to be done in Optimization's docs 's examples and this PR aims to modify only the docstrings here for API reference

ChrisRackauckas:
@DanielVandH yes, right now Optimization.jl is missing a proper tutorial, that's one of the next things in the line up to complete Optimization.jl, which to me looks like:

1. Complete the reference docstrings (i.e. this PR, along with docstrings to the algorithms)
2. https://github.com/SciML/Optimization.jl/issues/310
3. https://github.com/SciML/Optimization.jl/issues/291
4. https://github.com/SciML/Optimization.jl/issues/282
5. https://github.com/SciML/SciMLBase.jl/pull/182
6. New "real" tutorial that walks someone through the whole process, like DifferentialEquations.jl's ODE tutorial

DanielVandH:
I see, @ChrisRackauckas. I think these modifications are good then @Vaibhavdixit02. Thanks!

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,17 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
 * `ub`: the upper bounds for the optimization of `u`.
-* `lcons`:
-* `ucons`:
-* `sense`:
-* `kwargs`: The keyword arguments passed on to the solvers.
+* `lcons`: the vector of lower bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no lower bounds for the constraints.
+* `ucons`: the vector of upper bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no upper bounds for the constraints.
+* `sense`: the objective sense, can take `MaxSense` or `MinSense` from Optimization.jl.
```
ChrisRackauckas, line 311: These should be defined in SciMLBase, otherwise you cannot fully define the problem without the solver interface 😅.

ChrisRackauckas: Also, is it actually implemented everywhere? I don't think it's well tested? I'm scared to say it exists if it's ignored and doesn't throw errors on non-compliant solvers.

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,17 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
+* `u0`: the initial guess for the optima.
+* `p`: the parameters for the problem. Defaults to `NullParameters`.
 * `lb`: the lower bounds for the optimization of `u`.
 * `ub`: the upper bounds for the optimization of `u`.
-* `lcons`:
-* `ucons`:
-* `sense`:
-* `kwargs`: The keyword arguments passed on to the solvers.
+* `lcons`: the vector of lower bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no lower bounds for the constraints.
```
ChrisRackauckas, line 308:
```suggestion
    Defaults to `nothing`, implying no lower bounds for the constraints (i.e. the constraint bound is `-Inf`)
```

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,17 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
+* `ucons`: the vector of upper bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no upper bounds for the constraints.
```
ChrisRackauckas, line 310:
```suggestion
    Defaults to `nothing`, implying no upper bounds for the constraints (i.e. the constraint bound is `Inf`)
```

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,17 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
+* `sense`: the objective sense, can take `MaxSense` or `MinSense` from Optimization.jl.
+* `kwargs`: the keyword arguments passed on to the solvers.
 """
```
ChrisRackauckas, line 313:
```suggestion
* `kwargs`: the keyword arguments passed on to the solvers.

## Inequality and Equality Constraints

Both inequality and equality constraints are defined by the `f.cons` function in the `OptimizationFunction`
description of the problem structure. This `f.cons` is given as a function `f.cons(res,u,p)` which computes
the residual `res` of `u` on the constraints. For example, take `f.cons(res,u,p) = u[1] - u[2]` 
With these definitions, `lcons` and `ucons` define the bounds on the constraint that the solvers try to satisfy.
If `lcons` and `ucons` are `nothing`, then there are no constraints bounds, meaning that the constraint is satisfied when `-Inf < res < Inf` (which of course is always!). If `lcons[i] = ucons[i] = 0`, then the constraint is satisfied when `res = 0`, and so this implies the equality constraint ``u_1 = u_2`. If `lcons[i] = ucons[i] = a`, then ``u_1 - u_2 = a`` is the equality constraint. 

Inequality constraints are then given by making `lcons[i] != ucons[i]`. For example, `lcons[i] = -Inf` and `ucons[i] = 0` would imply the inequality constraint ``u_1 < u_2`` since any `res < 0` satisfies the constraint. Similarly, `lcons[i] = -1` and `ucons[i] = 1` would imply that `-1 < res < 1` is required or ``-1 < u_1 - u_2 < 1``.

Note that these vectors must be sized to match the number of constraints, with one set of conditions for each constraint.

"""
```

ChrisRackauckas:
I added a suggestion with a section that fully describes how `lcons` and `ucons` is used. @Vaibhavdixit02 @gaurav-arya see if this makes sense. 

@gaurav-arya one of the reasons for doing it like this is that it's the style that MOI and Optim expect, which makes it easy to wrap. We will need to do something special to handle NLopt.jl I think, so the easy thing for now would be to disallow inequality constraints with NLopt and just make them equalities to start, until we make a way to split it.

Though @Vaibhavdixit02 this does raise the question as to why we should define `ncons` in the OptimizationFunction if `lcons` and `ucons` are required and `ncons == length(lcons) == length(ucons)`. We should assert that's true (or nothing), or otherwise we should probably just drop the `ncons` and just assert `length(lcons) == length(ucons)` and use that for `ncons`.

src/problems/basic_problems.jl
DanielVandH: This new section is perfect for the original question that I had in my issue; addresses everything and is very clear with those examples.

src/problems/basic_problems.jl
Vaibhavdixit02: > otherwise you cannot fully define the problem without the solver interface

True, I'll move it here

>is it actually implemented everywhere? I don't think it's well tested?

Not everywhere - https://github.com/SciML/Optimization.jl/search?q=MAXSENSE but it is tested

>I'm scared to say it exists if it's ignored and doesn't throw errors on non-compliant solvers.

It doesn't throw errors that's true, but it should be easy to add and I think it is a part of the larger issue https://github.com/SciML/Optimization.jl/issues/282, I'll make a note of it there, should be handled together

src/scimlfunctions.jl
```diff
@@ -1537,23 +1537,27 @@ OptimizationFunction{iip}(f,adtype::AbstractADType=NoAD();
...
 - `hess(H,u,p)` or `H=hess(u,p)`: the Hessian of `f` with respect to `u`
 - `hv(Hv,u,v,p)` or `Hv=hv(u,v,p)`: the Hessian-vector product ``\frac{d^2 f}{du^2} v``.
-- `cons(res,x,p)` or `res=cons(x,p)`: the equality constraints vector, where the constraints
-  are satisfied when `res = 0`.
-- `cons_j(res,x,p)` or `res=cons_j(x,p)`: the Jacobian of the equality constraints.
-- `cons_h(res,x,p)` or `res=cons_h(x,p)`: the Hessian of the equality constratins, provided as
-  and array of Hessians with `res[i]` being the Hessian with respect to the `i`th output on `cons`.
+- `cons(res,x,p)` or `res=cons(x,p)`: the constraints function, should return a vector
+    with value of the `i`th constraint, evaluated at the current values of variables
+    inside the optimization routine, at `res[i]`. This takes just the function evaluations
```
Vaibhavdixit02, line 1542:
```suggestion
    inside the optimization routine. This takes just the function evaluations
```

Vaibhavdixit02:
>Though @Vaibhavdixit02 this does raise the question as to why we should define ncons in the OptimizationFunction if lcons and ucons are required and ncons == length(lcons) == length(ucons). We should assert that's true (or nothing), or otherwise we should probably just drop the ncons and just assert length(lcons) == length(ucons) and use that for ncons.

We don't do that anymore, it is indeed taken from the length of the bounds in the `instantiate_function`

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,30 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
+* `ucons`: the vector of upper bounds for the constraints passed to [OptimizationFunction](@ref).
+    Defaults to `nothing`, implying no upper bounds for the constraints (i.e. the constraint bound is `Inf`)
+* `sense`: the objective sense, can take `MaxSense` or `MinSense` from Optimization.jl.
+* `kwargs`: the keyword arguments passed on to the solvers.
+
+## Inequality and Equality Constraints
+
+Both inequality and equality constraints are defined by the `f.cons` function in the `OptimizationFunction`
+description of the problem structure. This `f.cons` is given as a function `f.cons(u,p)` which computes
+the value at `u` of the constraints. For example, take `f.cons(u,p) = u[1] - u[2]`
```
gaurav-arya, line 318:
```suggestion
the value of the constraints at `u`. For example, take `f.cons(u,p) = u[1] - u[2]`.
```

src/problems/basic_problems.jl
```diff
@@ -299,15 +299,30 @@ Any extra keyword arguments are captured to be sent to the optimizers.
...
+
+Both inequality and equality constraints are defined by the `f.cons` function in the `OptimizationFunction`
+description of the problem structure. This `f.cons` is given as a function `f.cons(u,p)` which computes
+the value at `u` of the constraints. For example, take `f.cons(u,p) = u[1] - u[2]` 
+With these definitions, `lcons` and `ucons` define the bounds on the constraint that the solvers try to satisfy.
+If `lcons` and `ucons` are `nothing`, then there are no constraints bounds, meaning that the constraint is satisfied when `-Inf < f.cons < Inf` (which of course is always!). If `lcons[i] = ucons[i] = 0`, then the constraint is satisfied when `f.cons(u,p)[i] = 0`, and so this implies the equality constraint ``u[1] = u[2]`. If `lcons[i] = ucons[i] = a`, then ``u[1] - u[2] = a`` is the equality constraint. 
+
+Inequality constraints are then given by making `lcons[i] != ucons[i]`. For example, `lcons[i] = -Inf` and `ucons[i] = 0` would imply the inequality constraint ``u[1] < u[2]`` since any `f.cons[i] < 0` satisfies the constraint. Similarly, `lcons[i] = -1` and `ucons[i] = 1` would imply that `-1 < f.cons[i] < 1` is required or ``-1 < u[1] - u[2] < 1``.
```
gaurav-arya, line 322:
```suggestion
Inequality constraints are then given by making `lcons[i] != ucons[i]`. For example, `lcons[i] = -Inf` and `ucons[i] = 0` would imply the inequality constraint ``u[1] <= u[2]`` since any `f.cons[i] <= 0` satisfies the constraint. Similarly, `lcons[i] = -1` and `ucons[i] = 1` would imply that `-1 <= f.cons[i] <= 1` is required or ``-1 <= u[1] - u[2] <= 1``.
```

src/problems/basic_problems.jl
gaurav-arya: In-place v.s. out-of-place is still a little confusing to me, but perhaps that's because of the state of the rest of the optimization docs. If it's not addressed by any other issues I would suggest https://github.com/SciML/Optimization.jl/issues/312 be added to the roadmap

gaurav-arya:
The docs clarify my issues! With the exception that in-place v.s. out-of-place, e.g. for the example constraint functions, is still a little confusing to me, but probably the rest of the optimization docs need to clarify that first. If it's not addressed by any other issues I would suggest SciML/Optimization.jl#312 be added to the roadmap, as the docs seem confusing about this point currently

src/scimlfunctions.jl
```diff
@@ -1537,23 +1537,27 @@ OptimizationFunction{iip}(f,adtype::AbstractADType=NoAD();
...
-  and array of Hessians with `res[i]` being the Hessian with respect to the `i`th output on `cons`.
+- `cons(x,p)` : the constraints function, should return a vector
+    with value of the `i`th constraint, evaluated at the current values of variables
+    inside the optimization routine. This takes just the function evaluations
+    and the equality or inequality assertion is applied by the solver based on the constraint
+    bounds passed as `lcons` and `ucons` to [OptimizationProblem](@ref), in case of equality
+    constraints `lcons` and `ucons` should be passed equal values.
+- `cons_j(res,x,p)` or `res=cons_j(x,p)`: the Jacobian of the constraints.
+- `cons_h(res,x,p)` or `res=cons_h(x,p)`: the Hessian of the constraints, provided as
+   an array of Hessians with `res[i]` being the Hessian with respect to the `i`th output on `cons`.
```
ChrisRackauckas, line 1548: This change is incorrect, which is what I think @gaurav-arya is confused about. We should have an in-place function for the constraints which corresponds to `OptimizationFunction{true}`, and then `res=cons(x,p)` should be the one created when `OptimizationFunction{false}`.

Basically, if `OptimizationFunction{true}`, then every single function should be mutating the first argument. `cons!(res,x,p)`, etc. The only exception is the `f` itself because it always returns a scalar (so there's no mutation possible). If `OptimizationFunction{false}`, then every function should return a vector.

Now I know that the solver backends have not been good at separating these interfaces (and generally use a combination of the two), but we should consider that an implementation detail for the solver wrappers. If a solver library only takes out of place versions for the constraints, then the wrapper should know that, check `if isinplace(prob)`, and if it is in place do `(x,p) -> f.cons(res,x,p); res)` with some `res` that it creates. If the solver only wants an in-place version, it should wrap the other direction. But the user shouldn't have to worry about this: these are implementation details about the solver that effect the performance of a solver, but every solver should try to work correctly (or error), with fully in-place definitions supporting a fully mutating non-allocating version (if any solvers support that), and fully out-of-place definitions supporting immutable arrays (if any solvers support that).

src/scimlfunctions.jl
```diff
@@ -1537,23 +1537,27 @@ OptimizationFunction{iip}(f,adtype::AbstractADType=NoAD();
 - `grad(G,u,p)` or `G=grad(u,p)`: the gradient of `f` with respect to `u`
 - `hess(H,u,p)` or `H=hess(u,p)`: the Hessian of `f` with respect to `u`
 - `hv(Hv,u,v,p)` or `Hv=hv(u,v,p)`: the Hessian-vector product ``\frac{d^2 f}{du^2} v``.
-- `cons(res,x,p)` or `res=cons(x,p)`: the equality constraints vector, where the constraints
-  are satisfied when `res = 0`.
-- `cons_j(res,x,p)` or `res=cons_j(x,p)`: the Jacobian of the equality constraints.
-- `cons_h(res,x,p)` or `res=cons_h(x,p)`: the Hessian of the equality constratins, provided as
-  and array of Hessians with `res[i]` being the Hessian with respect to the `i`th output on `cons`.
+- `cons(x,p)` : the constraints function, should return a vector
```
ChrisRackauckas, line 1540:
```suggestion
- `cons(res,x,p)` or `cons(x,p)` : the constraints function, should return or mutate a vector
```

PR #201, opened 2022-07-07
`responsible_map` for `AbstractArray`s

chriselrod:


ChrisRackauckas:
Can we get a test case for this?

I almost feel like this separate map should be its own package if there's time. It's starting to become a whole thing.

chriselrod:
> I almost feel like this separate map should be its own package if there's time. It's starting to become a whole thing.

Sure, that makes sense. Any name suggestions? `ReponsibleMap.jl`, `NarrowingMap.jl`, `StableMap.jl`, `UnionMap.jl`, `FastMap.jl` (mostly for symmetry with `FastBroadcast`, `map` itself probably shouldn't be slow, or at least not for the same reasons `Base.Broadcast` is), or something else?

chriselrod:
https://github.com/chriselrod/StableMap.jl
While fast in some cases:
```julia
julia> using StableMap, ForwardDiff, BenchmarkTools

julia> f(x) = x > 1 ? x : 1.0

julia> x = [ForwardDiff.Dual(0f0,1f0), ForwardDiff.Dual(2f0,1f0)];

julia> y = [ForwardDiff.Dual(2f0,1f0), ForwardDiff.Dual(0f0,1f0)];

julia> @btime map(f, $x)
  77.097 ns (4 allocations: 176 bytes)
2-element Vector{Real}:
               1.0
 Dual{Nothing}(2.0,1.0)

julia> @btime stable_map(f, $x)
  35.209 ns (1 allocation: 96 bytes)
2-element Vector{ForwardDiff.Dual{Nothing, Float64, 1}}:
 Dual{Nothing}(1.0,0.0)
 Dual{Nothing}(2.0,1.0)

julia> @btime map(f, $y)
  77.901 ns (4 allocations: 176 bytes)
2-element Vector{Real}:
 Dual{Nothing}(2.0,1.0)
               1.0

julia> @btime stable_map(f, $y)
  35.114 ns (1 allocation: 96 bytes)
2-element Vector{ForwardDiff.Dual{Nothing, Float64, 1}}:
 Dual{Nothing}(2.0,1.0)
 Dual{Nothing}(1.0,0.0)
```
It's very slow in others:
```julia
m(x) = x > 1.0 ? x : [x]
g(x) = Base.inferencebarrier(x > 1 ? x : 1.0)
h(x) = Base.inferencebarrier(x)
```
It still returns tight container types, but the performance is unexpectedly bad in many cases.
`m` is especially surprising, given that I'd have thought it'd be fast thanks to union splitting.

PR #202, opened 2022-07-08
Base.convert method to translate a `NonlinearFunction` to an `ODEfunction`

frankschae:


ChrisRackauckas:
This should get a unit test.

src/scimlfunctions.jl
```diff
@@ -3890,3 +3890,28 @@ for S in [:ODEFunction
...
+"""
+function Base.convert(::Type{ODEFunction}, f::NonlinearFunction)
+    iip = isinplace(f)
+    _f = iip ? (du, u, p, t) -> (f.f(du, u, p); nothing) : (u, p, t) -> f.f(u, p)
+
+    ODEFunction{iip}(_f;
+                     mass_matrix = f.mass_matrix,
+                     analytic = f.analytic,
+                     tgrad = f.tgrad,
+                     jac = f.jac,
```
ChrisRackauckas, line 3907: this would need to add a `t` to it as well?

src/scimlfunctions.jl
```diff
@@ -3890,3 +3890,28 @@ for S in [:ODEFunction
...
+Converts a NonlinearFunction into a ODEFunction.
+"""
+function Base.convert(::Type{ODEFunction}, f::NonlinearFunction)
+    iip = isinplace(f)
+    _f = iip ? (du, u, p, t) -> (f.f(du, u, p); nothing) : (u, p, t) -> f.f(u, p)
+
+    ODEFunction{iip}(_f;
+                     mass_matrix = f.mass_matrix,
+                     analytic = f.analytic,
+                     tgrad = f.tgrad,
```
ChrisRackauckas, line 3906: add `t`?

src/scimlfunctions.jl
```diff
@@ -3890,3 +3890,28 @@ for S in [:ODEFunction
...
+                     jac = f.jac,
+                     jvp = f.jvp,
```
ChrisRackauckas, line 3908: t

src/scimlfunctions.jl
```diff
@@ -3890,3 +3890,28 @@ for S in [:ODEFunction
...
+                     vjp = f.vjp,
```
ChrisRackauckas, line 3909: t

src/scimlfunctions.jl
frankschae: oh yeah, good catch. I'll add it with some tests.

PR #204, opened 2022-07-20
Add `Base.getindex` methods for `DEIntegrator`

AayushSabharwal:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1470

### DIFF
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -73,7 +73,8 @@ interpolated_integrator = integrator(0.0:1.0:10.0)
 @test interpolated_integrator[α, :] isa Vector
 @test interpolated_integrator[α, 2] isa Float64
 @test length(interpolated_integrator[α, 1:5]) == 5
-@test interpolated_integrator[α] ≈ 2interpolated_integrator[lorenz1.x] .+ interpolated_integrator[a] .* 2.0
+@test interpolated_integrator[α] ≈
+      2interpolated_integrator[lorenz1.x] .+ interpolated_integrator[a] .* 2.0
 @test collect(interpolated_integrator[t]) isa Vector
 @test collect(interpolated_integrator[t, :]) isa Vector
 @test interpolated_integrator[t, 2] isa Float64
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -37,111 +37,100 @@ p = [lorenz1.σ => 10.0,
 tspan = (0.0, 100.0)
 prob = ODEProblem(sys_simplified, u0, tspan, p)
 integrator = init(prob, Rodas4())
+step!(integrator, 100.0, true)
 
 @test_throws Any integrator[b]
-@test_throws Any integrator[b, 1]
-@test_throws Any integrator[b, 1:5]
-@test_throws Any integrator[b, [1, 2, 3]]
 @test_throws Any integrator['a']
-@test_throws Any integrator['a', 1]
-@test_throws Any integrator['a', 1:5]
-@test_throws Any integrator['a', [1, 2, 3]]
 
-@test integrator[a] isa AbstractVector
-@test integrator[a, 1] isa Real
-@test integrator[a, 1:5] isa AbstractVector
-@test integrator[a, [1, 2, 3]] isa AbstractVector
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
This needs tests. extend the symbolic indexing tests to use `init` and try these.

ChrisRackauckas (approved):

AayushSabharwal:
Codecov doesn't like this, it seems.

ChrisRackauckas:
Run the formatter. `using JuliaFormatter; format("path-to-repo")`

ChrisRackauckas:
Looks like real test failures?

AayushSabharwal:
That's weird. They passed when I ran them locally. I'll try and figure out what's up with this.

AayushSabharwal:
It seems that the fields with the same names in `ODEIntegrator` as `ODESolution` don't do the same thing/represent the same quantity, which makes sense. However, I'm at a bit of a loss as to what they actually are.

Additionally, do integrators support _all_ of the indexing capabilities of solutions? Are there some differences?

ChrisRackauckas:
> EDIT: Never mind, I think forwarding to integrator.sol should work

No, don't use that. That assumes someone is saving the solution. In fact, turn saving off in the tests.

> It seems that the fields with the same names in ODEIntegrator as ODESolution don't do the same thing/represent the same quantity, which makes sense. However, I'm at a bit of a loss as to what they actually are.

Integrators are only the current time point, solution is the whole time series. Because of this, indexing an integrator for a time series is undefined. `integrator[i]` is like `sol[i,j]` at the `j'th` time point.

AayushSabharwal:
Ah, got it, thanks! I'll update the tests first then fix the methods.

AayushSabharwal:
What does `sol[:, j]` do when `j` is an integer? I assumed timesteps would be floats

AayushSabharwal:
Why does `integrator(1.0)` work regardless of where `integrator` is in time? If they are only the current time, shouldn't it not work? Additionally, how do I turn of saving, as you mentioned to do?

AayushSabharwal:
Tests won't pass, so no point running them

AayushSabharwal:
The integrator tests pass now (considering I commented most of them since I can't explain why integrator works that way), and a couple come up as broken but it doesn't tell which.

test/downstream/integrator_indexing.jl
```diff
@@ -37,111 +37,100 @@ p = [lorenz1.σ => 10.0,
...
-@test_throws Any integrator['a', [1, 2, 3]]
 
-@test integrator[a] isa AbstractVector
-@test integrator[a, 1] isa Real
-@test integrator[a, 1:5] isa AbstractVector
-@test integrator[a, [1, 2, 3]] isa AbstractVector
+@test integrator[a] isa Real
+# @test integrator[a, 1] isa Real
+# @test integrator[a, 1:5] isa AbstractVector
+# @test integrator[a, [1, 2, 3]] isa AbstractVector
```
ChrisRackauckas, line 49: What do these mean? These should just throw.

test/downstream/integrator_indexing.jl
```diff
@@ -37,111 +37,100 @@ p = [lorenz1.σ => 10.0,
...
-@test integrator9.u isa Vector
-@test first(integrator9.u) isa Real
-@test length(integrator9.u) == 11
-@test length(integrator9.t) == 11
-@test collect(integrator9[t]) ≈ integrator3.t
-@test collect(integrator9[t, 1:5]) ≈ integrator3.t[1:5]
-
-integrator10 = integrator(0.1, idxs = 2)
-@test integrator10 isa Real
+# It doesn't because this returns a Vector{Vector{T}} and not a DiffEqArray
```
ChrisRackauckas, line 59: Good point, that would be good to solve separately.

ChrisRackauckas: Open an issue?

ChrisRackauckas:
Bring back the 3 that should throw and I think this is good.

test/downstream/integrator_indexing.jl
AayushSabharwal: Sure

AayushSabharwal:
Is there anything else left in this PR?

PR #209, opened 2022-08-01
Add `set_u!` method

### LINKED ISSUE #206: `set_u!` with symbols
Currently modifying the state of an integrator requires passing a `Vector` to `set_u!`, which is nontrivial when dealing with larger systems. Is there a way to do this symbolically, that I can't find? E.g. `set_u!(integrator, x, 1.23)`? Would this be feasible to implement?

AayushSabharwal:
Closes #206

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -333,6 +333,15 @@ function set_u!(integrator::DEIntegrator, u)
 end
 
 """
+    set_u!(integrator::DEIntegrator, sym, val)
+
+Set current state of symbol `sym` in `integrator` to `val`.
+"""
+function set_u!(integrator::DEIntegrator, sym, val)
+    error("set_u!: method has not been implemented for the integrator")
+end
+
+"""
     set_ut!(integrator::DEIntegrator, u, t)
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
PR to the integrator docs as well.

AayushSabharwal:
I've added the test to `downstream/integrator_indexing.jl` for now. Should I let it be there or move it somewhere else?

AayushSabharwal:
Shouldn't rebuilding the integrator docs just pick up the additional docstring for `set_u!` or is there something else I should add as well?

src/integrator_interface.jl
```diff
@@ -332,6 +332,28 @@ function set_u!(integrator::DEIntegrator, u)
     error("set_u!: method has not been implemented for the integrator")
 end
 
+"""
+    set_u!(integrator::DEIntegrator, sym, val)
+
+Set current state of symbol `sym` in `integrator` to `val`.
+"""
```
ChrisRackauckas, line 339: Make a `function set_u! end` and just put the docstrings together.

ChrisRackauckas (approved):

PR #211, opened 2022-08-04
depreciate vars in plot recipe for idxs

isaacsas:
I left the depreciation warning with `force = true` as it seems like this is something users should see (i.e. it doesn't only impact package developers who would see it in tests). 

I can turn it off though if that is preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 /Manifest.toml
 .*.swp
+
+# vscode stuff
+.vscode
+.vscode/*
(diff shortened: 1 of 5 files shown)


isaacsas:
The NonlinearSolve failure seems unrelated.

ChrisRackauckas:
Do a simultaneous docs PR?

isaacsas:
Realized in updating the docs I should update the integrator recipe too, but it was apparently broken from not being up to date with the solution plotting helper functions (I guess they were updated with a modified signature at some point but the integrator recipe didn't get updated too).

Issue #212, opened 2022-08-09
Many functions on `AbstractODESolution` return actual `ODESolution`s


nathanaelbosch:
https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/ode_solutions.jl defines a range of functions, which operate on `AbstractODESolution`s, and yet return `ODESolution`s. This is particularly confusing for cases such as `solution_new_retcode` or `solution_new_tslocation`. For example, if I implement my own `MySolution <: AbstractODESolution` then it silently gets transformed into a `ODESolution`. I would argue that defining these functions on `ODESolution` would therefore be more appropriate, so that an error is raised.

The relevant functions here are:
- build_solution
- solution_new_retcode
- solution_new_tslocation
- solution_slice
- sensitivity_solution

ChrisRackauckas:
Good point. These should probably be more restrictive.

nathanaelbosch:
Closed via #213.

PR #213, opened 2022-08-12
Restrict many ode_solutions.jl functions to actual `ODESolution`s

nathanaelbosch:
As stated in #212, many functions in `ode_solutions.jl` are defined for `AbstractODESolution` objects, yet they return actual `ODESolution`s. This PR restricts these functions to `sol::ODESolution`.

The PR also adds a simpler constructor that shortens the code quite a bit: Instead of providing *all* types, one can now build an ODESolution with just `ODESolution{T, N}(u, u_analytic, ...)`.Please let me know if this is a desired change or not! If not I'll just remove the second commit of this PR again.

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -218,7 +218,7 @@ function calculate_solution_errors!(sol::AbstractODESolution; fill_uanalytic = t
     end
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@chriselrod @YingboMa looks like there was a downstream break from LoopVectorization to ModelingToolkit: https://github.com/SciML/SciMLBase.jl/runs/7803028537?check_suite_focus=true#step:6:945

PR #215, opened 2022-08-15
SDE u_analytic with noise as NoiseFunction

rmsrosa:
This allows the computation of `sol.u_analytic` for problems of type `AbstractSDDEProblem` or `AbstractRODEProblem` with `noise` given as a `NoiseFunction`.

Previously, `sol.W[i]` would be called at https://github.com/SciML/SciMLBase.jl/blob/8bab6d3bc11c8db65d73875a2ea0f43bd854dd87/src/solutions/rode_solutions.jl#L118 regardless of being a proper vector-like `NoiseProcess` or a `NoiseFunction`, which would error for the latter. Checking whether `sol.W` is a `sol.W isa AbstractDiffEqArray{T, N, nothing} where {T, N}` and calling `first(sol.W(sol.t[i]))` instead fixes the problem.

I also make sure to `empty!(sol.u_analytic)` before filling it to avoid inadvertently doubling its size from a possible previous call to `calculate_solution_errors!(sol)`.

### DIFF
--- a/src/solutions/rode_solutions.jl
+++ b/src/solutions/rode_solutions.jl
@@ -114,8 +114,16 @@ function calculate_solution_errors!(sol::AbstractRODESolution; fill_uanalytic = 
     end
 
     if fill_uanalytic
-        for i in 1:length(sol)
-            push!(sol.u_analytic, f.analytic(sol.prob.u0, sol.prob.p, sol.t[i], sol.W[i]))
+        empty!(sol.u_analytic)
+        if sol.W isa AbstractDiffEqArray{T, N, nothing} where {T, N}
+            for i in 1:length(sol)
+            
+                push!(sol.u_analytic, f.analytic(sol.prob.u0, sol.prob.p, sol.t[i], first(sol.W(sol.t[i]))))
+            end
+        else
+            for i in 1:length(sol)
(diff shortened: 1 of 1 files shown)


src/solutions/rode_solutions.jl
```diff
@@ -114,8 +114,15 @@ function calculate_solution_errors!(sol::AbstractRODESolution; fill_uanalytic =
     end
 
     if fill_uanalytic
-        for i in 1:length(sol)
-            push!(sol.u_analytic, f.analytic(sol.prob.u0, sol.prob.p, sol.t[i], sol.W[i]))
+        empty!(sol.u_analytic)
```
ChrisRackauckas, line 117: when would this be non-empty?

rmsrosa: Not usually, but it might happen (it happened to me): issue #215

ChrisRackauckas:
format

rmsrosa:
I am trying, but I am getting this.
```julia
julia> JuliaFormatter.format(pkgdir(SciMLBase))
ERROR: AssertionError: currently .JuliaFormatter.toml accepts only "default" or "yas" or "blue" for the style configuration
```

rmsrosa:
ok, I updated JuliaFormatter and it worked.

PR #217, opened 2022-08-16
[Awaiting Review] PDESolution for MethodOfLines

xtalax:
Implements a PDESolution for MOL, with a basis to extend to other discrertizers.

Tested in MOL, and working.

This must be merged before it's sister pr in DiffEqBase
TODO:
- [x] Abstract type
- [x] PDESolution typedef and MOL constructor
- [x] Interception of sol construction and wrapping
- [x] Call methods
    - [x] ND Interpolations
- [x] `pdesol[u]`
- [x] Repeat for NonlinearSolution
- [x] tests
- [x] docs


xtalax:
A question that needs answering, how should getindex work for PDESolution? Its fairly clear for `sol[u]` `sol[t]` `sol[x]` `sol[u[10](t)]` etc., but less clear for what an index with an integer or list of integers/colons/ranges should mean.

src/SciMLBase.jl
```diff
@@ -367,6 +368,16 @@ $(TYPEDEF)
 """
 abstract type AbstractSDDEIntegrator{Alg, IIP, U, T} <: DEIntegrator{Alg, IIP, U, T} end
 
+# Discretizers
+"""
+$(TYPEDEF)
+
+## Type parameters:
+- `N` The number of discretized independent variables.
+- `M` The number of discretized dependent variables.
```
ChrisRackauckas, line 377: why do these need to be type parameters?

Why is this not an abstract array with time / without time?

xtalax: What should be an abstract array? They don't need to be type parameters, I just thought this made sense to expose.

src/SciMLBase.jl
```diff
@@ -4,6 +4,7 @@ using ConstructionBase
 using RecipesBase, RecursiveArrayTools, Tables
 using DocStringExtensions
 using LinearAlgebra
+using Interpolations
```
ChrisRackauckas, line 7: This shouldn't be in SciMLBase if we can avoid it.

ChrisRackauckas: What is the load time impact?

src/SciMLBase.jl
```diff
@@ -657,6 +673,10 @@ export AffineDiffEqOperator, DiffEqScaledOperator
 
 export DiffEqScalar, DiffEqArrayOperator, DiffEqIdentity
 
+export MOLMetadata
+
+export wrap_sol
```
ChrisRackauckas, line 678: isn't this private API?

src/discretizers/methodoflines_metadata.jl
```diff
@@ -0,0 +1,21 @@
+"""
+$(TYPEDEF)
+
+A type used to store data about a PDESystem, and how it was discretized by MethodOfLines.jl.
```
ChrisRackauckas, line 4: That doesn't sound Base. Isn't there a generalization here?

src/SciMLBase.jl
```diff
@@ -444,6 +452,11 @@ $(TYPEDEF)
 """
 abstract type AbstractDAESolution{T, N, S} <: AbstractODESolution{T, N, S} end
 
+"""
+$(TYPEDEF)
+"""
+abstract type AbstractPDESolution{T, N, S} <: AbstractTimeseriesSolution{T, N, S} end
```
ChrisRackauckas, line 458: Time dependent and time independent?

src/solutions/pde_solutions.jl
```diff
@@ -0,0 +1,190 @@
...
+                       typeof(interp)}(umap, sol, sol.errors, sol.t, ivgrid, ivs,
+                                       pdesys.dvs, metadata, sol.prob, sol.alg,
+                                       interp, sol.dense, sol.tslocation,
+                                       sol.retcode)
+end
+
+wrap_sol(sol::ODESolution, disc_data::MOLMetadata) = PDESolution(sol, disc_data)
+
+function (sol::PDESolution{T, N, S, D})(args...;
+                                        dv = nothing) where {T, N, S, D <: MOLMetadata}
```
ChrisRackauckas, line 93: This should just lower to a general metadata dispatch which would then be implemented in each PDE solver package. That would then move MOLMetadata to MethodOfLines.jl, and allow for NeuralPDEMetadata in NeuralPDE.jl

src/solutions/pde_solutions.jl
```diff
@@ -0,0 +1,190 @@
...
+                        out = zeros(T, length(sol.t), size(discu)...)
+                        for I in CartesianIndices(discu)
+                            out[:, I] .= solu[I]
+                        end
+                        u => out
+                    end
+                end)
+    # Build Interpolations
+    # TODO: Check if the grid is uniform and use the supported higher order interpolations
+    interp = Dict(map(keys(umap)) do k
```
ChrisRackauckas, line 70: Why is this build here? Or, why not insert a generic function call here?

src/solutions/pde_solutions.jl
```diff
@@ -0,0 +1,190 @@
...
+end
+
+function PDESolution(sol::ODESolution{T}, metadata::MOLMetadata) where {T}
+    odesys = sol.prob.f.sys
+
+    pdesys = metadata.pdesys
+    discretespace = metadata.discretespace
+
+    ivs = [discretespace.time, discretespace.x̄...]
+    ivgrid = (isequal(discretespace.time, x) ? sol.t : discretespace.grid[x] for x in ivs)
```
ChrisRackauckas, line 48: Why would a general PDE solution type have a "grid"? Domain.

src/SciMLBase.jl
xtalax: With:   
```
  5.509606 seconds (4.51 M allocations: 333.874 MiB, 0.22% gc time, 2.26% compilation time)

  0.000070 seconds (104 allocations: 8.281 KiB)

  0.000054 seconds (104 allocations: 8.281 KiB)

  0.000109 seconds (104 allocations: 8.281 KiB)

  0.000069 seconds (104 allocations: 8.281 KiB)
```

Without:
```
  3.993145 seconds (1.89 M allocations: 146.430 MiB, 0.34% gc time, 3.52% compilation time)

  0.000104 seconds (104 allocations: 8.281 KiB)

  0.000058 seconds (104 allocations: 8.281 KiB)

  0.000069 seconds (104 allocations: 8.281 KiB)

  0.000104 seconds (104 allocations: 8.281 KiB)
```

NB. Couldn't use btime since `using` has to be at top level, is there a better way to do this?

src/solutions/pde_solutions.jl
xtalax: @ChrisRackauckas Would a generic function call in SciMLBase call a function defined in MOL even if it does not use MOL, if the user was using MOL? Or do I misunderstand your point?

Where should this build go?

xtalax:
@

PR #224, opened 2022-08-21
Enable no recompilation by default for ODEs with Vector{Float64}

ChrisRackauckas:
This is a big move with a big potential payoff. Fingers crossed.


ChrisRackauckas:
All of the other failures are known v1.8 things (Enzyme and Sundials... gahhhh), so, here goes nothing.

PR #225, opened 2022-08-21
Use Tricks.jl so that isinplace is inferred at compile time

ChrisRackauckas:
Just playing around trying to make these things work. @yingboma @oxinabox for reference. Is this even possible? I think I can make the numbers into static values?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -4,7 +4,7 @@ $(SIGNATURES)
 Returns the number of arguments of `f` for each method.
 """
 function numargs(f)
-    return Tuple(num_types_in_tuple(m.sig) - 1 for m in Tricks.methods(f)) #-1 since f is the first parameter
+    return Tuple(num_types_in_tuple(m.sig) - 1 for m in methods(f)) #-1 since f is the first parameter
 end
 
 function numargs(f::RuntimeGeneratedFunctions.RuntimeGeneratedFunction{T, V, W, I}) where {
@@ -245,7 +245,7 @@ function isinplace(f, inplace_param_number, fname = "f", iip_preferred = true;
             # If so, no error
             for i in 1:length(nargs)
                 if nargs[i] < inplace_param_number &&
-                   any(isequal(Vararg{Any}), Tricks.methods(f).ms[1].sig.parameters)
+                   any(isequal(Vararg{Any}), methods(f).ms[1].sig.parameters)
                     # If varargs, assume iip
                     return iip_preferred
                 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -4,7 +4,7 @@ $(SIGNATURES)
 Returns the number of arguments of `f` for each method.
 """
 function numargs(f)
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -4,19 +4,19 @@ $(SIGNATURES)
...
+    return Tuple(num_types_in_tuple(m.sig) - 1 for m in Tricks.static_methods(f)) #-1 since f is the first parameter
 end
 
 """
 $(SIGNATURES)
 
 Get the number of parameters of a Tuple type, i.e. the number of fields.
 """
-function num_types_in_tuple(sig)
+Base.@pure function num_types_in_tuple(sig)
```
ChrisRackauckas, line 15: I probably lose any sense of inference in here. It probably needs hand rolled generated functions?

chriselrod:
```julia
julia> f(x, y, z) = z + y - z

julia> f!(w, x, y, z) = @. w = x * y - z
f! (generic function with 1 method)

julia> static_hasmethod(f, Tuple{Vararg{Any,4}})
false

julia> static_hasmethod(f!, Tuple{Vararg{Any,4}})
true
```

ChrisRackauckas:
```julia
            # Possible extra safety?
            # Find if there's a `f(args...)` dispatch
            # If so, no error
            for i in 1:length(nargs)
                if nargs[i] < inplace_param_number &&
                   any(isequal(Vararg{Any}), methods(f).ms[1].sig.parameters)
                   any(isequal(Vararg{Any}), Tricks.static_methods(f).ms[1].sig.parameters)
                    # If varargs, assume iip
                    return iip_preferred
                end
            end
```

is there something in Tricks for `has_vararg_method`?

oxinabox:
>  Is this even possible? I think I can make the numbers into static values?

Seems possible. I see no issues so far

ChrisRackauckas:
Some of the errors on ModelingToolkit look really weird... I think there's a bug with Tricks.jl and RuntimeGeneratedFunctions

oscardssmith:
@vtjnash was saying that Tricks.jl is a little buggy.

ChrisRackauckas:
Somehow, Tricks.jl doesn't play nicely with `Core.return_type` as evidenced by the failure in Downstream 😅 . I don't think it's quite ready for what we need.

ChrisRackauckas:
I think we'll wait until @vtjnash adds something similar to base Julia. This seems to be uncovering more than a few bugs when compared to the fully working tests on master.

oxinabox:
If you have time it would be nice to document this on Tricks.jl, and what the failure cases were.

oxinabox:
According to @vtjnash (when I made Tricks.jl) the correct way to do this was to add a `tfunc` for `hasmethod`, `isapplicable` and `methods` etc. 
And that it was "easy", it was just he wasn't convinced people would use it to write good code. He thought the code it might make would be too hard to reason about.
and  was going to argue with him about it at JuliaCon and then a pandemic happened

oscardssmith:
yeah, Chris and I have convinced him that we need to add it, now someone just has to bother doing so.

PR #227, opened 2022-08-23
extend summary of NoiseProblem to NoiseFunction

rmsrosa:
Allow a `NoiseProblem` to be created from a `NoiseFunction`.

Currently, `summary` for an `AbstractNoiseProblem` requires `prob.noise.W[1]`, which is fine for a `NoiseProcess` and others but not for `NoiseFunction`. This simply changes it `prob.noise.curW`.


rmsrosa:
Sorry, I made the PR from the wrong place.

PR #230, opened 2022-08-26
add unwrapping specializations for scimlfunctions

ChrisRackauckas:


ChrisRackauckas:
To do this correctly, we need to store whether the user chose to recompile or whether it was done automatically.

PR #233, opened 2022-08-29
Rode analytic with full `sol`

rmsrosa:
When solving a linear RODE `dX/dt = g(t, W_t) X` , where `W_t` is a Wiener process or some other noise, the exact solution is `X_t = X_0 exp(\int_0^t g(W_s) ds)`. Similarly for a non-homogeneous linear RODE `dX/dt = h(W_t) + g(W_t) X`. Problem is that `u_analytic` receives, at a given time t, only `W_t` not the whole history `W_s` for `0 \leq s \leq t`. This PR adds an alternative `Bool` field `analytic_full` to `RODEFunction` that, when set to `true`, `solve` passes the whole `sol` to `f.analytic`. The user is responsible for mutating `sol.u_analytic` themselves, from `sol`.

Here is an example I am adding to `BenchmarkTools.jl`  (and a more generic version to `SDEProblemLibrary.jl`).
```julia
g(t, W) = sin(W)
f(u, t, p, W) = g(t, W) * u
u0 = 1.0
tspan = (0.0, 1.0)

function u_analytic!(sol)
    empty!(sol.u_analytic)
    integral = 0.0
    push!(sol.u_analytic, u0)
    for i in 2:length(sol)
        integral += (g(sol.W.t[i], sol.W.W[i]) + g(sol.W.t[i-1], sol.W.W[i-1])) * (sol.W.t[i] - sol.W.t[i-1]) / 2
        push!(sol.u_analytic, u0 * exp(integral))
    end
end

ff = RODEFunction(f, analytic=u_analytic!, analytic_full=true)
prob = RODEProblem(ff, u0, tspan)

sol = solve(prob, RandomHeun(), dt=1/100)
plot(sol)
plot!(sol.t, sol.u_analytic)
```

plot_16

At first, I thought of just passing `sol.W` to `f.analytic(u, p, t, sol.W)` for constructing `sol.u_analytic[i]`, but this amounts to computing the whole integral at each step `i` instead of just updating `int_{t0}^{t-dt}` to `\int_{t0}^t`, being highly inefficient. So, I opted for this less elegant but much more efficient way.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -1180,7 +1180,7 @@ struct RODEFunction{iip, F, TMM, Ta, Tt, TJ, JVP, VJP, JP, SP, TW, TWt, TPJ, S, 
     observed::O
     colorvec::TCV
     sys::SYS
-    analytic_with_full_noise::FN
+    analytic_full::FN
 end
 
 """
@@ -2387,7 +2387,7 @@ function RODEFunction{iip, true}(f;
                                             DEFAULT_OBSERVED,
                                  colorvec = __has_colorvec(f) ? f.colorvec : nothing,
                                  sys = __has_sys(f) ? f.sys : nothing,
-                                 analytic_with_full_noise = __has_analytic_with_full_noise(f) = f.analytic_with_full_noise : nothing) where {iip}
+                                 analytic_full = __has_analytic_full(f) = f.analytic_full : nothing) where {iip}
     if jac === nothing && isa(jac_prototype, AbstractDiffEqLinearOperator)
         if iip
             jac = update_coefficients! #(J,u,p,t)
@@ -2427,10 +2427,10 @@ function RODEFunction{iip, true}(f;
                  typeof(jac), typeof(jvp), typeof(vjp), typeof(jac_prototype),
                  typeof(sparsity), typeof(Wfact), typeof(Wfact_t),
                  typeof(paramjac), typeof(syms), typeof(observed), typeof(_colorvec),
-                 typeof(sys), typeof(analytic_with_full_noise)}(f, mass_matrix, analytic, tgrad,
+                 typeof(sys), typeof(analytic_full)}(f, mass_matrix, analytic, tgrad,
                               jac, jvp, vjp, jac_prototype, sparsity,
                               Wfact, Wfact_t, paramjac, syms, observed,
-                              _colorvec, sys, analytic_with_full_noise)
+                              _colorvec, sys, analytic_full)
 end
 function RODEFunction{iip, false}(f;
                                   mass_matrix = __has_mass_matrix(f) ? f.mass_matrix : I,
@@ -2450,7 +2450,7 @@ function RODEFunction{iip, false}(f;
                                              DEFAULT_OBSERVED,
                                   colorvec = __has_colorvec(f) ? f.colorvec : nothing,
                                   sys = __has_sys(f) ? f.sys : nothing,
-                                  analytic_with_full_noise = __has_analytic_with_full_noise(f) = f.analytic_with_full_noise : nothing) where {iip}
+                                  analytic_full = __has_analytic_full(f) = f.analytic_full : nothing) where {iip}
     if jac === nothing && isa(jac_prototype, AbstractDiffEqLinearOperator)
         if iip
             jac = update_coefficients! #(J,u,p,t)
@@ -2491,7 +2491,7 @@ function RODEFunction{iip, false}(f;
                                                             jac, jvp, vjp, jac_prototype,
                                                             sparsity, Wfact, Wfact_t,
                                                             paramjac, syms, observed,
-                                                            _colorvec, sys, analytic_with_full_noise)
+                                                            _colorvec, sys, analytic_full)
 end
 function RODEFunction{iip}(f; kwargs...) where {iip}
     RODEFunction{iip, RECOMPILE_BY_DEFAULT}(f; kwargs...)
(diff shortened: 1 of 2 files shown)


rmsrosa:
I updated the docstring for `RODEFunction` (it seems it was still a copy and paste of `ODEFunction`). I have not worked with all the parameters, but I hope what I wrote is correct.

I also added some tests concerning building `RODEFunction`. There were none before.

I think it is all set, now (apart from some failing checks which I believe were there before and are not related to what I did). Let me know if I need to do something else.

src/scimlfunctions.jl
```diff
@@ -1126,33 +1126,33 @@ RODEFunction{iip,recompile}(f;
...
+  to determine that the equation is actually a random differential-algebraic equation (RDAE)
+  if `M` is singular. Currently, however, all solvers assume `M` is the identity and do not take this term into account. Special solvers still need to be written to use a non-identity mass matrix.
+- `analytic`: (u0,p,t,W)` or `analytic(sol)`: used to pass an analytical solution function for the analytical
+  solution of the RODE. Generally only used for testing and development of the solvers. The exact form depends on the field `analytic_full`.
+- `analytic_full`: a boolean to indicate whether to use the form `analytic(u0,p,t,W)` (if `false`) or the form `analytic!(sol)` (if `true`). The former is expected to return the solution `u(t)` of the equation, given the initial condition `u0`, the parameter `p`, the current time `t` and the value `W=W(t)` of the noise at the given time `t`. The latter case is useful when the solution of the RODE depends on the whole history of the noise, which is available in `sol.W.W`, at times `sol.W.t`. In this case, `analytic(sol)` must mutate explicitly the field `sol.u_analytic` with the corresponding expected solution at `sol.W.t` or `sol.t`.
```
ChrisRackauckas, line 1145: This is far past the line length 😅

ChrisRackauckas, line 1145:
```suggestion
- `analytic_full`: a boolean to indicate whether to use the form `analytic(u0,p,t,W)` (if `false`) 
  or the form `analytic!(sol)` (if `true`). The former is expected to return the solution `u(t)` of 
  the equation, given the initial condition `u0`, the parameter `p`, the current time `t` and the 
  value `W=W(t)` of the noise at the given time `t`. The latter case is useful when the solution 
  of the RODE depends on the whole history of the noise, which is available in `sol.W.W`, at 
  times `sol.W.t`. In this case, `analytic(sol)` must mutate explicitly the field `sol.u_analytic` 
  with the corresponding expected solution at `sol.W.t` or `sol.t`.
```

src/scimlfunctions.jl
```diff
@@ -1126,33 +1126,39 @@ RODEFunction{iip,recompile}(f;
...
-  Must be an AbstractArray or an AbstractSciMLOperator.
-- `analytic(u0,p,t)`: used to pass an analytical solution function for the analytical
-  solution of the ODE. Generally only used for testing and development of the solvers.
-- `tgrad(dT,u,p,t)` or dT=tgrad(u,p,t): returns ``\frac{\partial f(u,p,t)}{\partial t}``
-- `jac(J,u,p,t)` or `J=jac(u,p,t)`: returns ``\frac{df}{du}``
-- `jvp(Jv,v,u,p,t)` or `Jv=jvp(v,u,p,t)`: returns the directional derivative``\frac{df}{du} v``
-- `vjp(Jv,v,u,p,t)` or `Jv=vjp(v,u,p,t)`: returns the adjoint derivative``\frac{df}{du}^\ast v``
+- `mass_matrix`: the mass matrix `M` represented in the RODE function. Can be used
+  to determine that the equation is actually a random differential-algebraic equation (RDAE)
+  if `M` is singular. Currently, however, all solvers assume `M` is the identity and do not take this term into account. Special solvers still need to be written to use a non-identity mass matrix.
```
ChrisRackauckas, line 1142:
```suggestion
  if `M` is singular.
```

"Documentation at a distance" is hard to maintain, so we generally don't do that. The solvers should just error if it's not supported (I would think they already do? Worth checking)

src/scimlfunctions.jl
```diff
@@ -1126,33 +1126,39 @@ RODEFunction{iip,recompile}(f;
...
-  solution of the ODE. Generally only used for testing and development of the solvers.
-- `tgrad(dT,u,p,t)` or dT=tgrad(u,p,t): returns ``\frac{\partial f(u,p,t)}{\partial t}``
-- `jac(J,u,p,t)` or `J=jac(u,p,t)`: returns ``\frac{df}{du}``
-- `jvp(Jv,v,u,p,t)` or `Jv=jvp(v,u,p,t)`: returns the directional derivative``\frac{df}{du} v``
-- `vjp(Jv,v,u,p,t)` or `Jv=vjp(v,u,p,t)`: returns the adjoint derivative``\frac{df}{du}^\ast v``
+- `mass_matrix`: the mass matrix `M` represented in the RODE function. Can be used
+  to determine that the equation is actually a random differential-algebraic equation (RDAE)
+  if `M` is singular.
+- `analytic`: (u0,p,t,W)` or `analytic(sol)`: used to pass an analytical solution function for the analytical
+  solution of the RODE. Generally only used for testing and development of the solvers. The exact form depends on the field `analytic_full`.
```
ChrisRackauckas, line 1144:
```suggestion
  solution of the RODE. Generally only used for testing and development of the solvers. 
  The exact form depends on the field `analytic_full`.
```

rmsrosa:
Oh, thanks for taking care of the docstring instead of going thru a ping pong! 😅

PR #234, opened 2022-08-30
add Hessian of Lagrangian to OptimizationFunction

baggepinnen:
This PR adds the possibility of adding a Hessian of the Lagrangian to `OptimizationSystem`, i.e., a function that directly computes
$\dfrac{d^2}{dx^2} \big( \sigma f(x) + \lambda^T c(x) \big)$
rather than assembling this from individual constraint Hessians.

Many solvers, including Ipopt, MadNLP.jl and `Optim.IPNewton` want this function.

ref
- https://github.com/SciML/Optimization.jl/issues/343


Vaibhavdixit02 (approved):
Looks good. Thanks for handling it.

PR #246, opened 2022-09-09
PDESolutions

xtalax:
Adds a solution wrapper type system for PDEs, and a solution wrapping function, to be extended in discretizer packages.

Must be merged BEFORE https://github.com/SciML/DiffEqBase.jl/pull/825


src/solutions/basic_solutions.jl
```diff
@@ -77,3 +77,15 @@ function build_solution(prob::AbstractIntegralProblem,
                      typeof(prob), typeof(alg), typeof(chi)}(u, resid, prob, alg, retcode,
                                                              chi)
 end
+
+function wrap_sol(sol)
+    sys_type = if sol.prob.f.sys !== nothing
+        sol.prob.f.sys.metadata
```
ChrisRackauckas, line 83: shouldn't this just be done via `prob.problem_type` information?

xtalax: See https://github.com/SciML/DiffEqBase.jl/pull/825

ChrisRackauckas: why wait to change that?

xtalax: Done

PR #247, opened 2022-09-12
cODEFunction -> ODEFunction (bug left from removing convert).

hustf:
modified:   src/problems/bvp_problems.jl cODEFunction -> ODEFunction

Not covered by tests. 
```
#Working:
BVProblem(ode!, bc!, initial_guess, tspan, p)
# Will also work with this PR:
BVProblem{true}(ode!, bc!, initial_guess, tspan, p)
```
I'm not sure when and how enabling compilation affects start-up-time, but it could be a side effect of this fix.


ChrisRackauckas:
Added downstream testing to BoundaryValueDiffEq.jl

PR #250, opened 2022-09-13
Update dae_problems.jl

thomvet:
Typo in error message recently introduced.


ChrisRackauckas:
Thanks! Yeah, sucks that the error message itself is hard to test 😓

PR #251, opened 2022-09-14
DiffEqBase Hotfix for PDESolutions

xtalax:


xtalax:
Alternatively I can add problem_type to all problems

PR #252, opened 2022-09-15
Allow for setting the default specialization via Preferences

ChrisRackauckas:


src/scimlfunctions.jl
```diff
@@ -6,6 +6,9 @@ struct NoSpecialize <: AbstractSpecialization end
 struct FunctionWrapperSpecialize <: AbstractSpecialization end
 struct FullSpecialize <: AbstractSpecialization end
 
+const DEFAULT_SPECIALIZATION = Preferences.@load_preference("SpecializationLevel",
+                                                            AutoSpecialize)
```
devmotion, line 10: Does this actually work for non-default values? Can one load types from the TOML file? I assumed it would only support values that are supported by TOML, such as e.g. booleans, numbers, or strings.

ChrisRackauckas: Oh interesting... I guess we'd need to parse the string into a type definition?

PR #253, opened 2022-09-15
Indexing of parameters

### LINKED ISSUE #231: Integrator symbolic indexing should support parameters
`getindex(A::DEIntegrator, sym)` can flexibly return states, independent variable, and observed variables, but not system parameters. 

I would like subtypes of `AbstractDiffEqFunction{iip}` (those used by MTK) to have a kwarg and field equivalent to `syms`, called `paramsyms`. That would allow modifying the `getindex` method to support all possible system symbols. The changes here would need to be propagated to the contructor calls in `ModelingToolkit.jl`

https://github.com/SciML/SciMLBase.jl/blob/f348d7c377aaa8006069971f5dc3bba6b89fabce/src/integrator_interface.jl#L415-L437

AayushSabharwal:
Closes #231

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -49,7 +49,7 @@ $(TYPEDEF)
 abstract type AbstractODEFunction{iip} <: AbstractDiffEqFunction{iip} end
 
 @doc doc"""
-    ODEFunction{iip,F,TMM,Ta,Tt,TJ,JVP,VJP,JP,SP,TW,TWt,TPJ,S,S2,O,TCV} <: AbstractODEFunction{iip,recompile}
+    ODEFunction{iip,F,TMM,Ta,Tt,TJ,JVP,VJP,JP,SP,TW,TWt,TPJ,S,S2,S3,O,TCV} <: AbstractODEFunction{iip,recompile}
 
 A representation of an ODE function `f`, defined by:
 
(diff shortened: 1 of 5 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
This looks good now, thanks for sticking with it.

AayushSabharwal:
Thanks for reviewing my several failed attempts at rebasing 😅

I should mention, the indexing tests won't pass since the change from MTK isn't merged yet. For tests to work, we'll need this to be merged, and then see the results in the PR in MTK. Both repositories would also need to bump their compat versions for each other.

ChrisRackauckas:
Well this needs to be setup to have working tests before merging, so for now disable the tests that require ModelingToolkit.

oscardssmith:
One thing I'm worried about with this is won't it break the code of anyone who relies on the type parameters being in a specific order? Would it be be better to move the extra type parameter to the end?

AayushSabharwal:
That's actually a fair point. I'll shuffle stuff around if @ChrisRackauckas also thinks it should be that way.

src/scimlfunctions.jl
```diff
@@ -1971,6 +2060,10 @@ function SplitFunction{iip, recompile}(f1, f2;
                                        paramjac = __has_paramjac(f1) ? f1.paramjac :
                                                   nothing,
                                        syms = __has_syms(f1) ? f1.syms : nothing,
+                                       indepsym = __has_indepsym(f1) ? f1.indepsym :
+                                                  nothing,
+                                       paramsyms = __has_paramsyms(f) ? f.paramsyms :
```
ChrisRackauckas, line 2065:
```suggestion
                                       paramsyms = __has_paramsyms(f1) ? f1.paramsyms :
```

ChrisRackauckas:
Format.

I don't think those type parameters are breaking because no one is writing that specific of a dispatch? There is no recommended constructor that uses them.

PR #262, opened 2022-09-23
remake with symbols infrastructure

xtalax:
Infrastructure for remaking with symbols where appropriate

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -48,21 +48,18 @@ function remake(prob::ODEProblem; f = missing,
                 p = missing,
                 kwargs = missing,
                 _kwargs...)
-    if u0 === missing
-        u0 = prob.u0
-    else
-        if u0 isa Dict || eltype(u0) isa Pair
-            if hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :states)
-                u0 = handle_varmap(u0, prob.f.sys, field = :states,
-                                   defaults = getfield(prob.f.sys, :states) .=> prob.u0)
-            else
-                throw(ArgumentError("This problem does not support symbolic default maps with `remake`, i.e. it does not have a symbolic origin. Please use `remake` with the `u0` keyword argument as a vector of values, paying attention to the order of states."))
-            end
-        end
-    end
 
     if tspan === missing
         tspan = prob.tspan
+    end
+    defs = Dict([])
+    if hasproperty(prob.f, :sys)
+        if hasfield(typeof(prob.f.sys), :ps)
+            defs = mergedefaults(defs, prob.p, getfield(prob.f.sys, :ps))
+        end
+        if hasfield(typeof(prob.f.sys), :u0)
+            defs = mergedefaults(defs, prob.u0, getfield(prob.f.sys, :u0))
+        end
     end
 
     if p === missing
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This isn't split appropriately and will heavily invalidate. Instead, make this drop the arg `prob.f.sys` appropriately (checking for existence etc.) in a subfunction, and extend that on a dispatch for `AbstractSystem`

src/remake.jl
```diff
@@ -59,7 +59,13 @@ function remake(prob::ODEProblem; f = missing,
     if p === missing
         p = prob.p
     else
-        p = handle_varmap(p, prob)
+        @show typeof(prob.f.sys)
+        @show typeof(prob.f.sys.ps)
+        if hasproperty(prob.f, :sys) && hasproperty(prob.f.sys, :ps)
+            p = handle_varmap(p, prob.f.sys)
```
ChrisRackauckas, line 65: you're missing the basic dispatch here

src/remake.jl
```diff
@@ -58,6 +58,15 @@ function remake(prob::ODEProblem; f = missing,
 
     if p === missing
         p = prob.p
+    else
+        @show typeof(prob.f.sys)
+        @show hasproperty(prob.f, :sys)
+        @show hasproperty(prob.f.sys, :ps)
+        if hasproperty(prob.f, :sys) && hasproperty(prob.f.sys, :ps)
```
YingboMa, line 65: We should handle `u0` as well.

ChrisRackauckas (approved):

ChrisRackauckas:
Format

YingboMa (changes requested):
`prob.f.sys.xxx` is wrong. getproperty of a system only means to get the variable in the system. You should call `getfield` directly.

src/remake.jl
```diff
@@ -50,6 +50,14 @@ function remake(prob::ODEProblem; f = missing,
                 _kwargs...)
     if u0 === missing
         u0 = prob.u0
+    else
+        if u0 isa Dict || eltype(u0) isa Pair
+            if hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :states)
+                u0 = handle_varmap(u0, prob.f.sys, field = :states, defaults = prob.f.sys.defaults)
```
YingboMa, line 56: Shouldn't the default be constructed from the current states and parameters that the problem holds?

src/remake.jl
```diff
@@ -58,6 +66,14 @@ function remake(prob::ODEProblem; f = missing,
 
     if p === missing
         p = prob.p
+    else
+        if p isa Dict || eltype(p) isa Pair
+            if hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :ps)
+                p = handle_varmap(p, prob.f.sys, field = :ps, defaults = prob.f.sys.ps .=> prob.p)
```
YingboMa, line 72: This parameters could depend on the initial condition, so you should include `prob.u0` map as well.

src/remake.jl
```diff
@@ -50,6 +50,15 @@ function remake(prob::ODEProblem; f = missing,
                 _kwargs...)
     if u0 === missing
         u0 = prob.u0
+    else
+        if u0 isa Dict || eltype(u0) isa Pair
+            if hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :states)
+                u0 = handle_varmap(u0, prob.f.sys, field = :states,
+                                   defaults = getfield(prob.f.sys, :states) .=> prob.u0)
```
ChrisRackauckas, line 57: this should also have p

PR #271, opened 2022-10-09
Adds field indicating integrality of the opt vars

### LINKED ISSUE #270: Add `int` field to `OptimizationProblem`
A field indicating which variables have to fulfil an integrality constraint. If the bounds are `[0,1]` it is a binary variable.

ValentinKaisermayer:
closes #270

### DIFF
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -313,6 +313,7 @@ Any extra keyword arguments are captured to be sent to the optimizers.
 * `p`: the parameters for the problem. Defaults to `NullParameters`.
 * `lb`: the lower bounds for the optimization of `u`.
 * `ub`: the upper bounds for the optimization of `u`.
+* `int`: integrality indicator for `u`.
 * `lcons`: the vector of lower bounds for the constraints passed to [OptimizationFunction](@ref).
     Defaults to `nothing`, implying no lower bounds for the constraints (i.e. the constraint bound is `-Inf`)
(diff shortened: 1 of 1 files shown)


Vaibhavdixit02 (approved):
Looks good, but I am curious about how you plan to use it. Do you have an example you tested this out on that you can show?

ValentinKaisermayer:
> Looks good, but I am curious about how you plan to use it. Do you have an example you tested this out on that you can show?

I'm working on it, see https://github.com/SciML/ModelingToolkit.jl/issues/1786

ValentinKaisermayer:
For instance, now it is trivial to add support for it with `NOMAD.jl`, in:
https://github.com/SciML/Optimization.jl/blob/b67636eeb5ee854fa53569e67bff3a21c5ed7a97/lib/OptimizationNOMAD/src/OptimizationNOMAD.jl#L80

Depending on the variable bounds `lb` and `ub` and the integrality indicator `int` the variable types can be inferred and are either `"R"`, `"B"` or `"I"`.

`MTK` got support for binary and integer variable metadata in https://github.com/SciML/ModelingToolkit.jl/pull/1798.
And I'm planning to add support for other MOI solvers like `CBC`, or `Gurobi`, that support such variable types in  https://github.com/SciML/Optimization.jl/pull/381.

It should be just a matter of connecting all of it now. 😃

Vaibhavdixit02:
I see, cool!

PR #272, opened 2022-10-09
Fix integrator parameter indexing, enable tests

### LINKED ISSUE #273: Indexing of integrator/solution with collection doesn't return same type as passed in
https://github.com/SciML/SciMLBase.jl/blob/1ecf19e8404cb3e043520c0515483768c6bed0f7/src/integrator_interface.jl#L426

Why is there a `collect` here? If `sym isa AbstractArray`, collecting it just forces it to be a `Vector` which means a `Vector` is always returned. It's impossible to pass in an `SVector` and get back an `SVector`

AayushSabharwal:
- Indexing integrators with parameters was broken, fixed it
- This wasn't caught by tests because the tests were commented. When the feature was introduced it needed changes in MTK and would've failed CI.
- Closes #273

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -421,11 +421,10 @@ Base.@propagate_inbounds function Base.getindex(A::DEIntegrator,
 end
 
 Base.@propagate_inbounds function Base.getindex(A::DEIntegrator, sym)
-    if issymbollike(sym) || all(issymbollike, sym)
-        if sym isa AbstractArray
-            return map(s -> A[s], collect(sym))
-        end
+    if issymbollike(sym)
         i = sym_to_index(sym, A)
+    elseif all(issymbollike, sym)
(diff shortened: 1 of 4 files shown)


AayushSabharwal:
https://github.com/SciML/SciMLBase.jl/blob/1ecf19e8404cb3e043520c0515483768c6bed0f7/src/integrator_interface.jl#L426

Why is there a `collect` here? If `sym isa AbstractArray`, collecting it just forces it to be a `Vector` which means a `Vector` is always returned. It's impossible to pass in an `SVector` and get back an `SVector`

AayushSabharwal:
I'd also like to be able to do something about https://github.com/SciML/SciMLBase.jl/issues/273 here if possible.

AayushSabharwal:
What does this need to get merged 😅 I'm kind of waiting on this for (possibly) decent performance boosts

PR #281, opened 2022-10-13
Fix typos

zmoon:
Found and corrected some typos with the help of `codespell`. Could add to CI, would just have to figure out how to configure it to ignore `ND`, which it wants to correct to `AND` or `2ND`.


ChrisRackauckas:
oh wow, I had never heard of codespell. Thanks! This is really awesome!

zmoon:
If you are interested in adding <https://github.com/codespell-project/actions-codespell>,
looks like this would work:

```yaml
uses: codespell-project/actions-codespell@master
with:
  ignore_words_list: nd
```

ChrisRackauckas:
Thanks! we'll look into adding that across the org.

PR #286, opened 2022-10-21
changes interface to use cache

ValentinKaisermayer:


src/solutions/optimization_solutions.jl
```diff
@@ -17,25 +17,25 @@ Representation of the solution to an nonlinear optimization defined by an Optimi
   callback (`sol.retcode === :Terminated`), or whether it exited due to an error. For more
   details, see the return code section of the DifferentialEquations.jl documentation.
 """
-struct OptimizationSolution{T, N, uType, P, A, Tf, O} <: AbstractOptimizationSolution{T, N}
+struct OptimizationSolution{T, N, uType, C, A, Tf, O} <: AbstractOptimizationSolution{T, N}
     u::uType # minimizer
-    prob::P # optimization problem
+    cache::C # optimization cache
```
ChrisRackauckas, line 22: I'm not so sure about this, since that is an interface break. We'd need to do that to all solution types.

ValentinKaisermayer: `LinearSolution` does this already
https://github.com/SciML/SciMLBase.jl/blob/b655d0657f9aab56a7619aa0584c2ef569d670f6/src/solutions/basic_solutions.jl#L28

But we can stick the cache definition in here or an abstract cache type and dispatch on it and leave the method name as it is.

ChrisRackauckas: interesting, okay then go for it. It might have some memory consequences though?

src/solutions/optimization_solutions.jl
```diff
@@ -64,3 +64,7 @@ function Base.summary(io::IO, A::AbstractOptimizationSolution)
           type_color, eltype(A.u),
           no_color)
 end
+
+function observed(A::AbstractOptimizationSolution, sym)
+    getobserved(A)(sym, A.u, A.cache.p)
+end
```
ChrisRackauckas, line 70: Does this not just work by the abstractnotimesolution fallbacks already? I don't think this is correct.

ValentinKaisermayer: This?
https://github.com/SciML/SciMLBase.jl/blob/3a2c5ed100f7dc76ecce8233e2e377ac917fd158/src/solutions/solution_interface.jl#L137

ChrisRackauckas: oh it's a prob vs cache thing? Can we put this there so it's more clear?

ChrisRackauckas (approved):

ChrisRackauckas:
Setup a deprecation path using getproperty overloading?

src/solutions/optimization_solutions.jl
```diff
@@ -17,28 +17,43 @@ Representation of the solution to an nonlinear optimization defined by an Optimi
...
     N = ndims(u)
 
-    OptimizationSolution{T, N, typeof(u), typeof(prob), typeof(alg),
-                         typeof(minimum), typeof(original)}(u, prob, alg, minimum, retcode,
+    OptimizationSolution{T, N, typeof(u), typeof(cache), typeof(alg),
+                         typeof(minimum), typeof(original)}(u, cache, alg, minimum, retcode,
                                                             original)
 end
 
+function build_solution(prob::AbstractOptimizationProblem, args...; kwargs...)
```
ChrisRackauckas, line 50: why not just be adding a dispatch here like all of the other problems? Why does it need a separate function?

src/solutions/optimization_solutions.jl
```diff
@@ -17,28 +17,43 @@ Representation of the solution to an nonlinear optimization defined by an Optimi
...
 
-function build_solution(prob::AbstractOptimizationProblem,
-                        alg, u, minimum;
-                        retcode = ReturnCode.Default,
-                        original = nothing,
-                        kwargs...)
+function Base.getproperty(sol::OptimizationSolution, sym::Symbol)
+    if sym == :prob
+        Base.depwarn("`sol.prob` is deprecated. Use `sol.cache` instead.", "sol.prob")
+        return getfield(sol, :cache)
```
ChrisRackauckas, line 32: shouldn't this be `sol.cache.prob`?

ValentinKaisermayer: The cache does not hold the problem. It think it would be hard to keep in sync. But if you think it should, we can add it. 
Maybe @Vaibhavdixit02 has an additional insight.

src/solutions/optimization_solutions.jl
ValentinKaisermayer: I will add an `AbstractOptimizationCache` and dispatch on it?

src/solutions/optimization_solutions.jl
ValentinKaisermayer: I think we have to ditch the `sol.prob.p` or `sol.cache.p` altogether and have to use a `get_p` interface.

For the `MOI` solvers there needs to be a `MOI.AbstractNLPEvaluator` which hold `p` but we need a cache that is a `SciMLBase.AbstractOptimizationCache`. This can only be solved via compostion, that the cache holds the evaluator. 


```julia
mutable struct MOIOptimizationNLPEvaluator{T, F <: OptimizationFunction, uType, P, LB, UB, I,
    JT <: DenseOrSparse{T}, HT <: DenseOrSparse{T},
    CHT <: DenseOrSparse{T}, S} <: MOI.AbstractNLPEvaluator
    f::F
    u0::uType
    p::P
    lb::LB
    ub::UB
    int::I
    J::JT
    H::HT
    cons_H::Vector{CHT}
    lcons::Vector{T}
    ucons::Vector{T}
    sense::S
end

mutable struct MOIOptimizationNLPCache{E, O, OS} <: SciMLBase.AbstractOptimizationCache
    evaluator::E
    opt::O
    opt_setup::OS
end
```

src/cache_interface.jl
```diff
@@ -1,15 +1,7 @@
...
-function get_syms(::AbstractOptimizationCache)
-    error("`get_syms`: method has not been implemented for the cache")
-end
-function get_paramsyms(::AbstractOptimizationCache)
-    error("`get_paramsyms`: method has not been implemented for the cache")
-end
+get_p(cache::AbstractOptimizationCache) = cache.p
+get_observed(cache::AbstractOptimizationCache) = cache.f.observed
+get_syms(cache::AbstractOptimizationCache) = cache.f.syms
+get_paramsyms(cache::AbstractOptimizationCache)= cache.f.paramsyms
```
ChrisRackauckas, line 4: these aren't part of the interface. If they are to be added to the interface, it needs to be done correctly.

ValentinKaisermayer: Sry! How should I do it then?

ChrisRackauckas:
Please don't mix PRs. If you're going to add a whole `cache` interface, it needs to be done to all problem types and it needs to be in https://scimlbase.sciml.ai/dev/interfaces/Solutions/. Caches are somewhat weird because they are problem-specific in their nature because they are exposing internals. So we would need to come up with a whole thing in https://scimlbase.sciml.ai/dev/interfaces/Init_Solve/#init-and-the-Iterator-Interface, and then document the cache handling in Optimization.jl. But half-doing it is worse than not doing it.

ChrisRackauckas:
Let's do one PR to make it a cache, and another more planned PR that adds a whole cache interface documentation page, puts the recommendations in a way that covers the current ODE solver integrator and LinearSolve cache, and then describes how those should all be documented, makes sure DiffEq and LinearSolve follow it, and then Optimization matches all of that.

ValentinKaisermayer:
What do you mean by all problem types? Not just the `OptimizationProblem`?

ChrisRackauckas:
If it's a SciMLBase interface, it should be defined for https://scimlbase.sciml.ai/stable/#Domains-of-SciML

ValentinKaisermayer:
Thats a big task. 

The `LinearCache` is not even a subtype of an abstract cache type in `SciMLBase`.
https://github.com/SciML/LinearSolve.jl/blob/43a66899d5a9a8164cbd0f8215e2fbd447d1753b/src/common.jl#L11

Nor follows its solution type the interface in https://scimlbase.sciml.ai/dev/interfaces/Solutions/#Common-Field-Names:
https://github.com/SciML/SciMLBase.jl/blob/2269c506b5ae932bcd14a1360ad2899632f52202/src/solutions/basic_solutions.jl#L22-L29

ValentinKaisermayer:
The problem is not that we need a cache interface but the solution has no interface definition but is just based on common fields, which already `LinearSolution` does not comply with. So we need an actual solution interface that allows for some flexibility since the caches will be too different.

Like this

```julia
get_p(sol::OptimizationSolution) = sol.cache.p
get_observed(sol::OptimizationSolution) = sol.cache.f.observed
get_syms(sol::OptimizationSolution) = sol.cache.f.syms
get_paramsyms(sol::OptimizationSolution)= sol.cache.f.paramsyms

has_observed(sol::OptimizationSolution) = get_observed(sol) !== nothing
has_syms(sol::OptimizationSolution) = get_syms(sol) !== nothing
has_paramsyms(sol::OptimizationSolution) = get_paramsyms(sol) !== nothing
```

ChrisRackauckas:
Yes, the LinearSolution does not define interface functions for this reason: it's just all internal to LinearSolve.jl. If we commit to making it be an actual documented interface, then it needs to become "the" actual documented interface, which yes is a big haul and shouldn't be in the same PR.

ValentinKaisermayer:
We can stick the problem into the cache and hence be able to have it in the solution, so it adheres to https://scimlbase.sciml.ai/dev/interfaces/Solutions/#Common-Field-Names

ValentinKaisermayer:
I will put the `prob` in the cache and forward access to it, so that it has the common fields. Then we can merge it as is? 

Sry if you meant this all along.

ChrisRackauckas (approved):

src/solutions/optimization_solutions.jl
```diff
@@ -17,28 +17,38 @@ Representation of the solution to an nonlinear optimization defined by an Optimi
   callback (`sol.retcode === :Terminated`), or whether it exited due to an error. For more
   details, see the return code section of the DifferentialEquations.jl documentation.
 """
-struct OptimizationSolution{T, N, uType, P, A, Tf, O} <: AbstractOptimizationSolution{T, N}
+struct OptimizationSolution{T, N, uType, C <: AbstractOptimizationCache, A, Tf, O} <:
```
ChrisRackauckas, line 20: this will likely cause a test failure?

ValentinKaisermayer: Yes it will.

ValentinKaisermayer: We could make a `build_solution` for `AbstracOptimizationProblem`, wrap the problem it in a `DefaultOptimizationCache` and warn the user that he should implement a cache interface.

ValentinKaisermayer: Also the `AbstractOptimizationCache` needs a big docstring, explaining what we expect from it.

ChrisRackauckas: Yes, I think we need to be much more careful here. I didn't catch this before, but we should make sure that this has as nice of an upgrade path as possible, because otherwise there will be a version mismatch issue

src/solutions/optimization_solutions.jl
```diff
@@ -34,11 +59,26 @@ function build_solution(prob::AbstractOptimizationProblem,
     T = eltype(eltype(u))
     N = ndims(u)
 
-    OptimizationSolution{T, N, typeof(u), typeof(prob), typeof(alg),
-                         typeof(minimum), typeof(original)}(u, prob, alg, minimum, retcode,
+    Base.depwarn("`build_solution(prob::AbstractOptimizationProblem, args...; kwargs...)` is deprecated." *
+                 " Consider implementing an `AbstractOptimizationCache` instead.",
+                 "build_solution(prob::AbstractOptimizationProblem, args...; kwargs...)")
```
ChrisRackauckas, line 64: Because this is a warning, please make sure the upgrade path is ready to be merged by EOD and that we get the versions out. We shouldn't throw warnings from our own packages for more then 24 hours.

PR #287, opened 2022-10-21
Add ReturnCodes for NonlinearSolve.jl

utkarsh530:
Meant for https://github.com/SciML/NonlinearSolve.jl/pull/92


utkarsh530:
I did not remove these conversion rules, because I am not sure if they will not break things (Although it might be fine since some of these are only used by `NonlinearSolve.jl`)  https://github.com/SciML/SciMLBase.jl/pull/287/files#diff-4c19a4b567174fa747f0e1554877954e98ae7b8799729961e524ce48fbd0b4c3L15

PR #294, opened 2022-10-28
Implements init dispatch

ValentinKaisermayer:
The current entrypoint is
```julia
solve(prob::OptimizationProblem, alg, args...; kwargs...)
```
and this method needs to be defined:
```julia
__solve(prob::OptimizationProblem, alg::SomeAlg, args...; kwargs...)
```
which is still possible, if no cache interface is defined.
If a cache interface should be defined the following methods have to be defined:
```julia
supports_opt_cache_interface(alg::SomeAlg) = true
__solve(cache::SomeAlgsOptimizationCache)::AbstractOptimizationSolution
__init(prob::OptimizationProblem, alg::SomeAlg, args...; kwargs...)::SomeAlgsOptimizationCache
```
The entry point is as in `CommonSolve`:
```julia
init(prob::OptimizationProblem, alg, args...; kwargs...)
solve!(cache::AbstractOptimizationCache)
```

### DIFF
--- a/src/solutions/optimization_solutions.jl
+++ b/src/solutions/optimization_solutions.jl
@@ -45,7 +45,8 @@ $(TYPEDEF)
 
 Representation the default cache for an optimization problem defined by an `OptimizationProblem`.
 """
-mutable struct DefaultOptimizationCache{F <: OptimizationFunction, P} <: AbstractOptimizationCache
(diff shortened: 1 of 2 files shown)


ValentinKaisermayer:
Needed for https://github.com/SciML/Optimization.jl/pull/381

Issue #297, opened 2022-11-04
Symbolic remake does not work for nested systems


ValentinKaisermayer:
https://github.com/SciML/SciMLBase.jl/blob/d7e3d316a014967414f416efa35e4d1aca5458e5/src/remake.jl#L59

ValentinKaisermayer:
```julia
using ModelingToolkit
using SciMLBase

function makesys(name)
    @parameters t a = 1.0
    @variables x(t)=0.0
    D = Differential(t)
    ODESystem([D(x) ~ -a * x]; name)
end

function makecombinedsys()
    sys1 = makesys(:sys1)
    sys2 = makesys(:sys2)
    @parameters t b = 1.0
    ODESystem(Equation[], t, [], [b]; systems = [sys1, sys2], name = :foo)
end

sys = makecombinedsys()
prob = ODEProblem(sys, Pair[])
prob_new = SciMLBase.remake(prob, p=Dict(sys1.a => 3.0, b => 4.0))
@show prob.p prob_new.p

prob.p = [1.0, 1.0, 1.0]
prob_new.p = [4.0]
```

You would have to query `ModelingToolkit.parameters(sys)`. Apperently a dimension check is missing aswell!

ValentinKaisermayer:
Also no system from MTK has a field `u0`:
https://github.com/SciML/SciMLBase.jl/blob/d7e3d316a014967414f416efa35e4d1aca5458e5/src/remake.jl#L62

ValentinKaisermayer:
I think also `handle_varmap` is wrong, since it just again uses the field via `getfield(sys, field)`. From MTK:
```julia
function SciMLBase.handle_varmap(varmap, sys::AbstractSystem; field = :states, kwargs...)
    out = varmap_to_vars(varmap, getfield(sys, field); kwargs...)
    return out
end
```
We could change the interface to 
```julia
function SciMLBase.handle_varmap(varmap, sys::AbstractSystem; type = :states, kwargs...)
    out = if type == :states
        varmap_to_vars(varmap, states(sys); kwargs...)
    elseif type == :parameters
        varmap_to_vars(varmap, parameters(sys); kwargs...)
    else
        error("unsupported")
    end
    return out
end
```

ValentinKaisermayer:
I think the best approach is to separate all of the symbolic varmap handling into MTK:

- #310 
- https://github.com/SciML/ModelingToolkit.jl/pull/1941
- #311

PR #299, opened 2022-11-04
Fix symbolic remake

ValentinKaisermayer:


ValentinKaisermayer:
closes #297

YingboMa:
format

ChrisRackauckas:
This needs tests

ValentinKaisermayer:
This is somewhat tricky since it is a circular dependency. This PR will fail until https://github.com/SciML/ModelingToolkit.jl/pull/1920 is merged. But https://github.com/SciML/ModelingToolkit.jl/pull/1920 will fail until this PR is merged.

ValentinKaisermayer:
I think this should not work:
https://github.com/SciML/DiffEqBase.jl/blob/7f565c60fffc18587e3b9c2a029866db679c2d5a/test/remake_tests.jl#L86-L94
Why should you be allowed to change from `NullParameters` to something else?

src/remake.jl
```diff
@@ -125,6 +127,15 @@ function remake(prob::ODEProblem; f = missing,
         _f = ODEFunction{isinplace(prob), specialization(prob.f)}(f)
     end
 
+    if prob.p isa NullParameters
+        p isa NullParameters || throw(ArgumentError("Can not change `NullParameters`."))
```
ChrisRackauckas, line 131:
```suggestion
        p isa NullParameters || throw(ArgumentError("Cannot change `SciMLBase.NullParameters`."))
```

test/runtests.jl
```diff
@@ -37,5 +37,7 @@ end
         @time @safetestset "Symbol and integer based indexing of integrators" begin include("downstream/integrator_indexing.jl") end
         @time @safetestset "Unitful interoplations" begin include("downstream/unitful_interpolations.jl") end
         @time @safetestset "Integer idxs" begin include("downstream/integer_idxs.jl") end
+        # @time @safetestset "Autodiff Remake" begin include("downstream/remake_autodiff.jl") end
```
ChrisRackauckas, line 40: what happened to this one?

ValentinKaisermayer: No idea, wasn't in the tests. I added it but commented out

ValentinKaisermayer:
Since you can change `f` it is possible that `p` and `u0` are not the same dimension or type as the old one.

test/runtests.jl
ChrisRackauckas: add it?

ChrisRackauckas:
This downstream test failure looks real: https://github.com/SciML/SciMLBase.jl/actions/runs/3422714840/jobs/5700446041

ValentinKaisermayer:
The problem is that this functionality need methods from MTK. But `SciMLBase` should not depend on it. So the `mergedefaults` and `handle_varmap` is introduced here and implemented in MTK. Which would need https://github.com/SciML/ModelingToolkit.jl/pull/1920. So we will get compat issues. 

A better way could be to implement a base variant of `remake` here and MTK adds a version which adds support for symbolics variable maps, i.e. `Vector{Pair}` 
or `Dict`. This would avoid this circular dependency. However, the fields to change are handled via keyword arguments, so no dispatch. We could add a `__remake(prob, f, p, u0)` to dispatch on.

ValentinKaisermayer:
Better still may be seperate methods for changing `f`, `p` and `u0` so that in MTK not everything has to be copy pasted for the symbolic remake of `p` and `u0`.

ChrisRackauckas:
What's the path to mergable here?

ValentinKaisermayer:
I couldn't figure out a type dispatch, so that MTK just implements two methods that allow a symbolic remake if it is loaded. So I split the current functionality and the one that considers nested systems. If this is merged, MTK needs to implement those methods in https://github.com/SciML/ModelingToolkit.jl/pull/1920 (needs a version constraint on SciMLBase)
```julia
function __varmap_to_vars end # a call to `varmap_to_vars` in MTK
function __parameters end # a call to `parameters(sys)`
function __states end # a call to `states(sys)`
```
After https://github.com/SciML/ModelingToolkit.jl/pull/1920 is merged, the old path can be removed in a follow up PR. 

I already state my view on the underlying issue. `SciMLBase` knows nothing about `AbstractSystems` nor anything about symbolic variables, nevertheless it is supposed to provide the functionality to remake with a symbolic map.

ValentinKaisermayer:
SciMLBase would just have the most basic variant in it:
```julia
if p === missing && u0 === missing
    p, u0 = prob.p, prob.u0
else # at least one of them has a value
    if p === missing
        p = prob.p
    end
    if u0 === missing
        u0 = prob.u0
    end
    if eltype(p) <: Pair || eltype(u0) <: Pair # one is a symbolic map
        p, u0 = process_p_u0_symbolic(prob, p, u0)
    end
end

# overloaded in MTK to intercept symbolic remake
function process_p_u0_symbolic end
```
@ChrisRackauckas What do you think about that approach?

ValentinKaisermayer:
I implemented three PRs that should separate out the symbolic handling to MTK.
1. #310 
2. https://github.com/SciML/ModelingToolkit.jl/pull/1941
3. #311

ValentinKaisermayer:
Closing this in favor of moving the symbolic handling to MTK

PR #300, opened 2022-11-04
add `_unwrap_val` to `SciMLBase`

oscardssmith:
This fixes a bug in `NonlinearSolve` and allows us to remove duplicated code from `OrdinaryDiffEq` and `LinearSolve` @YingboMa

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -479,3 +479,6 @@ function mergedefaults(defaults, varmap, vars)
         defaults
     end
 end
+
+_unwrap_val(::Val{B}) where {B} = B
+_unwrap_val(B) = B


oscardssmith:
@YingboMa I don't think that's needed. The OrdinaryDiffEq version is just `_unwrap_val(B) = B` which covers the bool and nothing case.

YingboMa:
Yeah, exactly, that's why I asked. Could you check what actually relies on `_unwrap_val(Nothing) = nothing`?

oscardssmith:
I really think the `LinearSolve` version is wrong. It turns `Val(nothing)` into `nothing` but `nothing` to `Nothing`. I think it's just a bug that doesn't really matter much.

oscardssmith:
it would require us to change `LinearSolve.OperatorAssumptions{Nothing}` to `LinearSolve.OperatorAssumptions{nothing}`

oscardssmith:
I think this is ready to merge. No matter what we do with LinearSolve this is good as is.

YingboMa (approved):

PR #284, opened 2022-10-18
Fix integrator parameter indexing, enable tests

AayushSabharwal:
Contains the fixes of #272 without the change that broke tests

PR #279, opened 2022-10-13
Revert "Fix integrator parameter indexing, enable tests"

YingboMa:
Reverts SciML/SciMLBase.jl#272 It's a breaking change.

PR #301, opened 2022-11-05
Revert "Revert "Fix integrator parameter indexing, enable tests""

ChrisRackauckas:
Reverts SciML/SciMLBase.jl#279

See https://github.com/SciML/SciMLBase.jl/pull/284 . This is actually the behavior that we wanted. We just need to test this better to make it more solidified.

### DIFF
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -58,6 +58,22 @@ step!(integrator, 100.0, true)
 @test integrator[γ] == 2.0
 @test integrator[(lorenz1.σ, lorenz1.ρ)] isa Tuple
 
+@test length(integrator[[lorenz1.x, lorenz2.x]]) == 2
+@test integrator[[γ, lorenz1.σ]] isa Vector{Float64}
+@test length(integrator[[γ, lorenz1.σ]]) == 2
+
+@variables q(t)[1:2] = [1.0, 2.0]
+eqs = [D(q[1]) ~ 2q[1]
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+integrator2 = init(prob2, Tsit5())
+
+@test integrator2[q] isa Vector{Float64}
+@test length(integrator2[q]) == length(q)
+@test integrator2[collect(q)] == integrator2[q]
+
 @testset "Symbolic set_u!" begin
     @variables u(t)
     eqs = [D(u) ~ u]
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -73,6 +73,7 @@ integrator2 = init(prob2, Tsit5())
 @test integrator2[q] isa Vector{Float64}
 @test length(integrator2[q]) == length(q)
 @test integrator2[collect(q)] == integrator2[q]
+@test integrator[(q...,)] isa NTuple{length(q), Float64}
 
 @testset "Symbolic set_u!" begin
     @variables u(t)
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -73,7 +73,7 @@ integrator2 = init(prob2, Tsit5())
 @test integrator2[q] isa Vector{Float64}
 @test length(integrator2[q]) == length(q)
 @test integrator2[collect(q)] == integrator2[q]
-@test integrator[(q...,)] isa NTuple{length(q), Float64}
+@test_broken integrator[(q...,)] isa NTuple{length(q), Float64}
 
 @testset "Symbolic set_u!" begin
     @variables u(t)
--- a/test/downstream/integrator_indexing.jl
+++ b/test/downstream/integrator_indexing.jl
@@ -54,8 +54,9 @@ step!(integrator, 100.0, true)
 @test integrator[lorenz1.x] isa Real
 @test integrator[t] isa Real
 @test integrator[α] isa Real
-# @test integrator[γ] isa Real
-# @test integrator[γ] == 2.0
+@test integrator[γ] isa Real
+@test integrator[γ] == 2.0
+@test integrator[(lorenz1.σ, lorenz1.ρ)] isa Tuple
 
 @testset "Symbolic set_u!" begin
     @variables u(t)
--- a/test/downstream/symbol_indexing.jl
+++ b/test/downstream/symbol_indexing.jl
@@ -66,8 +66,9 @@ sol = solve(prob, Rodas4())
 @test sol[α] isa Vector
 @test sol[α, 3] isa Float64
 @test length(sol[α, 5:10]) == 6
-# @test sol[γ] isa Real
-# @test sol[γ] == 2.0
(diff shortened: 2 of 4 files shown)


YingboMa (approved):

AayushSabharwal:
For two variables `x` and `y`, `sol[1:2, :]` returns a 2xN matrix on the latest SciMLBase, so `sol[[x,y]]` should too, right?

AayushSabharwal:
There will likely need to be some restructuring to support parameter indexing properly. What if `p` is a parameter, `x` is a variable, and the user tries to do `u[[x, p]]`? Currently, we get a `Vector{Any}`, with the first element being the timeseries of `x` and the second being the value of `p`.

The current "best" solution I can think of to fix the above comment, this one, and support type-stable indexing (`sol[(x, y)]` returns a `Vector{Tuple}`), is something like

```julia
Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, sym)
    if issymbollike(sym)
        if sym isa AbstractArray
            return A[collect(sym)]
        end
        i = sym_to_index(sym, A)
    elseif all(issymbollike, sym)
        if all_symbols_are_parameters(sym)
            return getindex.((A,), sym)
        else
            return [getindex.((A,), sym, i) for i in 1:length(A)]
        end
    else
        i = sym
    end
    # ...
end
```

AayushSabharwal:
I _think_ I have a solution that works exactly the same as indexing by numbers does, and does not require `all_symbols_are_parameters`. It disallows indexing parameters and variables at the same time.

AayushSabharwal:
Look like `test/downstream/symbol_indexing.jl:L93` needs to have `isa` instead of `is`

test/downstream/symbol_indexing.jl
```diff
@@ -66,8 +66,34 @@ sol = solve(prob, Rodas4())
...
+@variables q(t)[1:2] = [1.0, 2.0]
+eqs = [D(q[1]) ~ 2q[1]
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+sol2 = solve(prob2, Tsit5())
+
+@test sol2[q] isa Vector{Vector{Float64}}
+@test sol2[(q...,)] is Vector{NTuple{length(q), Float64}}
```
ChrisRackauckas, line 93:
```suggestion
@test sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
```

test/downstream/symbol_indexing.jl
```diff
@@ -66,8 +66,34 @@ sol = solve(prob, Rodas4())
...
+eqs = [D(q[1]) ~ 2q[1]
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+sol2 = solve(prob2, Tsit5())
+
+@test sol2[q] isa Vector{Vector{Float64}}
+@test sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
+@test length(sol2[q]) == length(sol)
```
ChrisRackauckas, line 94:
```suggestion
@test length(sol2[q]) == length(sol2)
```

test/downstream/symbol_indexing.jl
```diff
@@ -66,8 +66,34 @@ sol = solve(prob, Rodas4())
...
+@variables q(t)[1:2] = [1.0, 2.0]
+eqs = [D(q[1]) ~ 2q[1]
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+sol2 = solve(prob2, Tsit5())
+
+@test sol2[q] isa Vector{Vector{Float64}}
+@test sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
```
ChrisRackauckas, line 93:
```suggestion
@test_broken sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
```

AayushSabharwal: This isn't broken though

ChrisRackauckas: Oops? 😅  https://github.com/SciML/SciMLBase.jl/actions/runs/3553162451/jobs/5968519445#step:6:778 Wrong test lol

test/downstream/symbol_indexing.jl
```diff
@@ -66,8 +66,34 @@ sol = solve(prob, Rodas4())
...
+@variables q(t)[1:2] = [1.0, 2.0]
+eqs = [D(q[1]) ~ 2q[1]
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+sol2 = solve(prob2, Tsit5())
+
+@test sol2[q] isa Vector{Vector{Float64}}
+@test_broken sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
```
ChrisRackauckas, line 93:
```suggestion
@test sol2[(q...,)] isa Vector{NTuple{length(q), Float64}}
```

test/downstream/integrator_indexing.jl
```diff
@@ -54,8 +54,26 @@ step!(integrator, 100.0, true)
...
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+integrator2 = init(prob2, Tsit5())
+
+@test integrator2[q] isa Vector{Float64}
+@test length(integrator2[q]) == length(q)
+@test integrator2[collect(q)] == integrator2[q]
+@test integrator[(q...,)] isa NTuple{length(q), Float64}
```
ChrisRackauckas, line 76:
```suggestion
@test_broken integrator[(q...,)] isa NTuple{length(q), Float64}
```

test/downstream/integrator_indexing.jl
```diff
@@ -54,8 +54,26 @@ step!(integrator, 100.0, true)
...
+       D(q[2]) ~ 2.0]
+@named sys2 = ODESystem(eqs, t, [q...], [])
+sys2_simplified = structural_simplify(sys2)
+prob2 = ODEProblem(sys2, [], (0.0, 5.0))
+integrator2 = init(prob2, Tsit5())
+
+@test integrator2[q] isa Vector{Float64}
+@test length(integrator2[q]) == length(q)
+@test integrator2[collect(q)] == integrator2[q]
+@test_broken integrator[(q...,)] isa NTuple{length(q), Float64}
```
ChrisRackauckas, line 76: The integrator is missing something the solution interface has @AayushSabharwal

ChrisRackauckas: Continue on with this one, I'm sure it's just missing a piece, but since this is already a pretty major fix I'm merging and tagging.

PR #305, opened 2022-11-09
Throw warnings if symbol return codes are used

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02 @ValentinKaisermayer make note of this. Optimization.jl got a special overload in the constructor to not throw deprecation warnings because the solvers need to be updated, but this should get done ASAP. It's now the only package in SciML not natively using ReturnCode enum internally and the only one not throwing depwarns.

PR #306, opened 2022-11-10
ImplicitDiscreteProblem and ImplicitDiscreteFunction

YingboMa:


src/problems/implicit_discrete_problems.jl
```diff
@@ -0,0 +1,115 @@
...
+0 = f(u_{n+1}, u_{n}, p, t_{n+1}, integ)
+```
+
+`f` should be specified as `f(un,p,t)` (or in-place as `f(unp1,un,p,t)`), and
+`u_0` should be an AbstractArray (or number) whose geometry matches the desired
+geometry of `u`.  Note that we are not limited to numbers or vectors for `u₀`;
+one is allowed to provide `u₀` as arbitrary matrices / higher dimension tensors
+as well. ``u_{n+1}`` only depends on the previous iteration ``u_{n}`` and
+``t_{n+1}``. The default ``t_{n+1}`` is chosen adaptively, and when ``dt`` is
+specified, we have ``t_n = t_0 + n*dt``. `integ` constraints the field:
```
ChrisRackauckas, line 21:
```suggestion
specified, we have ``t_n = t_0 + n*dt``. `integ` contains the fields:
```

src/problems/implicit_discrete_problems.jl
```diff
@@ -0,0 +1,115 @@
...
+    ImplicitDiscreteProblem{isinplace}(f,u0,tspan,p=NullParameters(),callback=nothing)
+
+Defines a discrete problem with the specified functions.
+"""
+function ImplicitDiscreteProblem(f::ImplicitDiscreteFunction, u0, tspan::Tuple,
+                                 p = NullParameters(); kwargs...)
+    ImplicitDiscreteProblem{isinplace(f, 6)}(f, u0, tspan, p; kwargs...)
+end
+
+function ImplicitDiscreteProblem(f::Base.Callable, u0, tspan::Tuple, p = NullParameters();
```
ChrisRackauckas, line 111:
```suggestion
function ImplicitDiscreteProblem(f, u0, tspan::Tuple, p = NullParameters();
```

ChrisRackauckas (approved):

PR #310, opened 2022-11-12
adds process_p_u0_symbolic

ValentinKaisermayer:
In order to be able to move everything related to processing symbolic `p` and `u0` maps to MTK.
- Next step is to implement the interface in MTK:  https://github.com/SciML/ModelingToolkit.jl/pull/1941
- And then to use it instead: https://github.com/SciML/SciMLBase.jl/pull/311

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -133,6 +133,9 @@ function remake(prob::ODEProblem; f = missing,
     end
 end
 
+# overloaded in MTK to intercept symbolic remake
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
We should make this have a single catchall dispatch that throws an error saying ModelingToolkit is required for this functionality, but I don't want that to delay the process so we can do that with the next SciMLBase PR

PR #313, opened 2022-11-13
Use Symbol for OptimizationSolution retcode

### LINKED ISSUE #312: Handling return codes of optimization problem
Currently the return codes are `SciMLBase.ReturnCode`: 
https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/optimization_solutions.jl#L26

However, Optimization.jl yields a `Symbol`. For MOI solvers the MOI return code is hence always converted to `ReturnCode.Failure = 9`

ValentinKaisermayer:
closes #312


ChrisRackauckas:
No, this is exactly the kind of error-prone behavior we are removing.

PR #314, opened 2022-11-18
Adds retcodes for optimization problems

### LINKED ISSUE #312: Handling return codes of optimization problem

ValentinKaisermayer:
closes #312

### DIFF
--- a/src/retcodes.jl
+++ b/src/retcodes.jl
@@ -318,6 +318,66 @@ EnumX.@enumx ReturnCode begin
     * successful_retcode = true
     """
     FloatingPointLimit
+
+    """
+    `ReturnCode.Optimal`
+
+    The optimization problem was successfully solved to an optimal solution.
+
+    ## Properties
+
+    * successful_retcode = true
+    """
+    Optimal
+
+    """
+    `ReturnCode.LocallyOptimal`
+
+    The optimization problem was successfully solved to a local optimal solution. Typical for 
+    non-linear problems.
+
+    ## Properties
+
+    * successful_retcode = true
+    """
+    LocallyOptimal
+
+    """
+    `ReturnCode.Infeasible`
+
+    The optimization problem was proven to be infeasible by the solver.
+
+    ## Properties
+
+    * successful_retcode = false
+    """
+    Infeasible
+
+    """
+    `ReturnCode.LocallyInfeasible`
+
+    The optimization problem was proven to be locally infeasible by the solver. Typical for 
+    non-linear problems.
+
+    ## Properties
+
+    * successful_retcode = false
+    """
+    LocallyInfeasible
+
+    """
+    `ReturnCode.MaxTime`
+
+    A failure exit state of the solver. If this return code is given, then the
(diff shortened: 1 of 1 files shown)


src/retcodes.jl
```diff
@@ -318,6 +318,66 @@ EnumX.@enumx ReturnCode begin
...
+    """
+    `ReturnCode.Optimal`
+
+    The optimization problem was successfully solved to an optimal solution.
+
+    ## Properties
+
+    * successful_retcode = true
+    """
+    Optimal
```
ChrisRackauckas, line 331: This is just success

src/retcodes.jl
```diff
@@ -318,6 +318,66 @@ EnumX.@enumx ReturnCode begin
...
+    `ReturnCode.LocallyOptimal`
+
+    The optimization problem was successfully solved to a local optimal solution. Typical for 
+    non-linear problems.
+
+    ## Properties
+
+    * successful_retcode = true
+    """
+    LocallyOptimal
```
ChrisRackauckas, line 343: I think this is just implied by success for local optimization

src/retcodes.jl
```diff
@@ -348,6 +376,10 @@ function Base.convert(::Type{ReturnCode.T}, retcode::Symbol)
         ReturnCode.ConvergenceFailure
     elseif retcode == :Failure || retcode == symfalse
         ReturnCode.Failure
+    elseif retcode == :Infeasible || retcode == :INFEASIBLE ||
+           retcode == :DUAL_INFEASIBLE || retcode == :LOCALLY_INFEASIBLE ||
+           retcode == :INFEASIBLE_OR_UNBOUNDED
```
ChrisRackauckas, line 381: note that this conversion will become no longer automatic rather soon. I think we can keep it around as `symbol_to_retcode(retcode)`, but it won't be `convert` soon (since that invalidates everything).

ValentinKaisermayer: OK, but that is just a change to the solution constructor later?

ChrisRackauckas: I guess we could put it in the solution constructor? I hadn't actually thought about that. For everything except optimization, the conversion from Symbols is being entirely eliminated. It would probably be best to do this conversion in the MOI wrapper script, but we can do a slow move to that so it's never breaking.

ChrisRackauckas (approved):

src/retcodes.jl
ValentinKaisermayer: So a `_status_to_sciml_retcode` method in `Optimization.jl` and the solution only accepts `SciMLBase.ReturnCode`?

ChrisRackauckas: that accepts the MOI enums and spits out SciMLBase.ReturnCodes

PR #338, opened 2022-12-09
adds solve_time to opt. solution

### LINKED ISSUE #337: OptimizationSolution should include the time it took

ValentinKaisermayer:
closes #337


ChrisRackauckas:
Looks like this ends up being downstream breaking? That's not good: the libraries should be robust to a change like this. Do you know why the failures are popping up from this?

ValentinKaisermayer:
Because I forgot the type in the constructor 😅

ChrisRackauckas:
ahh that would do it. Run the formatter too and let's merge and tag this.

PR #342, opened 2022-12-12
Improved Symbolic Indexing

AayushSabharwal:
Uses SymbolicIndexingInterface.jl
Refer to https://github.com/SciML/ModelingToolkit.jl/pull/1988


AayushSabharwal:
I'm fairly certain the rest of the tests will only pass once MTK gets bumped and this can depend on the new version

AayushSabharwal:
If paramsyms/indepsym is passed to the SciMLFunction, then I guess there isn't a reason to not fallback to it

AayushSabharwal:
This needs RecursiveArrayTools for tests to pass as well

AayushSabharwal:
I didn't realise that some functions contain the sys but shouldn't use it for indexing, so it complains that they don't implement the interface. There's no way (as of now) to check if the interface is implemented, so it looks like defaulting `is_state_sym(::Any, _) = false` should do the trick

AayushSabharwal:
Now this should be compatible with previous MTK versions as well

AayushSabharwal:
Rerunning CI here _should_ make everything passed 🤞🏻

ChrisRackauckas:
Sounds suspicious.

AayushSabharwal:
Should I close this to not waste CI then reopen once its tagged?

ChrisRackauckas:
Told you, sus.

ChrisRackauckas:
Just gotta wait a bit haha https://github.com/JuliaRegistries/General/pull/74053

AayushSabharwal:
Dunno what is up with me today 😅 a bit all over the place

AayushSabharwal:
Finally merged

AayushSabharwal:
Apparently reopening doesn't rerun CI

ChrisRackauckas:
Still failing

AayushSabharwal:
That failure confuses me. In the system at `MTK/test/components.jl:99`, why does `states(sys)` say that there are two states (`capacitor.v` and `capacitor.i`) but the solution is Nx1 (timeseries of one element)

ChrisRackauckas:
The other must be an observed variable. How are those treated?

AayushSabharwal:
If a symbol doesn't match a state variable, it falls back to using `observed` as it did before. However, the system says its a state variable, so it never falls back.

It seems `ODAEProblem` does some additional simplification, which seems backward to me. Why would a problem modify the system? In any case, it makes the information that the system has invalid, which causes problems

ChrisRackauckas:
Might need to wait until @YingboMa is back

xtalax:
How is this going - this being merged seems like a prerequisite for `save_idxs`  as `SymbolicIndexingInterface` seems to solve a lot of the same problems

oscardssmith:
it looks like this just needs a format

ChrisRackauckas:
Boom

AayushSabharwal:
Is the ODAEProblem stuff fixed?

YingboMa:
Yes

oscardssmith:
See https://github.com/SciML/SciMLBase.jl/pull/379 which makes this also work for `DAESolution`.

PR #343, opened 2022-12-14
Fix parameter access from solution

ValentinKaisermayer:


ChrisRackauckas:
Needs tests

ValentinKaisermayer:
I think all of the solution interfaces need a refactor there is so much code duplication, that it makes it really hard to maintain.

ChrisRackauckas:
It is undergoing a heavy refactor right now in the generalization via SymbolicIndexingInterface.jl (https://github.com/SciML/SciMLBase.jl/pull/342). We should finish the generalization and test it all downstream well before starting the refactor. One of the things that will need a heavy refactor is the plot recipes. But yes, now that things can hold and query systems, all of that should only use the SymbolicIndexingInterface.jl functions and the fallbacks should be deprecated.

test/downstream/Manifest.toml
```diff
@@ -0,0 +1,1434 @@
+# This file is machine-generated - editing it directly is not advised
```
ChrisRackauckas, line 1: no manifest

ValentinKaisermayer: ups, sry

PR #346, opened 2022-12-20
Correct batch signature for scalar  f in or out

ArnoStrouwen:


src/problems/basic_problems.jl
```diff
@@ -310,27 +310,35 @@ Integral problems are multi-dimensional integrals defined as:
...
 whose geometry matches the space being integrated.
+This space is bounded by the lowerbound `lb` and upperbound `ub`,
+which are `Number`s or `AbstractVector`s with the same geometry as `u`.
 
 ## Problem Type
 
 ### Constructors
 
-IntegralProblem{iip}(f,lb,ub,p=NullParameters();
+IntegralProblem(f,lb,ub,p=NullParameters();
```
ChrisRackauckas, line 322: why this change?

ChrisRackauckas, line 322:
```suggestion
IntegralProblem{iip}(f,lb,ub,p=NullParameters();
```

ArnoStrouwen: Well there was already a sentence:
> Additionally, we can supply iip like IntegralProblem{iip}(...) as true or false to declare at
> compile time whether the integrator function is in-place.

PR #361, opened 2023-01-09
Add iscomplex type param to PDESolutions

xtalax:


ChrisRackauckas:
What is this? And won't other constructors need to be updated for this?

xtalax:
Just something to dispatch on for when the solution is complex, though on second thought I could do this with the metadata

PR #370, opened 2023-01-19
Add BVPFunction

ErikQQY:
This PR adds ```BVPFunction```. I think it would be nice for BoundaryValueDiffEq.jl to have this feature, since we have all kinds of ```*Function``` for all kinds of problems.

It is very interesting that after I added the ```BVPFunction``` feature to SciMLBase.jl and used the simple pendulum example to test if the ```BVPFunction``` can be directly used to solve the problem, and tada!🎉

bvpfunction

simplependulum


ErikQQY:
Sure, I will follow the BVP overhaul and keep this PR as a record in case a BVPFunction is needed.

ErikQQY:
I added ```bcjac``` and ```bcjac_prototype```, but the test cases are not covering all of the new features. Need some review to continue on this.

src/problems/bvp_problems.jl
```diff
@@ -78,50 +78,52 @@ every solve call.
...
 TruncatedStacktraces.@truncate_stacktrace BVProblem 3 1 2
 
-function BVProblem(f::AbstractODEFunction, bc, u0, tspan, args...; kwargs...)
+function BVProblem(f::AbstractBVPFunction, bc, u0, tspan, args...; kwargs...)
     BVProblem{isinplace(f, 4)}(f, bc, u0, tspan, args...; kwargs...)
 end
 
 function BVProblem(f, bc, u0, tspan, p = NullParameters(); kwargs...)
-    BVProblem(ODEFunction(f), bc, u0, tspan, p; kwargs...)
+    BVProblem(BVPFunction(f, bc), bc, u0, tspan, p; kwargs...)
```
avik-pal, line 115: If `BVPFunction` stores `bc` should be duplicate the storage in `BVProblem` again?

ErikQQY: Yeah, indeed, the problem construction process in ```BVProblem``` is similar with ```SDEProblem```, they both have duplicate ```bc``` in ```BVProblem``` and ```g``` in ```SDEProblem```, but if we want to have a ```BVPFunction``` stores both ```f``` and ```bc```, duplicating ```bc``` could really making the problem constructor concise. We need to note that when we are unpacking ```BVProblem```, we actually get a ```BVPFunction``` but not ```f```.

https://github.com/SciML/SciMLBase.jl/blob/4d926fccf201c68f8194c66e5580cec104bd047d/src/problems/sde_problems.jl#L130-L132

avik-pal: In that case it's fair to stay consistent.

Maybe we can add an additional dispatch on BVPFunction where we automatically pull out the `bc` during problem construction that way user wont have to do `BVProblem(BVPFunction(f, bc), bc...)` and instead can specify `BVProblem(BVPFunction(f, bc)....)`

ErikQQY: I think you mean a dispatch on ```BVProblem``` right? Just updated and now we can directly specify ```BVProblem(BVPFunction(f, bc).....)``` to construct a ```BVProblem```.
Do we need to do the same for ```SDEProblem``` and ```SDEFunction```? The problem construct of ```SDEProblem``` also need users to do somthing like ```SDEProblem(SDEFunction(f, g), g)```, see here: https://docs.sciml.ai/DiffEqDocs/stable/tutorials/sde_example/#Using-Higher-Order-Methods

avik-pal: @ChrisRackauckas  do you think this is a valid API choice? Is there a particular reason other Problem Types don't have this?

ChrisRackauckas: That was done so it would be non-breaking. But I'd be happy to just change the SDEProblem one at this point since it has effectively deprecated the `prob.g` a long time back.

ErikQQY: Updated both ```BVProblem``` and ```SDEProblem```.

ErikQQY: It looks like the change of ```SDEProblem``` is a breaking one.

ErikQQY: Sure, I am working on fixing StochasticDiffEq.jl errors

ChrisRackauckas: yes

src/problems/sde_problems.jl
```diff
@@ -124,13 +124,17 @@ end
...
-    SDEProblem{isinplace(f)}(f, g, u0, tspan, p; kwargs...)
+    SDEProblem{isinplace(f, 4)}(f, g, u0, tspan, p; kwargs...)
 end
 
 function SDEProblem(f, g, u0, tspan, p = NullParameters(); kwargs...)
     SDEProblem(SDEFunction(f, g), g, u0, tspan, p; kwargs...)
 end
 
+function SDEProblem(f::AbstractSDEFunction, u0, tspan, p = NullParameters(); kwargs...)
+    SDEProblem(f, f.g, u0, tspan, p; kwargs...)
```
ChrisRackauckas, line 135: this fallback won't work since it will drop information like Jacobian types.

ErikQQY:
Now BVP and SDE problems can be constructed using either:

```julia
BVProblem(f, bc, u0, tspan)
```
```julia
BProblem(f, g, u0, tspan)
```

or

```julia
BVProblem(fun, u0, tspan)
```
```julia
SDEProblem(fun, u0, tspan)
```
new_bvp
new_sde

ErikQQY:
I am still a little concerned the breaking changes would affect a lot, as can be seen from the CI status, the API changes especially ```SDEFunction``` and ```SDEProblem``` would make a lot of packages fail, so I need a confirmation before moving forward(update the downstream usage of ```SDEFunction``` and ```SDEProblem``` ). @ChrisRackauckas @avik-pal

avik-pal:
If the SDE thing requires a lot of patching, can we decouple it from the BVPFunction work? The BVP function is non-breaking IIUC.

ChrisRackauckas (approved):

PR #375, opened 2023-01-27
Decrement version

chriselrod:
We don't have a 1.81.1 yet, so decrement version back to that and tag a release? https://github.com/JuliaRegistries/General/blob/master/S/SciMLBase/Versions.toml


chriselrod:
https://github.com/SciML/SciMLBase.jl/commit/c7ea381317329cf53acba5b9413903f3a022a78d

PR #386, opened 2023-02-06
Problem indexing

xtalax:


ChrisRackauckas:
you need to bring in the rest of the PR since master reverted.

PR #387, opened 2023-02-06
Problem indexing

xtalax:


xtalax:
@ChrisRackauckas this should be finished

src/problems/problem_interface.jl
```diff
@@ -0,0 +1,51 @@
...
+    else
+        error("Invalid indexing of problem: $sym is not a symbol")
+    end
+end
+
+function Base.setindex!(prob::AbstractSciMLProblem, val, sym)
+    if has_sys(prob.f)
+        if issymbollike(sym)
+            params = getparamsyms(prob)
+            states = Symbol.(setdiff(getsyms(prob), params))
```
ChrisRackauckas, line 30: why not just use states?

xtalax: symbolic indexing interface wasn't working, and states is an mtk function

ChrisRackauckas: It's not an MTK function https://github.com/SciML/SymbolicIndexingInterface.jl/blob/master/src/interface.jl#L27

xtalax: if I use states, it can't find the index for some reason

ChrisRackauckas (approved):

src/SciMLBase.jl
```diff
@@ -10,6 +10,7 @@ using DocStringExtensions
 using LinearAlgebra
 using Statistics
 using Distributed
+using Setfield
```
ChrisRackauckas, line 13:
```suggestion
```

PR #388, opened 2023-02-06
Enable `integ[sys.p]` and `sol[sys.p]`

TorkelE:
Currently, this does not work:
```julia
using DifferentialEquations, ModelingToolkit, Plots

@parameters t σ ρ β
@variables x(t) y(t) z(t)
D = Differential(t)

eqs = [D(x) ~ σ * (y - x),
    D(y) ~ x * (ρ - z) - y,
    D(z) ~ x * y - β * z]

@named sys = ODESystem(eqs)

u0 = [x => 1.0, y => 0.0, z => 0.0]
p = [σ => 28.0, ρ => 10.0, β => 8 / 3]
tspan = (0.0, 10.0)

oprob = ODEProblem(sys, u0, tspan, p, jac = true)
integ = init(oprob)
sol = solve(oprob)

integ[sys.σ] # Does not work.
sol[sys.σ] # Does not work.
```
(inputting only `σ`, however, does)

With this PR the 
```
Base.getindex(A::AbstractTimeseriesSolution, sym)
Base.getindex(A::DEIntegrator, sym)
```
functions check this type of possibility, and returns the proper output.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -445,6 +445,8 @@ Base.@propagate_inbounds function Base.getindex(A::DEIntegrator, sym)
                 return A.t
             elseif has_sys(A.f) && is_param_sym(A.f.sys, sym)
                 return A.p[param_sym_to_index(A.f.sys, sym)]
+            elseif has_sys(A.f) && Symbol(sym) in Set(Symbol(A.f.sys.name,:₊,x) for x in getparamsyms(A))
+                return A.p[findfirst(x -> isequal(Symbol(A.f.sys.name,:₊,x), Symbol(sym)), getparamsyms(A))]
             elseif has_paramsyms(A.f) && Symbol(sym) in getparamsyms(A)
                 return A.p[findfirst(x -> isequal(x, Symbol(sym)), getparamsyms(A))]
             else
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -93,6 +93,8 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, s
                 return A.t
             elseif has_sys(A.prob.f) && is_param_sym(A.prob.f.sys, sym)
                 return A.prob.p[param_sym_to_index(A.prob.f.sys, sym)]
+            elseif has_sys(A.prob.f) && Symbol(sym) in Set(Symbol(A.prob.f.sys.name,:₊,x) for x in getparamsyms(A))
+                return A.prob.p[findfirst(x -> isequal(Symbol(A.prob.f.sys.name,:₊,x), Symbol(sym)), getparamsyms(A))]
             elseif has_paramsyms(A.prob.f) && Symbol(sym) in getparamsyms(A)
                 return A.prob.p[findfirst(x -> isequal(x, Symbol(sym)), getparamsyms(A))]
             else


TorkelE:
I have not currently added any tests. The SciML tests seem rather high-level, ideally I would like to test them using MTK, but that is not used. Not sure how to actually test this without MTK though.

src/solutions/solution_interface.jl
```diff
@@ -93,6 +93,8 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, s
                 return A.t
             elseif has_sys(A.prob.f) && is_param_sym(A.prob.f.sys, sym)
                 return A.prob.p[param_sym_to_index(A.prob.f.sys, sym)]
+            elseif has_sys(A.prob.f) && Symbol(sym) in Set(Symbol(A.prob.f.sys.name,:₊,x) for x in getparamsyms(A))
```
ChrisRackauckas, line 96: @YingboMa is this fine? It should use a generator, but fine with supporting that way of de-namespacing?

TorkelE:
I can provide some additional detail of what is going on:

Currently, `Base.getindex(A::DEIntegrator, sym)` checks for a couple of cases (`sym` is a state of `A`, `sym` is a parameter of `A`, and so on). It some point it has a default behaviour calling `observed(A, sym)`. This one, however, only returns an observable or state, not a parameter (that is why `sys.u` works as integrator indexing and not `sys.p`). I checked at modifying `observed(A, sym)`, but that is a true rabbit hole.

Next, I wanted to use something like 
```julia
Set(parameters(A.prob.f.sys, x) => x for x in parameters(A.prob.f.sys))
```
instead of 
```julia
Set(Symbol(A.prob.f.sys.name,:₊,x) for x in getparamsyms(A))
```
however, it turned out `parameters(sys)` generates an empty array within SciML base (while it works as expected within ModelingToolki). I could have done
```julia
Set(ModelingToolkit.parameters(A.prob.f.sys, x) => x for x in ModelingToolkit.parameters(A.prob.f.sys))
```
But having SciMLBase depend on MTK seemed bad/

src/integrator_interface.jl
```diff
@@ -445,6 +445,10 @@ Base.@propagate_inbounds function Base.getindex(A::DEIntegrator, sym)
                 return A.t
             elseif has_sys(A.f) && is_param_sym(A.f.sys, sym)
                 return A.p[param_sym_to_index(A.f.sys, sym)]
+            elseif has_sys(A.f) &&
```
YingboMa, line 448: Should we just ask the user to write `sys = complete(sys)`?

TorkelE: Is this always the case? I am only familiar with Catalyst (and not that familiar with `complete`, but would one always want this to be the case?

YingboMa: If you have `sys = complete(sys)`, `sys.x` will no longer namespace.

ChrisRackauckas: I don't know, I think that if this is correct, then it's a nice thing to add. Even without requiring a complete system.

ChrisRackauckas:
I think the main thing is to just use the SymbolicIndexingInterface states and parameters functions to have MTK do this. It'll be less brittle than putting the + in here.

TorkelE:
Ok, I am not familiar enough with this to be certain how to do it "properly" without talking it through with someone first. As long as it gets done relatively soon and doesn't become an issue that just lingers on I am happy.

TorkelE:
Superseded by https://github.com/SciML/SciMLBase.jl/pull/388

PR #391, opened 2023-02-15
Add `stats` / `destats` field to everything

avik-pal:


ChrisRackauckas:
can we just do this to all now?

avik-pal:
> can we just do this to all now?

Done

avik-pal:
The DelayDiffEq failure seems unrelated

ChrisRackauckas:
@YingboMa should we just destats -> stats with a deprecation path in the getproperty?

avik-pal:
@YingboMa bump

YingboMa (commented):
Still need the destats -> stats depreciation

avik-pal:
> Still need the destats -> stats depreciation

I bumped it to confirm if we wanted the deprecation. Will do that.

avik-pal:
The `test_nowarn` & `test_warn` tests really dislike this change :cry:

ChrisRackauckas (approved):

ChrisRackauckas:
Let's make sure the ArrayInterface stuff is more done before breaking some tests again with this 😅 . Other than that, I think this is good to go.

ChrisRackauckas:
Is downstream all set?

avik-pal:
Let's not merge it yet. I will run some of the final tests today and
confirm.

src/deprecated.jl
```diff
@@ -13,3 +13,6 @@ const DEProblem = AbstractDEProblem
 const DEAlgorithm = AbstractDEAlgorithm
 const DESolution = AbstractSciMLSolution
 const SciMLSolution = AbstractSciMLSolution
+
+# Deprecated `destats`
+@deprecate has_destats(x) has_stats(x)
```
devmotion, line 18: This will export `has_destats` even though currently it is not exported. This can be avoided by
```suggestion
@deprecate has_destats(x) has_stats(x) false
```

avik-pal:
The interface tests for ODE.jl are working locally & (Stochastic)DelayDiffEq.jl is also working. We should be good to go.

ChrisRackauckas:
Everything should have a deprecation path? Let me know if there are still any missing, but I think I caught them all by now.

PR #392, opened 2023-02-15
symbolic save idxs, new observed

xtalax:
Merge order is:
1. https://github.com/SciML/SymbolicIndexingInterface.jl/pull/7
2. This
3. https://github.com/SciML/OrdinaryDiffEq.jl/pull/1867
4. https://github.com/SciML/ModelingToolkit.jl/pull/2071

Am trying to test each part seperately but a lot of this is very tightly coupled, I will test it in composed in the MTK PR

### DIFF
--- a/src/solutions/dae_solutions.jl
+++ b/src/solutions/dae_solutions.jl
@@ -27,7 +27,7 @@ https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/
   exited due to an error. For more details, see
   [the return code documentation](https://docs.sciml.ai/SciMLBase/stable/interfaces/Solutions/#retcodes).
 """
-struct DAESolution{T, N, uType, duType, uType2, DType, tType, P, A, ID, DE} <:
+struct DAESolution{T, N, uType, duType, uType2, DType, tType, P, A, ID, DE, MType} <:
        AbstractDAESolution{T, N, uType}
     u::uType
     du::duType
@@ -39,6 +39,7 @@ struct DAESolution{T, N, uType, duType, uType2, DType, tType, P, A, ID, DE} <:
     interp::ID
     dense::Bool
     tslocation::Int
+    sym_map::MType
     destats::DE
     retcode::ReturnCode.T
 end
@@ -53,6 +54,7 @@ function build_solution(prob::AbstractDAEProblem, alg, t, u, du = nothing;
                                  HermiteInterpolation(t, u, du),
                         retcode = ReturnCode.Default,
                         destats = nothing,
+                        sym_map = nothing,
                         kwargs...)
     T = eltype(eltype(u))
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Isn't this needed in every solution type for the indexing?

xtalax:
is `build_explicit_observed_function` used for every system type? If so then yes, but I think it would require updating every `__init__` to feed through the symbol map. I will have a think about this

YingboMa:
Are the test failures related?

src/solutions/solution_interface.jl
```diff
@@ -73,7 +73,11 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, s
...
             return A[collect(sym)]
         end
-        i = sym_to_index(sym, A)
+        if hasfield(typeof(A), :sym_map) && !isnothing(A.sym_map)
+            i = get(A.sym_map, sym, nothing)
+        else
+            i = sym_to_index(sym, A)
+        end
     elseif all(issymbollike, sym)
         if has_sys(A.prob.f) && all(Base.Fix1(is_param_sym, A.prob.f.sys), sym) ||
```
YingboMa, line 82: `i` isn't defined in this branch. Maybe this causes the test failure.

xtalax: that branch always returns

xtalax:
Remaining failures look unrelated

src/solutions/dae_solutions.jl
```diff
@@ -39,6 +39,7 @@ struct DAESolution{T, N, uType, duType, uType2, DType, tType, P, A, ID, DE} <:
     interp::ID
     dense::Bool
     tslocation::Int
+    sym_map::MType
```
oscardssmith, line 42: Can this be done using `prob.f.sys`? Requiring explicit maps isn't great.

xtalax: no, because save_idxs isn't known at sys construction stage, only at solve

oscardssmith: checking out the OrdinaryDiffEq part of this makes this all make a lot more sense to me.

src/problems/ode_problems.jl
```diff
@@ -108,15 +108,18 @@ struct ODEProblem{uType, tType, isinplace, P, F, K, PT} <:
...
                                          u0, tspan, p = NullParameters(),
                                          problem_type = StandardODEProblem();
+                                         dense_output = true,
                                          kwargs...) where {iip}
         _tspan = promote_tspan(tspan)
         new{typeof(u0), typeof(_tspan),
             isinplace(f), typeof(p), typeof(f),
             typeof(kwargs),
-            typeof(problem_type)}(f, u0, _tspan, p, kwargs, problem_type)
+            typeof(problem_type)}(f, u0, _tspan, p, kwargs, problem_type, dense_output)
```
YingboMa, line 122: Why do we need this?

xtalax: otherwise it breaks observed for odae and others

YingboMa: Why would it break that?

xtalax: It was too hard to catch all the cases and not break everything without doing this. I would need some kind of signal from problem construction time to post solve regrardless, to signal what to call the observed functions with

YingboMa: What signal do you need other than save_idxs?

xtalax: we don't have save_idxs as problem construction time

YingboMa: Yeah, but maybe we should.

xtalax: @ChrisRackauckas thoughts on the above?

PR #413, opened 2023-03-13
stability improvement Integrals

ArnoStrouwen:


ChrisRackauckas:
Will need to fix the downstream test.

https://github.com/SciML/SciMLBase.jl/actions/runs/4402440188/jobs/7709607641#step:6:830

PR #422, opened 2023-03-21
Enable symbol interfacing to solution

TorkelE:
Currently, this is possible for accessing values of a solution object:
```julia
using DifferentialEquations, ModelingToolkit, Plots

@parameters t σ ρ β
@variables x(t) y(t) z(t)
D = Differential(t)

eqs = [D(x) ~ σ * (y - x),
    D(y) ~ x * (ρ - z) - y,
    D(z) ~ x * y - β * z]

@named sys = ODESystem(eqs)

u0 = [x => 1.0, y => 0.0, z => 0.0]
p = [σ => 28.0, ρ => 10.0, β => 8 / 3]
tspan = (0.0, 10.0)

oprob = ODEProblem(sys, u0, tspan, p, jac = true)
integ = init(oprob)
sol = solve(oprob)

sol[x]
sol[sys.x]

plot(sol, idxs=x)
plot(sol, idxs=[x])
plot(sol, idxs=sys.x)
plot(sol, idxs=[sys.x])
```

This PR adds a single check, enabling `getindex` on solution objects using Symbols. This enables:
```julia
plot(sol, idxs=:x)
plot(sol, idxs=[:x])
sol[:x]
```

For good measure (I can remove this, or add more), I have also added an error check that gives an error if one try to inde with symbols that have no corresponding property. E.g.
```julia
plot(sol, idxs=:X)
plot(sol, idxs=[:X])
sol[:X]
```
now gives errors:
```
Tried to index solution with a Symbol that was not found in the system using `getproperty`.
```

### DIFF
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -96,7 +96,11 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, s
             elseif has_paramsyms(A.prob.f) && Symbol(sym) in getparamsyms(A)
                 return A.prob.p[findfirst(x -> isequal(x, Symbol(sym)), getparamsyms(A))]
             else
-                return observed(A, sym, :)
+                if (sym isa Symbol) && has_sys(A.prob.f) && hasproperty(A.prob.f.sys, sym)
+                    return observed(A, getproperty(A.prob.f.sys, sym), :)
+                else
+                    return observed(A, sym, :)
+                end
             end
         else
             observed(A, sym, :)
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -249,12 +249,16 @@ DEFAULT_PLOT_FUNC(x, y, z) = (x, y, z) # For v0.5.2 bug
         Base.depwarn("To maintain consistency with solution indexing, keyword argument vars will be removed in a future version. Please use keyword argument idxs instead.",
                      :f; force = true)
         (idxs !== nothing) &&
-            error("Simultaneously using keywords vars and idxs is not supported. Please only use idxs.")
+            error("Simultaneously using keywords vars and idxs is not supported. Please only use idxs.")            
         idxs = vars
     end
 
     syms = getsyms(sol)
-    int_vars = interpret_vars(idxs, sol, syms)
+    if idxs isa Symbol
+        int_vars = interpret_vars([idxs], sol, syms)
+    else
+        int_vars = interpret_vars(idxs, sol, syms)
+    end
     strs = cleansyms(syms)
 
     tscale = get(plotattributes, :xscale, :identity)
@@ -456,9 +460,18 @@ function interpret_vars(vars, sol, syms)
                     var_int = tmp
                 end
             elseif issymbollike(var)
-                found = sym_to_index(var, syms)
-                var_int = found == nothing && getindepsym_defaultt(sol) == var ? 0 :
-                          something(found, var)
+                found = sym_to_index(var, syms)                
+                if (var isa Symbol) && has_sys(sol.prob.f)
+                    if hasproperty(sol.prob.f.sys, var)
+                        var_int = found == nothing && getindepsym_defaultt(sol) == var ? 0 :
+                            something(found, getproperty(sol.prob.f.sys, var))
+                    else
+                        error("Tried to index solution with a Symbol that was not found in the system using `getproperty`.")
+                    end
+                else
+                    var_int = found == nothing && getindepsym_defaultt(sol) == var ? 0 :
+                            something(found, var)
+                end
             else
                 var_int = var
             end
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -96,8 +96,12 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractTimeseriesSolution, s
             elseif has_paramsyms(A.prob.f) && Symbol(sym) in getparamsyms(A)
                 return A.prob.p[findfirst(x -> isequal(x, Symbol(sym)), getparamsyms(A))]
             else
-                if (sym isa Symbol) && has_sys(A.prob.f) && hasproperty(A.prob.f.sys, sym)
-                    return observed(A, getproperty(A.prob.f.sys, sym), :)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This should be added to the integrator indexing as well. And the problems.

https://github.com/SciML/SciMLBase.jl/blob/master/src/problems/problem_interface.jl
https://github.com/SciML/SciMLBase.jl/blob/master/src/integrator_interface.jl

TorkelE:
Currently, guven this PR, tehse are the interfaces that works/don't work:
```julia
oprob[x] = 2.0      # Works.
oprob[sys.x] = 3.0  # Errors.
oprob[:x] = 2.0     # Errors.
oprob[x]            # Works.
oprob[sys.x]        # Errors.
oprob[:x]           # Errors.

oprob[σ] = 3.0      # Works.
oprob[sys.σ] = 2.0  # Errors.
oprob[:σ] = 1.0     # Works.
oprob[σ]            # Works.
oprob[sys.σ]        # Errors.
oprob[:σ]           # Works.

integ[x] = 1.0      # Errors.
integ[sys.x] = 1.0  # Errors.
integ[:x] = 1.0     # Errors.
integ[x]            # Works.
integ[sys.x]        # Works.
integ[:x]           # Errors.

integ[σ] = 2.0      # Errors.
integ[sys.σ] = 4.0  # Errors.
integ[:σ] = 2.0     # Errors.
integ[σ]            # Works.
integ[sys.σ]        # Errors.
integ[:σ]           # Works.

sol[x]              # Works.
sol[sys.x]          # Works.
sol[:x]             # Works.

plot(sol, idxs=x)           # Works.
plot(sol, idxs=[x])         # Works.
plot(sol, idxs=sys.x)       # Works.
plot(sol, idxs=[sys.x])     # Works.
plot(sol, idxs=:x)          # Works.
plot(sol, idxs=[:x])        # Works.
```
I will start adding those I can while keeping code sensible.

oscardssmith:
Note that while the `plots` interfaces "work" they work using a very different mechanism from other indexing which means that they don't work for people with non-MTK based `sys`.

TorkelE:
Yeah, I was hoping fixing indexing for `sol` would automatically do it for `plot(...; idxs=...)` as well, be required separate implementation.

TorkelE:
Although maybe you meant more generally, as if someone does 
```
using DifferentialEquations
f(u,p,t) = 1.01*u
u0 = 1/2
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan
sol = solve(prob)
```
they cannot then use `plot(sol;idxs=:x)` stuff? The same should be true for using `x` and `sys.x` stuff, but we could always add a check with a specific error message if people tries to do this kind of stuff.

TorkelE:
These should now be possible as well@
```julia
oprob[x]            # Worked before.
oprob[sys.x]        # Works now.
oprob[:x]           # Works now.

oprob[σ]            # Worked before.
oprob[sys.σ]        # Works now.
oprob[:σ]           # Worked before.
```

TorkelE:
Added these as well:
```
integ[x]            # Worked before.
integ[sys.x]        # Worked before.
integ[:x]           # Works now.

integ[σ]            # Worked before.
integ[sys.σ]        # Works now.
integ[:σ]           # Worked before.
```

isaacsas:
Previously (I thought) the interface for changing `u0` or `p` was via `remake` and creating a new problem. Presumably this new functionality is intended to allow changing them (i.e. their components) directly, hence establishing a new interface. However, this interface will potentially lead to incorrect results for at least `JumpProblem` (I don't know about all the other problem types and what assumptions they make about their internal `p` and `u0` fields). 

If `JumpProblem` is the only case that will have issues with this we can define dispatches, but I'm not sure more generally if other problem types might also have problem components (i.e. in some `prob.field`) that depend on `u0` and `p`, and hence would also need updating if either is directly altered.

isaacsas:
If the mutating problem `u0` and `p` functionality is kept, can you factor out
```julia
Base.getindex(prob::AbstractSciMLProblem, sym)
```
to an internal function
```julia
_getindex(prob::AbstractSciMLProblem, sym)
```
that is `Base.getindex` just calls? This would allow downstream packages to reuse this functionality when creating custom dispatches (as, for example, `JumpProblem` would need).

TorkelE:
Sorry, maybe some misunderstanding. 

Both `remake` (via creating a new problem) and `getindex(problem...)` (via changing the current one) are interfaces for changing `u0` and `p`. Both of these are currently supported this PR doesn't;t add anything new (it simply permits you to use `getindex` with some additional arguments). Whichever of these two is the "go to" interface that we use in doc I am agnostic about.

Exactly what is the problem this may cause that `remake` is currently taking into account? We could try rebuilding the aggregator here as well. However, again, if this is a bug we have, it is bug that already exists in MTK, and not something added by this PR.

TorkelE:
I tried to dig through. It seems JumpProcesses.jl have a special `remake` for `JumpProblem` that includes a call
`update_parameters!(thing.massaction_jump, dprob.p; kwargs...)`
would we need to add this to `setindex` in the case of a JumpProblem having it's `p` changed? Is there anything else that would be missing? If it is just that it would be a easy thing to fix that in this PR as well.

ChrisRackauckas:
> Previously (I thought) the interface for changing u0 or p was via remake and creating a new problem. Presumably this new functionality is intended to allow changing them (i.e. their components) directly, hence establishing a new interface. However, this interface will potentially lead to incorrect results for at least JumpProblem (I don't know about all the other problem types and what assumptions they make about their internal p and u0 fields).

`remake` doesn't mutate.

ChrisRackauckas:
> If things look OK I will add some tests as well.

Yes please this looks like a great set of tests

TorkelE:
> `remake` doesn't mutate.

Right, I meant that `remake` (which I generally think is the recommended approach) does not mutate, as opposed to this approach which does.

TorkelE:
(tests should come tomorrow morning)

TorkelE:
I have modified the `setindex!(problem)`, having it call an internal function. This enables us to make a separate change to JumpProcesses which calls the internal function, but also updates the parameters, when called on a JumpProblem. Unless anyone come with further feedback I don't have any further changes to make.

I've run the formatted and it is happy, but for some reason the "src/solutions/pde_solutions.jl" file fails the format check, I don't really know about this file, so unsure what is going on.

PR #424, opened 2023-03-22
Add `get_p` for NonlinearSolution

YingboMa:


src/solutions/nonlinear_solutions.jl
```diff
@@ -36,6 +36,8 @@ TruncatedStacktraces.@truncate_stacktrace NonlinearSolution 1 2
 
 const SteadyStateSolution = NonlinearSolution
 
+SciMLBase.get_p(p::AbstractNonlinearSolution) = p.prob.p
```
ChrisRackauckas, line 39: ? I don't get this. Aren't the `get` functions all in ModelingToolkit?

YingboMa: https://github.com/SciML/SciMLBase.jl/blob/b5f4bc4a3622c495601d3cee92c61e1a1ae2d3ad/src/solutions/optimization_solutions.jl#L88 it's an interface function used by SciMLBase.

YingboMa:
The formatter failure isn't related.

ChrisRackauckas (approved):

PR #428, opened 2023-03-29
add a reverse overload for EnsembleSolution

ChrisRackauckas:


ChrisRackauckas:
@TorkelE it looks like your new indexing stuff is causing failures on master? Let me know what should be done to fix these tests. 

https://github.com/SciML/SciMLBase.jl/actions/runs/4556105610/jobs/8036045290?pr=428#step:6:822

TorkelE:
That's unexpected? Will investigate

PR #430, opened 2023-04-04
Fix indexing problem

TorkelE:
Re: https://github.com/SciML/SciMLBase.jl/pull/428


TorkelE:
There's some error on 1.6 that I haven't been unable to reproduce yet. However, this fixes things on the Julai v1 version.

ChrisRackauckas:
I'd be fine with `if VERSION >= v"1.8"` on those tests.

PR #432, opened 2023-04-09
Add missing prob kwargs in Optimization calls

ChrisRackauckas:


ChrisRackauckas:
@YingboMa can you look at https://github.com/SciML/SciMLBase.jl/actions/runs/4650151632/jobs/8228964554?pr=432#step:6:1566 ?

PR #441, opened 2023-05-10
Hook up `NoInit` in scimlBase

oscardssmith:
When combined with https://github.com/SciML/Sundials.jl/pull/401, this allows IDA to be used with `NoInit`. The test for this will be added to the sundials PR once this is merged.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -284,6 +284,8 @@ algorithm to use.
 function initialize_dae!(integrator::DEIntegrator)
     error("initialize_dae!: method has not been implemented for the integrator")
 end
+function initialize_dae!(integrator::DEIntegrator, initializealg::NoInit)
+end
 
 """
     auto_dt_reset!(integrator::DEIntegrator)


ChrisRackauckas:
Add a test on this? Would be good just to have a simple one

oscardssmith:
I was somewhat unsure about how to test it since there aren't any integrators defined in SciMLBase. (and `DummyIntegrator` in the tests is way too simple to be able to run `init` on). How would you recommend testing this?

ChrisRackauckas:
I guess downstream testing in Sundials should be sufficient.

YingboMa:
This broke MTKStdlib https://github.com/SciML/ModelingToolkit.jl/actions/runs/5008816715/jobs/8977085888?pr=2161#step:6:875

oscardssmith:
ah, that's not good. Will fix. How would you feel about me moving `BrownFullBasic` and `Shampine` to SciMLBase` (defaulting to errors). I really don't like the current situation where the initialization strategies only exist in `OrdinaryDiffEq` since it makes it really hard to get consistent behavior across solvers.

PR #442, opened 2023-05-18
Don't over specialize dae init

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/actions/runs/5008816715/jobs/8977085888?pr=2161#step:6:875


oscardssmith:
IMO this is wrong. someone who passes in an init other than `NoInit` would be rightfully surprised if initialization did nothing.

ChrisRackauckas (approved):

PR #445, opened 2023-05-25
Add NLStats struct for nonlinear solver stat tracking

oscardssmith:
This will make it possible for nonlinear solvers to keep track of their statistics which (among other things) will make it so DAE solvers don't skip counting the cost of initialization.

### DIFF
--- a/src/solutions/nonlinear_solutions.jl
+++ b/src/solutions/nonlinear_solutions.jl
@@ -4,14 +4,16 @@ $(TYPEDEF)
 Statistics from the nonlinear equation solver about the solution process.
 ## Fields
 - nf: Number of function evaluations.
-- nsolve: The number of linear solves `W\b` required for the integration.
-- njacs: Number of Jacobians calculated during the integration.
-- nsteps: Total number of iterations for the nonlinear solvers.
+- njacs: Number of Jacobians created during the solve.
+- nfactors: Number of factorzations of the jacobian required for the solve.
+- nsolve: Number of linear solves `W\b` required for the solve.
+- nsteps: Total number of iterations for the nonlinear solver.
 """
 mutable struct NLStats
     nf::Int
-    nsolve::Int
     njacs::Int
+    nfactors::Int
+    nsolve::Int
     nsteps::Int
 end
 
--- a/src/solutions/nonlinear_solutions.jl
+++ b/src/solutions/nonlinear_solutions.jl
@@ -1,3 +1,21 @@
+
+"""
+$(TYPEDEF)
+Statistics from the nonlinear equation solver about the solution process.
+## Fields
+- nf: Number of function evaluations.
+- nsolve: The number of linear solves `W\b` required for the integration.
+- njacs: Number of Jacobians calculated during the integration.
+- nnonliniter: Total number of iterations for the nonlinear solvers.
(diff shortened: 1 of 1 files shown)


src/solutions/nonlinear_solutions.jl
```diff
@@ -1,3 +1,20 @@
+
+"""
+$(TYPEDEF)
+Statistics from the nonlinear equation solver about the solution process.
+## Fields
+- nf: Number of function evaluations.
+- nsolve: The number of linear solves `W\b` required for the integration.
```
ChrisRackauckas, line 7: split factorizations and backsolves?

oscardssmith: yeah. not sure if that makes sense, I was just modeling this on the diffeq one. I'm happy to combine these

ChrisRackauckas: why would that not make sense?

oscardssmith: I think it does.

oscardssmith: fixed?

src/solutions/nonlinear_solutions.jl
```diff
@@ -1,3 +1,22 @@
...
+- nf: Number of function evaluations.
+- njacs: Number of Jacobians created during the solve.
+- nfactors: Number of factorzations of the jacobian required for the solve.
+- nsolve: Number of linear solves `W\b` required for the solve.
+- nsteps: Total number of iterations for the nonlinear solver.
+"""
+mutable struct NLStats
+    nf::Int
+    njacs::Int
+    nfactors::Int
```
ChrisRackauckas, line 15: this isn't following the same form as destats. Can you please make it look more uniform? I think it would be nfacts?

oscardssmith: destats doesn't have anything for factorizations. It has
```
- nf: Number of function evaluations. If the differential equation is a split function,
  such as a `SplitFunction` for implicit-explicit (IMEX) integration, then `nf` is the
  number of function evaluations for the first function (the implicit function)
- nf2: If the differential equation is a split function, such as a `SplitFunction`
  for implicit-explicit (IMEX) integration, then `nf2` is the number of function
  evaluations for the second function, i.e. the function treated explicitly. Otherwise
  it is zero.
- nw: The number of W=I-gamma*J (or W=I/gamma-J) matrices constructed during the solving
  process.
- nsolve: The number of linear solves `W\b` required for the integration.
- njacs: Number of Jacobians calculated during the integration.
- nnonliniter: Total number of iterations for the nonlinear solvers.
- nnonlinconvfail: Number of nonlinear solver convergence failures.
- ncondition: Number of calls to the condition function for callbacks.
- naccept: Number of accepted steps.
- nreject: Number of rejected steps.
- maxeig: Maximum eigenvalue over the solution. This is only computed if the
  method is an auto-switching algorithm.
  ```
  It has `nsolve` and `njacs` but nothing for the number of factorizations (possibly because it will always be the same as `njacs`?)

oscardssmith: bump @ChrisRackauckas

PR #446, opened 2023-06-01
Argument order for `NoSpecialized` `DAEFunction` was wrong

oscardssmith:


src/scimlfunctions.jl
```diff
@@ -3194,7 +3194,28 @@ function DAEFunction{iip, specialize}(f;
                                  _colorvec, sys)
     end
 end
-
+struct DAEFunction{iip, specialize, F, Ta, Tt, TJ, JVP, VJP, JP, SP, TW, TWt, TPJ, S, S2,
```
ChrisRackauckas, line 3197: Duplicated https://github.com/SciML/SciMLBase.jl/blob/4ef1c09f0b7db4ade308d7e819415f2ae84e384a/src/scimlfunctions.jl#L1663 ?

oscardssmith: yeah. Typo (I copied it so I could see both on the same screen and forgot to delete). I also am working on getting a test for this.

test/function_building_error_messages.jl
```diff
@@ -309,6 +309,7 @@ DAEFunction(dfoop, vjp = dvjp)
 dvjp(res, du, u, v, p, gamma, t) = [1.0]
 DAEFunction(dfiip, vjp = dvjp)
 DAEFunction(dfoop, vjp = dvjp)
+DAEFunction{true, SciMLBase.NoSpecialize}(dfiip, observed = 1)
```
ChrisRackauckas, line 312: that's not a valid observed function?

oscardssmith: all this is testing is that the observed gets put in the right place. Before this would error because it would try to put the the `observed` into the spot for `syms` and break.

ChrisRackauckas: This should check the function form and throw an error... in the future.

PR #447, opened 2023-06-01
Remove unnecessary branch in `getindex()`

staticfloat:
This branch does not seem to be necessary, and also encodes the requirement that `sys` has a `name` property, which may not always be the case.


ChrisRackauckas:
> and also encodes the requirement that sys has a name property, which may not always be the case.

What sys doesn't have a name?

ChrisRackauckas:
Which other branch handles the high level name space?

staticfloat:
Ah, I didn't understand the original case properly.  I didn't see any documentation that an `AbstractSystem` requires a `name` property, so I assumed it was optional.  Perhaps this should be changed to a `get_name()` call, like the other properties?

Issue #449, opened 2023-06-02
Warn users when using type unstable problem parameters


csimal:
Wrapping problem parameters in an array is an easy mistake to make that can hurt performance when parameters don't all have the same type.

It would be nice to display a message when the parameters of a problem are type unstable, and suggest to wrap parameters in something like a tuple instead, both for new users learning Julia/DiffEq and for old ones that still occasionally make this kind of mistake (me included). This kind of gotcha is easy to miss, so telling the user about it in context seems like an effective way to teach best practices.

![image](https://github.com/SciML/SciMLBase.jl/assets/45844176/e1249e09-8e9b-41c0-8ac0-afe9b4c1fc7f)

I have a working proof of concept [here](https://github.com/SciML/SciMLBase.jl/compare/master...csimal:SciMLBase.jl:master) and I'm ready to turn it into a PR, modulo possible adjustments to fit to the style guide.

ChrisRackauckas:
RFC: @YingboMa @chriselrod @Vaibhavdixit02  . I see the good and bad of this. It would be good to make sure there is a way to turn this off. Maybe `performance_warn = false`, but it's true by default, and then it's a general thing that we should warn about potential performance issues. I think it might be a nice thing to be louder about but I'd like to hear the comments from others. I think if we do this for differential equations then we should do it for all problem types.

csimal:
> I think if we do this for differential equations then we should do it for all problem types.
Of course! I just implemented a minimal example, but it's just one line to add to problem constructors.

chriselrod:
If it were up to me, type instabilities wouldn't even compile. ;)

ChrisRackauckas:
Thinking about this a bunch, I really like this. SciML is very focused on high performance. People are mad when they don't get that performance, and most of the time it's their fault. For most things people do in Julia, performance doesn't matter. You don't want Julia to be loud about every type instability in Plots.jl or HTTP.jl because you want to rely on that dynamism. But you do want to be loud about it in the simulators because those are made to be fast. A lot of the "bad benchmarks" come down to "well I didn't know any better". Well here you go, you were told by a warning to not do this if you want a good and fast simulation, and you ignored it. It's your fault. Assigning fault with software is something that I have learned over time is a very helpful meta idea. When users are told they did something wrong, they listen. When they aren't told, they think they used the software correctly. We would say obvious dynamic dispatches in a simulation is bad code smell, so we should tell people.

[...]

PR #454, opened 2023-06-08
Add performance warnings when creating problem instances

csimal:
Following the discussion in #449, this PR adds performance warnings displayed when the user creates a problem with type unstable parameters (typically an `Array{Any}`).

One thing still missing is adding keyword arguments to Problem constructors to disable the warning.


src/performance_warnings.jl
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+"""
+    warn_paramtype(p, warn_performance=PERFORMANCE_WARNINGS)
+
+Inspect the type of `p` and emit a warning if it could hurt
+performance when used to hold problem parameters.
+
+The warning can be turned off by setting `warn_performance` to `false`.
+To turn it off globally, call `set_performance_warnings!(false)`.
```
ChrisRackauckas, line 37: We should tell people to just add a preference to their startup, similar to how it's done in https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep/#Controlling-Function-Specialization-and-Precompilation

csimal: To clarify, you want me to change the docs to tell people what command to run, and ditch `set_performance_warnings!` ?

ChrisRackauckas: Yeah, and let them know it can be done in a startup.jl

ChrisRackauckas: For clarification, I am not sure `set_performance_warnings!` works as suggested, since you'd need to set the preference before the package is `using`ed, but it requires the package has been `using`ed in order to be called. That's why the Preferences system has a way of setting preferences via UUID.

src/performance_warnings.jl
```diff
@@ -0,0 +1,34 @@
...
+To turn it off globally within the active project you can execute the following code, or put it in your `startup.jl`.
+
+```julia
+using Preferences, UUIDs
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PerformanceWarnings" => false)
+```
+"""
+function warn_paramtype(p, warn_performance = PERFORMANCE_WARNINGS)
+    if warn_performance && should_warn_paramtype(p)
+        @info(WARN_PARAMTYPE_MESSAGE)
```
devmotion, line 32: Would it make sense to print a warning? From the name of the preference and function I'd assumed it would print one:
```suggestion
        @warn(WARN_PARAMTYPE_MESSAGE)
```

csimal: In the docs, it looks like `@warn` should be used when something went wrong. Here it's mostly a little heads up to the user, but otherwise doesn't affect the rest of the program. More like a warning about unused variables from the C compiler than a non-fatal error.
I don't feel strongly either way, so I will defer to you guys.

ChrisRackauckas: I'd go with warn, since we have other warnings that are mostly performance minded.

ChrisRackauckas:
Needs a rebase to the new format and I think this may be good to go.

csimal:
> Needs a rebase to the new format and I think this may be good to go.

:tada: but how do I do that exactly? (kinda new at open source contributions)

src/performance_warnings.jl
```diff
@@ -0,0 +1,34 @@
...
+
+should_warn_paramtype(p::AbstractArray) = !isconcretetype(eltype(p))
+should_warn_paramtype(::Dict{K, V}) where {K, V} = !isconcretetype(V)
+should_warn_paramtype(p::AbstractArray{<:AbstractArray}) = any(should_warn_paramtype, p)
+should_warn_paramtype(p::Tuple) = any(should_warn_paramtype, p)
+should_warn_paramtype(p::NamedTuple) = any(should_warn_paramtype, p)
+should_warn_paramtype(p) = false
+
+const WARN_PARAMTYPE_MESSAGE = """
+Using arrays or dicts to store parameters of different types can hurt performance. \
```
ChrisRackauckas, line 12:
```suggestion
Using arrays or dicts to store parameters of different types can hurt performance.
```

PR #457, opened 2023-06-16
improve `Vararg` checking

oscardssmith:
This code really shouldn't be messing around with function signatures, but if it is, it should do so better. This allows recognizing things like `g(x::Vararg{Vector{T}, N}) where {N, T}`.


ChrisRackauckas:
This needs a test.

ChrisRackauckas (approved):

ChrisRackauckas:
Test failure is real.

ChrisRackauckas:
bump

oscardssmith:
obsoleted by https://github.com/SciML/SciMLBase.jl/pull/1019

PR #467, opened 2023-07-13
add `WeightedEnsembleSolution`

oscardssmith:
This allows for easy indexing of an ensemble solution with importance weights. quantiles TODO.


chriselrod:
A `WeightedEnsembleProblem` would be similarly useful.

oscardssmith:
should be easy enough to add.

chriselrod:
> should be easy enough to add.

See https://github.com/SciML/SciMLBase.jl/pull/470

PR #469, opened 2023-07-14
`DataFrame` export

oscardssmith:
This allows getting a `DataFrame` from an ODE or Ensemble solution.


ChrisRackauckas:
ensemble indexing tests fail with this?

devmotion:
I always assumed the solutions support the Tables interface? Doesn't `DataFrame(x)` support anything that supports the Tables interface?

PR #470, opened 2023-07-14
Add  `WeightedEnsembleProblem`

chriselrod:


src/ensemble/ensemble_problems.jl
```diff
@@ -35,3 +35,22 @@ function EnsembleProblem(; prob,
     safetycopy = prob_func !== DEFAULT_PROB_FUNC)
     EnsembleProblem(prob, prob_func, output_func, reduction, u_init, safetycopy)
 end
+
+struct WeightedEnsembleProblem{T, T2, T3, T4, T5, T6} <: AbstractEnsembleProblem
+  ensembleprob::EnsembleProblem{T, T2, T3, T4, T5}
+  weights::T6
+end
```
oscardssmith, line 42: why this rather than just T1<: abstractensembleproblem?

chriselrod: How many kinds of `AbstractEnsembleProblem` are there?
I only knew of `EnsembleProblem` and now `WeightedEnsembleProblem`, and the latter doesn't make sense. 

But I agree, I'll change it to make it easier to add support for other problem types.

oscardssmith: I mostly just think it's a better idea since it gets rid of a lot of type parameters.

chriselrod: > I mostly just think it's a better idea since it gets rid of a lot of type parameters.

What is the value in that?
The way I had it before reduces the size of the printed struct.

ChrisRackauckas:
@vpuri3 is there a version we should just pull from SciMLOperators? We really need to get tests non red ASAP.

PR #473, opened 2023-07-22
Support W_prototype in ODEFunction

gaurav-arya:
Adds a barebones `W_prototype` to the ODEFunction.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -390,7 +390,7 @@ See the `modelingtoolkitize` function from
 automatically symbolically generating the Jacobian and more from the
 numerically-defined functions.
 """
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I think to handle other downstream tests, SDE, DDE, and SDDE would be required since they use the same nonlinear solver code paths.

gaurav-arya:
Not necessarily because of the check useed in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1996 here: https://github.com/SciML/OrdinaryDiffEq.jl/blob/e476b24726ffa639d61696178eeab0f5e8622171/src/derivative_utils.jl#L851

I don't think it would be too much work to add it and test if you'd rather do it all at once. But AFAIU it can also be done after

PR #479, opened 2023-08-11
Partial implementation of Zygote AD with symbolic indexing

contradict:
Symbolic indexing of state and observed variables. Time index must be specified for observed symbols.

This is a rebase and fixup of https://github.com/lamorton/scimlbase.jl/tree/zygote_rules


ext/ZygoteExt.jl
```diff
@@ -0,0 +1,97 @@
...
+            VA.k,
+            dprob,
+            VA.alg,
+            VA.interp,
+            VA.dense,
+            0,
+            VA.destats,
+            VA.alg_choice,
+            VA.retcode)
+        (Δ′, nothing, nothing)
```
ChrisRackauckas, line 42: This is the wrong formatting.

ChrisRackauckas: It won't revert this if you run the formatter, so put it all to one line and then reformat with the right format.

ext/ZygoteExt.jl
```diff
@@ -0,0 +1,97 @@
+module ZygoteExt
+
+using Zygote: pullback
+using ZygoteRules: @adjoint
+using SciMLBase: ODESolution, issymbollike, sym_to_index, remake, getobserved
+
+@adjoint function getindex(VA::ODESolution, i::Int, j::Int)
```
ChrisRackauckas, line 7: why does this need an overload?

contradict: Added a comment about this, without this there is a method ambiguity.

ext/ZygoteExt.jl
contradict: Changed, maybe fixed?

src/solutions/chainrules.jl
```diff
@@ -0,0 +1,72 @@
...
+        end
+    end
+    VA[sym, j], ODESolution_getindex_pullback
+end
+
+function ChainRulesCore.rrule(::typeof(getindex), VA::ODESolution, sym)
+    function ODESolution_getindex_pullback(Δ)
+        i = issymbollike(sym) ? sym_to_index(sym, VA) : sym
+        if i === nothing
+            throw("AD of purely-symbolic slicing for observed quantities is not yet supported. Work around this by using `A[sym,i]` to access each element sequentially in the function being differentiated.")
```
ChrisRackauckas, line 64:
```suggestion
            throw(error("AD of purely-symbolic slicing for observed quantities is not yet supported. Work around this by using `A[sym,i]` to access each element sequentially in the function being differentiated."))
```

src/solutions/zygote.jl
```diff
@@ -0,0 +1,22 @@
...
+        (Δ′, nothing)
+    end
+    VA[i], ODESolution_getindex_pullback
+end
+
+@adjoint function getindex(VA::ODESolution, sym)
+    function ODESolution_getindex_pullback(Δ)
+        i = issymbollike(sym) ? sym_to_index(sym, VA) : sym
+        if i === nothing
+            throw("Zygote AD of purely-symbolic slicing for observed quantities is not yet supported. Work around this by using `A[sym,i]` to access each element sequentially in the function being differentiated.")
```
ChrisRackauckas, line 14:
```suggestion
            throw(error("Zygote AD of purely-symbolic slicing for observed quantities is not yet supported. Work around this by using `A[sym,i]` to access each element sequentially in the function being differentiated."))
```

ChrisRackauckas:
Alright cool this looks good to go if tests pass. I hope we can get `sol[sym]` working ASAP afterwards since that's probably the most commonly used call, but at least there's a workaround for now.

contradict:
Oh, it fails on 1.6 since package extensions don't exist yet. Should I skip the tests on 1.6 or is there a workaround to get the extensions to load?

ChrisRackauckas:
Just skip v1.6. We're getting close to dropping it (please LTS come soon).

ChrisRackauckas (approved):

ChrisRackauckas:
https://github.com/SciML/SciMLBase.jl/actions/runs/5837232743/job/15843572165?pr=479#step:6:975

test/downstream/remake_autodiff.jl
```diff
@@ -1,27 +1,51 @@
...
         sensealg = BacksolveAdjoint(autojacvec = ZygoteVJP())))
 end
 
-u0 = [1.0 1.0]
-p = [1.5 1.0 1.0 1.0]
 du01, dp1 = Zygote.gradient(sum_of_solution, u0, p)
+
+# These tests depend on a ZygoteRule in a package extension
+# package exentsions do not exist before 1.9, so they cannot work.
+if __VERSION__ >= v"1.9"
```
ChrisRackauckas, line 33:
```suggestion
if VERSION >= v"1.9"
```

PR #483, opened 2023-08-22
Fix ODEFunction constructor

YingboMa:


ChrisRackauckas (approved):

ChrisRackauckas:
How did this come up? AutoSpecialize is almost always going to be preferred to NoSpecialize

YingboMa:
JuliaSimCompiler switches to `NoSpecialize` when the IR is sufficiently big.

YingboMa:
It helped compare to FullSpecialize, but I didn't compare it to `AutoSpecialize`.

PR #487, opened 2023-09-06
Fix BVPFunction errors

ErikQQY:
Try to fix ```BVPFunction``` errors

PR #489, opened 2023-09-07
Refactor SDEProblem constructor

ErikQQY:
Follow #487

Now we can define an ```SDEProblem``` by:

1. ```prob = SDEProblem(f, g, u0, tspan)```
2. ```prob = SDEProblem(SDEFunction(f, g), u0, tspan)``` (Before we construct ```SDEProblem``` using```SDEProblem(SDEFunction(f, g), g, u0, tspan)```)

**This PR would break a lot of downstream packages, especially the SDE stuff**


ChrisRackauckas:
I'm merging this to get it included in the v2 breaking changes. @ErikQQY can you prioritize fixing the downstream usage here a bit? I think we need to just get it done now so that way it's with the breaking change and thus everything updates appropriately.

ErikQQY:
Sure, I am working on the downstream errors.

PR #494, opened 2023-09-14
Fix adjoints for ODESolution indexing

frankschae:
Can/Should we add an upstream test to the Core1 group of SciMLSensitivity?


ChrisRackauckas:
Yes, please add upstream tests to scimlsensitivity. I would add all core tests.

PR #496, opened 2023-09-16
Make ODEProblem into a mutable struct

wsmoses:
Required for use in a reverse-mode EnzymeRule. Otherwise there is a mixed activity for having both floats and vector{floats}, not behind a pointer.


ChrisRackauckas:
I'm not sure we can take that since the semantics of a mutable struct vs struct will make it harder on parallelism. We should chat about this.

PR #497, opened 2023-09-17
add integrand interface

lxvm:
Hi,

As discussed in https://github.com/SciML/Integrals.jl/issues/169 and https://github.com/SciML/Integrals.jl/pull/175, Integrals.jl needs integrand wrapper types to support features such as inplace evaluation, batched evaluation, and inplace and batched evaluation at the same time. This pr adds 3 structs which support these 3 distinct interfaces and are intended as an eventual replacement for the `batch` and `nout` keywords to `IntegralProblem`. (I don't think that those need to be removed just yet, since we can dispatch on these wrapper types to opt into new behaviors.)
1. `InplaceIntegrand(f!, result::AbstractArray)`: the caller provides an inplace function `f!(y, x, p)` and an array `result` which will store the solution (i.e. it has the same shape as the integrand output array `y`, but could have a different type due to units of the limits).
2. `BatchIntegrand(f!, y::AbstractVector, x::AbstractVector; max_batch)`: the caller provides an inplace function `f!(y, x, p)` of the form `y .= f.(x, Ref(p))` and pre-allocated buffers that parallelize the calculation of the integrand over multiple quadrature nodes. For multi-dimensional integrands, the elements of `x` can be StaticArrays
3. `InplaceBatchIntegrand(f!, result::AbstractArray, y::AbstractArray, x::AbstractVector; max_batch)`: the caller provides an inplace function `f!(y, x, p)` that evaluates an array-valued inplace integrand at multiple points simultaneously. The solution is written to `result`. The buffer y is of size `(size(result)..., size(x)...)` and should be resize-able in its outer dimension (i.e. an `ElasticArray`)

I would appreciate any feedback on this pr. In particular, for multidimensional batched integrands it is nice to allow `x` to be a vector of `SVector`s instead of a `Matrix`, which is the current style. For the C libraries that only use `Matrix` we can use `reinterpret(reshape, ...)` to bring it into `SVector` form.


ChrisRackauckas:
There's no need for it to be a completely separate system from what already exists. Some uniformity would be good here. Make it `IntegralFunction` and `BatchIntegralFunction`, both as `AbstractSciMLFunction`s following a similar form to what is done in the others. In-place should just be the first argument and can be determined automatically using the machinery that's used in the other SciMLFunctions. That would make this feel much more in place (pun intended).

src/scimlfunctions.jl
```diff
@@ -2261,6 +2261,184 @@ end
...
+
+The remaining functions are optional and mainly used for accelerating the usage of `f`:
+- `jac`: unused
+- `paramjac`: unused
+- `analytic`: unused
+- `syms`: unused
+- `jac_prototype`: unused
+- `sparsity`: unused
+- `colorvec`: unused
+- `observed`: unused
```
ChrisRackauckas, line 2300: Those arguments can be removed.

src/scimlfunctions.jl
```diff
@@ -2261,6 +2261,184 @@ end
...
+
+Since most arguments are unused, the following constructor provides the essential behavior:
+
+```julia
+IntegralFunction(f, [I]; kws..)
+```
+
+If `I` is present, `f` is interpreted as in-place, and otherwise `f` is assumed to be
+out-of-place.
```
ChrisRackauckas, line 2309: Why is this needed? This should just be auto-determined.

ChrisRackauckas: Also, when passed it's `iip`

ChrisRackauckas:
I took a pass through it:

1. We will need some things like `observed` in the future, but for now leave it off. We generally add things as the interface is built out, and the MTK side of Integrals.jl is just missing for now so we can ignore it until it's there.
2. I'm a little weary that we cannot do proper dispatch checking on the batch version, but we at least we can test too many/too few arguments for appropriate errors.
3. The extra values were renamed `prototype` values to match other parts of the interfaces. What's expected then is that `init` will make a `copy` so that the solves are thread-safe. Something we need to keep in mind with the implementation (I don't think we've done much thread safety in the testing since there's also C programs involved).

Let me know if there's anything here you'd disagree with.

ChrisRackauckas:
> About thread safety, I don't think any integration library does multi-threaded integrand evaluation because of thread-safety issues, but with the BatchIntegralFunction the user does the parallelization themselves.

It's more if the user calls `solve` while multithreading. If this uses the same caches between the different solves (because of the same problem), then there will be clashing and incorrect results. 

> For the batch version, the only type check we can do is on the output_prototype. I think we should decide right away how to handle this. I don't know of any pure Julia libraries that offer inplace batched integrands, but the way the C libraries handle this is that the output is a vector when batching out-of-place and the output is a matrix when batching in-place. Tentatively, we could check output_prototype isa AbstractVector when batching out-of-place and output_prototype isa ElasticArray && integral_prototype isa AbstractArray && ndims(output_prototype) == ndims(integral_prototype) + 1. Packages like ArraysOfArrays.jl could wrap an ElasticArray into a vector of views, and it is an elegant solution we could handle as well, but I think we need to decide on the number of dimensions of output_prototype and I vote for the former, i.e. ndim > 1 (in line with C libraries), instead of the latter, i.e ndim == 1 (which is a possibility using Julia). I might be backwards-thinking on this

That might be overthinking it. I think we can just match the C libraries unless we gain some extra feature or performance.

test/function_building_error_messages.jl
```diff
@@ -601,3 +601,33 @@ BVPFunction(bfoop, bciip, vjp = bvjp)
...
+ioop(u, p) = p * u
+iiip(y, u, p) = y .= u * p
+i1(u) = u
+itoo(y, u, p, a) = y .= u * p
+
+IntegralFunction(ioop)
+IntegralFunction(iiip, Float64[])
+
+@test_throws SciMLBase.IntegrandMismatchFunctionError IntegralFunction(ioop, Float64[])
+@test_throws SciMLBase.IntegrandMismatchFunctionError IntegralFunction(i1)
```
ChrisRackauckas, line 616:
```suggestion
@test_throws SciMLBase.IntegrandMismatchFunctionError IntegralFunction(iiip)
```

ChrisRackauckas:
> Do most SciML packages wrap the user's input into a SciMLFunction, or expect the user to provide it themselves?

I missed that part. The problem type always has a default dispatch to wrap in the simplest function type. The IntegralProblem should default to wrapping the `f` in a `IntegralFunction`

ChrisRackauckas:
> I notice some function types have these backwards compatibility methods, so should I define those?

No need for those.

> but this would be breaking behavior unless we make IntegralFunction callable. 

Yes make it callable and just forward the args.

> remove the nout and batch keywords to IntegralProblem, since these can be obtained from output_prototypes and BatchIntegralFunctions

Yes let's go for it.

> Replace the limits [lb, ub] with a single domain argument, similar to how Symbolics.jl [handles integrals](https://github.com/JuliaSymbolics/Symbolics.jl/blob/master/src/integral.jl). Also add a method for converting [lb, ub] arguments into a domain

That sounds good.

src/problems/basic_problems.jl
```diff
@@ -364,30 +354,58 @@ compile time whether the integrator function is in-place.
...
+        if batch == 0
+            IntegralFunction(f)
+        else
+            BatchIntegralFunction(f, max_batch=batch)
+        end
+    end
+    IntegralProblem(g, args...; kwargs...)
+end
+function IntegralProblem(f, args...; kwargs...)
+    IntegralProblem{isinplace(f, 3)}(f, args...; kwargs...)
```
ChrisRackauckas, line 408:
```suggestion
function IntegralProblem(f, args...; nout = nothing, batch = nothing, kwargs...)
    if nout !== nothing || batch !== nothing
       @warn "`nout` and `batch` keywords are deprecated in favor of inplace `IntegralFunction`s or `BatchIntegralFunction`s. See the updated Integrals.jl documentation for details."
    end
    
    iip = isinplace(f, 3)
    g = if iip
        nout = nout === nothing ? 1 : nout
        output_prototype = Vector{Float64}(undef, nout)
        if batch == 0
            IntegralFunction(f, output_prototype)
        else
            BatchIntegralFunction(f, output_prototype, max_batch=batch)
        end
    else
        if batch == 0
            IntegralFunction(f)
        else
            BatchIntegralFunction(f, max_batch=batch)
        end
    end
    IntegralProblem{iip}(g, args...; kwargs...)
```

The previous approach would always throw the warning even when those keyword arguments weren't passed.

ChrisRackauckas:
The simple `IntegralProblem(f, (lb,ub), p)` dispatch for non-AbstractIntegralFunction is something that should still be considered standard and make work without error or warning. I think that's handled correctly, but it's a bit hard to see.

However, I think it's safe to merge this into the v2.0. The reason is because Integrals.jl is the only consumer (that I know of) of the SciMLBase.jl stuff, and so labeling it as a major means that we can do the breaking release and it won't actually be put into action until a PR bumps the lower bound on Integrals.jl. That will give us a little leeway for testing it out in implementation without being breaking to anyone's workflows until it's clear that it works.

Because of that, I'm going to merge. Let's do it, and we can fix what comes up during the bump of Integrals.jl

test/function_building_error_messages.jl
```diff
@@ -614,4 +614,40 @@ bvjp(u, v, p, t) = [1.0]
 @test_throws SciMLBase.NonconformingFunctionsError BVPFunction(bfoop, bciip, vjp = bvjp)
 bvjp(du, u, v, p, t) = [1.0]
 BVPFunction(bfiip, bciip, vjp = bvjp)
+BVPFunction(bfoop, bciip, vjp = bvjp)
```
ChrisRackauckas, line 617:
```suggestion
```

ChrisRackauckas (approved):

PR #501, opened 2023-09-22
Make ODEProblem into a mutable struct, v2

ChrisRackauckas:
Adds a big warning not to use it. This is only for EnzymeRules to work given current limitations!!!


Datseris:
I am hoping that this didn't have any impact to performance downstream when stepping ODE Integrators, right?

ChrisRackauckas:
It shouldn't. I didn't notice an immediate impact. Do mention if you see one.

PR #502, opened 2023-09-22
Add PyCall extension

LilithHafner:


LilithHafner:
I'm not sure how to make sense of the test failures. The [last merged PR](https://github.com/SciML/SciMLBase.jl/pull/503) had 55/72 successes, this has 54/71. Tests pass locally for me.

ChrisRackauckas:
It's not you it's me 😅 

We just did the v2 release yesterday, and this is package used by hundreds of other packages so the breaking update will take awhile to propagate through all of the downstream packages and pass.

LilithHafner:
What's the status here? Is there anything I should do to help?

ChrisRackauckas:
The tests are failing because it doesn't find pyjulia to be installed. Try moving the conda pip stuff to before the pycall is loaded? Somehow it needs to install that package on the Julia side to test it

LilithHafner:
Ah, lovely! And I can use [this CI failure](https://github.com/SciML/SciMLBase.jl/actions/runs/6371083896/job/17292733089?pr=502) as a guide, thanks!

LilithHafner:
Python tests pass! 🎉

ChrisRackauckas:
Awesome, thanks for pushing through it!

PR #507, opened 2023-09-26
Fix SDEProblem constructor issues

ErikQQY:
Tested locally, this patch may fix all the issues


ErikQQY:
Test failings are caused by version conflicts

PR #517, opened 2023-10-01
Fix tests failings for TwoPointBVPFunction

ErikQQY:
This should fix the tests failings in DiffEqBase.jl https://github.com/SciML/DiffEqBase.jl/pull/940

@ChrisRackauckas @avik-pal

### DIFF
--- a/src/problems/bvp_problems.jl
+++ b/src/problems/bvp_problems.jl
@@ -108,7 +108,7 @@ struct BVProblem{uType, tType, isinplace, P, F, BF, PT, K} <:
     problem_type::PT
     kwargs::K
 
-    @add_kwonly function BVProblem{iip}(f::AbstractBVPFunction{iip, TP}, bc, u0, tspan,
+    @add_kwonly function BVProblem{iip}(f::AbstractBVPFunction{iip, TP}, u0, tspan,
         p = NullParameters(); problem_type=nothing, kwargs...) where {iip, TP}
         _tspan = promote_tspan(tspan)
         warn_paramtype(p)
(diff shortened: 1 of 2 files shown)


src/problems/bvp_problems.jl
```diff
@@ -119,25 +119,25 @@ struct BVProblem{uType, tType, isinplace, P, F, BF, PT, K} <:
         else
             @assert prob_type === problem_type "This indicates incorrect problem type specification! Users should never pass in `problem_type` kwarg, this exists exclusively for internal use."
         end
-        return new{typeof(u0), typeof(_tspan), iip, typeof(p), typeof(f), typeof(bc),
-            typeof(problem_type), typeof(kwargs)}(f, bc, u0, _tspan, p, problem_type,
+        return new{typeof(u0), typeof(_tspan), iip, typeof(p), typeof(f), typeof(f.bc),
```
ChrisRackauckas, line 122: Why is the bc out here at all? Isn't it in the BVPFunction?

ChrisRackauckas: @avik-pal is it actually needed?

avik-pal: The `bc` was kept outside similar to the SDEProblem interface carrying around the `g` outside (I think?)

ChrisRackauckas: That was just fixed in the v2.0. I thought this was fixed in the v2.0 as well 😅. Fix and slip it in ASAP.

ErikQQY: The  modifications in ```BVProblem``` is similiar to ```SDEProblem```, just try to make them consistent

avik-pal: SDEProblem is still carrying around `g` https://github.com/SciML/SciMLBase.jl/blob/75b192581832ae89231df0d51a0fe9624e61470e/src/problems/sde_problems.jl#L88-L89

ErikQQY: Yeah, ```SDEProblem``` is still carrying ```g```, so in this way I think we can still use ```prob.g``` to directly access ```g```

ChrisRackauckas:
@avik-pal is this still needed after https://github.com/SciML/SciMLBase.jl/pull/518 ?

ChrisRackauckas:
This PR is now out of date because of the changes in https://github.com/SciML/SciMLBase.jl/pull/518, so closing, but a similar PR to the new function form is required.

PR #527, opened 2023-10-20
Promote initial state when converting from non-homogeneous python lists.

LilithHafner:
Fixes the problem reported in https://github.com/SciML/diffeqpy/issues/111#issuecomment-1772859284


LilithHafner:
Waiting for test to fail before pushing the fix

LilithHafner:
Python tests finally passed :)

PR #532, opened 2023-10-27
feat: indexing rework with new SymbolicIndexingInterface

AayushSabharwal:


src/ensemble/ensemble_solutions.jl
```diff
@@ -207,18 +207,8 @@ end
     end
 end
 
-Base.@propagate_inbounds function Base.getindex(x::AbstractEnsembleSolution, s, ::Colon)
-    return [xi[s] for xi in x]
-end
-
-Base.@propagate_inbounds function Base.getindex(x::AbstractEnsembleSolution,
```
oscardssmith, line 214: I am very happy you were able to get rid of this.

src/integrator_interface.jl
```diff
@@ -417,58 +417,87 @@ function getobserved(integrator::DEIntegrator)
...
+function SymbolicIndexingInterface.parameter_index(A::DEIntegrator, sym)
+    parameter_index(A.f, sym)
+end
+
+function SymbolicIndexingInterface.is_independent_variable(A::DEIntegrator, sym)
+    is_independent_variable(A.f, sym)
+end
+
+function SymbolicIndexingInterface.is_observed(A::DEIntegrator, sym)
+    return !is_variable(A, sym) && !is_parameter(A, sym) && !is_independent_variable(A, sym) && symbolic_type(sym) == ScalarSymbolic()
```
oscardssmith, line 453: This doesn't seem ideal. It seems like there should possibly be an easier way to define all of these.

src/scimlfunctions.jl
```diff
@@ -4308,3 +4308,82 @@ for S in [:ODEFunction
         end
     end
 end
+
+function SymbolicIndexingInterface.is_variable(fn::AbstractSciMLFunction, sym)
+  has_sys(fn) && return is_variable(fn.sys, sym)
+  has_syms(fn) && return Symbol(sym) in fn.syms
```
oscardssmith, line 4314: Any chance we can get rid of this part? i.e. in the `SciMLFunction` constructor we could turn `syms`/`paramsyms` into a `sys` with the appropriate behavior.

src/solutions/ode_solutions.jl
```diff
@@ -92,83 +92,44 @@ end
...
-            A.t,
-            sol.prob.f.sys,
-            observed,
-            p)
-    else
-        syms = hasproperty(sol.prob.f, :syms) && sol.prob.f.syms !== nothing ?
-               [sol.prob.f.syms[idxs]] : nothing
-        DiffEqArray(A.u, A.t, syms, getindepsym(sol), observed, p)
-    end
+    return DiffEqArray(A.u, A.t, p, sol)
```
oscardssmith, line 96: this is happy-making :)

oscardssmith:
Overall, this looks great! I think there might be a little more cleanup possible by getting rid of `syms`/`paramsyms` and always using `sys`, but this already is a substantial clean up.

src/solutions/ode_solutions.jl
AayushSabharwal: My favourite part of this PR was removing all the `getindex` methods for `AbstractTimeseriesSolution`.

src/integrator_interface.jl
oscardssmith: that sounds great!

ChrisRackauckas:
Needs a rebase?

AayushSabharwal:
Yep, almost done

AayushSabharwal:
Rebased

AayushSabharwal:
Should SciMLBase also export `getp` and `setp`?

hexaeder:
In my code I make heavy use of `odefunction.syms` to get the symbols for plotting purposes, find the right indices in the `u` vector, and so on.  Is there a recommended way to get the symbols now?
I found `odefunction.sys.variables` but this again uses some internals that might change in the future...

AayushSabharwal:
The recommended way is to import `SymbolicIndexingInterface` and use `variable_symbols(odefunction)`

ChrisRackauckas:
Indeed. The main purpose of this transition is to establish a well-defined interface that this is all based off of, rather than having to know internals voodoo for what's setup related to symbolic indexing. This is https://github.com/SciML/SymbolicIndexingInterface.jl, with our downstream packages implemented the documented interface https://docs.sciml.ai/SymbolicIndexingInterface/stable/api/ (and tutorials coming soon). If you stick to the documented interface and version control with respect to that, you will get guarantees of correctness in the setup which relying on internals could never give you. Hopefully this makes everything easier to maintain in the future, not just for us but for anyone like you who downstream is trying to use/extend the symbolic indexing in any way.

hexaeder:
Thanks a lot for the details, I'll look into that package! Pleas don't get me wrong, I did not mean to complain about the "breakage" at all -- that's what you get for using internals. I really like the transition to more and more tested and guaranteed interfaces. Luckily we have funding for a major rewrite next your so we can base our package on modern SciML standards removing many of those 3 year old hacks...

ChrisRackauckas:
Yes no worries. Interfaces are great, but you definitely have to rip off a few bandaids to put one in place where there wasn't one before. I mean, `.syms` served us well from 2016-2022, so it did its job 😅. I couldn't have predicted we needed an interface back when it was put in. But now we are in a different place and really formalizing what's going on is the only way for people to make use of all of its functionality correctly and in a way that allows maximum performance.

Datseris:
> Indeed. The main purpose of this transition is to establish a well-defined interface that this is all based off of, rather than having to know internals voodoo for what's setup related to symbolic indexing. This is https://github.com/SciML/SymbolicIndexingInterface.jl, with our downstream packages implemented the documented interface https://docs.sciml.ai/SymbolicIndexingInterface/stable/api/ (and tutorials coming soon). If you stick to the documented interface and version control with respect to that, you will get guarantees of correctness in the setup which relying on internals could never give you. Hopefully this makes everything easier to maintain in the future, not just for us but for anyone like you who downstream is trying to use/extend the symbolic indexing in any way.

Great, thanks for the work @AayushSabharwal and @ChrisRackauckas and everyone else working on this. I was just about to interface ODESystem in DynamicalSystems.jl and allow using symbols for e.g., `set_parameter` or `continuation`. Now I'll start learning and doing it the proper way!

PR #534, opened 2023-10-30
Relax type of alg in ensemble solve for optimization

Vaibhavdixit02:
This will allow using the existing code for Optimization.jl ensembles


src/ensemble/basic_ensemble_solve.jl
```diff
@@ -57,10 +57,10 @@ function __solve(prob::EnsembleProblem{<:AbstractVector{<:AbstractSciMLProblem}}
 end
 
 function __solve(prob::AbstractEnsembleProblem,
-    alg::Union{AbstractDEAlgorithm, Nothing},
+    alg::A,
     ensemblealg::BasicEnsembleAlgorithm;
     trajectories, batch_size = trajectories,
-    pmap_batch_size = batch_size ÷ 100 > 0 ? batch_size ÷ 100 : 1, kwargs...)
+    pmap_batch_size = batch_size ÷ 100 > 0 ? batch_size ÷ 100 : 1, kwargs...) where {A}
```
ChrisRackauckas, line 63:
```suggestion
    pmap_batch_size = batch_size ÷ 100 > 0 ? batch_size ÷ 100 : 1, kwargs...) where {A <: SciMLAlgorithm}
```

I think for this to work it needs a dispatch that's not `EnsembleAlgorithm`?

Vaibhavdixit02: Because of the automatic alg selection in DifferentialEquations.jl?

Vaibhavdixit02: I think I could work around this by making a stuct subtype of SciMLAlgorithm and wrap the optimizer's alg in that and handle it in `__solve` so we might not need to do this then

ChrisRackauckas: Can you double check some cases with this and defaults? Tests seem to pass

src/ensemble/ensemble_problems.jl
```diff
@@ -44,6 +44,26 @@ function EnsembleProblem(; prob,
...
+    output_func = DEFAULT_OUTPUT_FUNC,
+    reduction = DEFAULT_REDUCTION,
+    u_init = nothing, p = nothing,
+    safetycopy = prob_func !== DEFAULT_PROB_FUNC) where {uType}
+    EnsembleProblem(prob; prob_func, output_func, reduction, u_init, safetycopy)
+end
+
+function EnsembleProblem(; prob,
+    trajectories::Int,
+    prob_func,
```
ChrisRackauckas, line 59: this doesn't define a prob_func?

Vaibhavdixit02: Yes, because it'll need to access generated u0s, like https://github.com/SciML/Optimization.jl/pull/620/files#diff-6eea19b562ee1839438c6fa37d876324d501594af403caf5c836620ad7e3b793R6 or as a global variable so

test/runtests.jl
```diff
@@ -78,6 +78,12 @@ end
         @time @safetestset "Ensemble solution statistics" begin
             include("downstream/ensemble_stats.jl")
         end
+        # @time @safetestset "Ensemble Optimization and Nonlinear problems" begin
+        #     include("downstream/ensemble_nondes.jl")
+        # end
```
ChrisRackauckas, line 83: Is this intended?

Vaibhavdixit02: Yeah it was to show that other tests all pass, to get this one working it needs the dispatches we have been discussing

ChrisRackauckas: I'm lost. What's the version that works with all tests?

Vaibhavdixit02: The one without type assertion in the __solve method

ChrisRackauckas: Can you put it this in the form you believe works for everything and we can check the downstream tests and these tests?

PR #537, opened 2023-11-03
Move over the rest of pirating rules

ChrisRackauckas:
Will require a DiffEqBase update


ChrisRackauckas:
Needs to just merge because CI doesn't work well in cases of removing pirating. Downstream will update to require this.

PR #538, opened 2023-11-04
 Add constructor to convert NLLS to OptimizationProblem

Vaibhavdixit02:
Fixes https://github.com/SciML/Optimization.jl/issues/622

### DIFF
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -688,6 +688,14 @@ function OptimizationProblem(f, args...; kwargs...)
     OptimizationProblem{true}(OptimizationFunction{true}(f), args...; kwargs...)
 end
 
+function OptimizationProblem(prob::NonlinearLeastSquaresProblem; kwargs...)
+    if isinplace(prob)
+        throw(ArgumentError("Converting NonlinearLeastSquaresProblem to OptimizationProblem is not supported with in-place functions yet."))
+    end
+    optf = OptimizationFunction(sum ∘ prob.f, grad = (Jv, u, p) -> prob.f.jvp(Jv, prob.f(u, p), u, p), kwargs...)
+    return OptimizationProblem(optf, prob.u0, prob.p; prob.kwargs..., kwargs...)
+end
+
 isinplace(f::OptimizationFunction{iip}) where {iip} = iip
 isinplace(f::OptimizationProblem{iip}) where {iip} = iip
 
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -688,11 +688,18 @@ function OptimizationProblem(f, args...; kwargs...)
     OptimizationProblem{true}(OptimizationFunction{true}(f), args...; kwargs...)
 end
 
-function OptimizationProblem(prob::NonlinearLeastSquaresProblem; kwargs...)
+function OptimizationFunction(f::NonlinearFunction, adtype::AbstractADType = NoAD(); kwargs...)
+    if isinplace(f)
+        throw(ArgumentError("Converting NonlinearFunction to OptimizationFunction is not supported with in-place functions yet."))
+    end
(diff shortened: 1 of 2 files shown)


src/problems/basic_problems.jl
```diff
@@ -688,6 +688,14 @@ function OptimizationProblem(f, args...; kwargs...)
     OptimizationProblem{true}(OptimizationFunction{true}(f), args...; kwargs...)
 end
 
+function OptimizationProblem(prob::NonlinearLeastSquaresProblem; kwargs...)
+    if isinplace(prob)
+        throw(ArgumentError("Converting NonlinearLeastSquaresProblem to OptimizationProblem is not supported with in-place functions yet."))
+    end
+    optf = OptimizationFunction(sum ∘ prob.f, grad = (Jv, u, p) -> prob.f.jvp(Jv, prob.f(u, p), u, p), kwargs...)
```
ChrisRackauckas, line 695: this assumes jvp is defined? That's a pretty strong assumption?

ChrisRackauckas, line 695: `sum(abs2, prob.f(u,p))`, not just `sum`

Vaibhavdixit02: Yeah..I realise now that I was assuming this to come from being populated by an instatiate_function equivalent but it'll have to be provided by the user to be available here right?

Vaibhavdixit02: I think the loss for NLLS already has the square

ChrisRackauckas: @avik-pal wait, I thought the interface would be the same as nonlinearsolve, i.e. returning the vector of residuals. `f(du,u,p)` and then `sum(abs2,du)` is the loss. Now I'm looking at the tests and see the user is required to write it. Having that summation be implicit enforces it's actually a least squares problem.

avik-pal: Right it is same as NonlinearProblem, `sum ∘ prob.f` should be `sum(abs2,....)`

Vaibhavdixit02: IIUC what Chris meant was the other way around, the NLLS problem should do the squaring implicitly and not ask the user to do that (and then the loss here would just be the sum)

ChrisRackauckas: Okay I see what happened. https://github.com/SciML/NonlinearSolve.jl/commit/b8aca894e5168085f7f06ff656506c107f8788c6 the original test was just weird, but it was since fixed https://github.com/SciML/NonlinearSolve.jl/commit/883a3928cb4bf6893ea9a18e2c362b520891d5af. This means it's `sum(abs2,...)`.

Vaibhavdixit02: Ah yes, I had seen the previous version and added this

ChrisRackauckas:
Looks ready to merge? In-place is going to be much more common than out of place, so that would be good to support, but a proper error message is fine for now.

test/downstream/nllsopt.jl
```diff
@@ -0,0 +1,29 @@
+using NonlinearSolve, Optimization, OptimizationNLopt, ForwardDiff
+import FastLevenbergMarquardt, LeastSquaresOptim
```
ChrisRackauckas, line 2: These downstream deps would be required. But just remove `FastLevenbergMarquardt` and `LeastSquaresOptim`?

ChrisRackauckas, line 2:
```suggestion
using NonlinearSolve, Optimization, OptimizationNLopt, ForwardDiff
```

Vaibhavdixit02:
Yeah should be ready. But I don't completely like the interface, it needs the adtype as positional arg for the OptimizationProblem constructor for the "automatic" conversion but can't see any other way to do it for now.

ChrisRackauckas (approved):

PR #542, opened 2023-11-14
Add some additional methods to override the stored p and t

avik-pal:


avik-pal:
The warnings in https://github.com/JuliaDiff/SparseDiffTools.jl/pull/274 should go away with this.

PR #545, opened 2023-11-24
Print return code for NonTimeseries Solutions as well

avik-pal:


ChrisRackauckas:
Does every solution type have a retcode?

avik-pal:
Good point, I am conditionally printing it if available

PR #549, opened 2023-12-04
typos CI

ArnoStrouwen:


ChrisRackauckas:
nope, that seems to have slipped through 😅

PR #567, opened 2023-12-21
Add a nlls trait to BVProblem

avik-pal:
## Checklist

## Additional context

The main compilation cost for BVDiffEq is coming from solving NLLS problems. Having this allows us to have a type stable path for NLLS BVPs.

### DIFF
--- a/src/problems/bvp_problems.jl
+++ b/src/problems/bvp_problems.jl
@@ -138,22 +138,32 @@ struct BVProblem{uType, tType, isinplace, nlls, P, F, PT, K} <:
         end
 
         if nlls === nothing
+            if !hasmethod(length, Tuple{typeof(_u0)})
+                # If _u0 is a function for initial guess we won't be able to infer
+                __u0 = _u0 isa Function ?
+                       (hasmethod(_u0, Tuple{typeof(p), typeof(first(_tspan))}) ?
+                        _u0(p, first(_tspan)) : _u0(first(_tspan))) : nothing
+            else
+                __u0 = _u0
+            end
             # Try to infer it
-            if problem_type isa TwoPointBVProblem
+            if __u0 isa Nothing
+                _nlls = Nothing
+            elseif problem_type isa TwoPointBVProblem
                 if f.bcresid_prototype !== nothing
                     l1, l2 = length(f.bcresid_prototype[1]), length(f.bcresid_prototype[2])
-                    _nlls = l1 + l2 != length(_u0)
+                    _nlls = l1 + l2 != length(__u0)
                 else
                     # iip without bcresid_prototype is not possible
                     if !iip
                         l1 = length(f.bc[1](u0, p))
                         l2 = length(f.bc[2](u0, p))
-                        _nlls = l1 + l2 != length(_u0)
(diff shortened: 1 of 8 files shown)


src/problems/bvp_problems.jl
```diff
@@ -115,18 +124,50 @@ struct BVProblem{uType, tType, isinplace, P, F, PT, K} <:
...
+                    end
+                end
+            else
+                if f.bcresid_prototype !== nothing
+                    _nlls = length(f.bcresid_prototype) != length(_u0)
+                else
+                    if iip
+                        _nlls = false # Should we assume `true` instead?
+                    else
+                        _nlls = length(f.bc(FFakeSolutionObject(u0), p, tspan)) != length(_u0)
```
ErikQQY, line 162: Should be ```FakeSolutionObject```
```suggestion
                        _nlls = length(f.bc(FakeSolutionObject(u0), p, tspan)) != length(_u0)
```

src/solutions/ode_solutions.jl
```diff
@@ -115,6 +115,8 @@ struct ODESolution{T, N, uType, uType2, DType, tType, rateType, P, A, IType, S,
     stats::S
     alg_choice::AC
     retcode::ReturnCode.T
+    resid::R
+    original::O
```
ErikQQY, line 119: What about creating a ```BVPSolution``` to specifically handle these details only in the BVP context?

avik-pal: I am open to that. We can do that by subtyping AbstractODESolution.

src/solutions/ode_solutions.jl
```diff
@@ -117,6 +117,88 @@ struct ODESolution{T, N, uType, uType2, DType, tType, rateType, P, A, IType, S,
...
+    interp::IType
+    dense::Bool
+    tslocation::Int
+    stats::S
+    alg_choice::AC
+    retcode::ReturnCode.T
+    resid::R
+    original_ivp::OI
+    original_nlsolve::ON
+end
```
avik-pal, line 179: @ErikQQY thoughts on this?

ErikQQY: 👍I think with this we can specifically handle solutions in the BVP context

src/solutions/ode_solutions.jl
ChrisRackauckas: IMO it just seems a bit wasteful code wise. It's just two more fields, and `original` makes sense in the IVP case as well (we should use it with ODEInterface)

avik-pal:
I am happy with the current state of this. To summarize

1. Adds an `nlls` trait to BVProblem -- we try to infer it but err on the side of caution and set it to Nothing if we can reliably figure it out. It is the job of the upstream library to figure out how to deal with this case
2. Adds `original` to the ODESolution. I agree with @ChrisRackauckas that we probably don't want to unnecessarily copy the code into a BVPSolution since original might be useful for ODEs as well. `resid` not for IVPs but that is probably not a good enough reason to copy so much code.

ChrisRackauckas (approved):

PR #573, opened 2023-12-27
[WIP]Allow state and iterations as arguments in callback

Vaibhavdixit02:
This needs the solver wrappers other than OptimizationOptimisers to have it implemented

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -51,19 +51,19 @@ The callback function `callback` is a function which is called after every optim
 step. Its signature is:
 
 ```julia
-callback = (u, loss_val, other_args) -> false
+callback = (u, loss_val, other_args...) -> false
 ```
 
 where `u` and `loss_val` are the current optimization variables and loss/objective value
 in the optimization loop and `other_args` can be the extra things returned from
-the optimization `f`. This allows for saving values from the optimization and
-using them for plotting and display without recalculating. The callback should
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -51,19 +51,19 @@ The callback function `callback` is a function which is called after every optim
...
+callback = (u, loss_val, other_args...) -> false
 ```
 
 where `u` and `loss_val` are the current optimization variables and loss/objective value
 in the optimization loop and `other_args` can be the extra things returned from
-the optimization `f`. This allows for saving values from the optimization and
-using them for plotting and display without recalculating. The callback should
-return a Boolean value, and the default should be `false`, such that the optimization
-gets stopped if it returns `true`.
+the optimization `f` or the state of the optimizer and iteration count. This allows
```
ChrisRackauckas, line 59: " or the state of the optimizer and iteration count" what do you mean by that?

Vaibhavdixit02: I'll refine it. It's a placeholder to make sure I update it as soon as the update happens in the wrappers.

The point is that the callback will have a current state of the optimizer if it's available and iteration counts

ChrisRackauckas: I don't understand how that would work. Wouldn't that need to be an argument? If so, how is that argument uniformized? I don't see a struct definition or docs to go along with that.

PR #574, opened 2023-12-29
Aqua CI

ArnoStrouwen:
I'm not sure how DiffEqScalar works, since the reported method ambiguity seems like it should be regularly called.
For the type piracy, this can be solved by giving L a type.
If the type is `AbstractDiffEqOperator` then the method can stay here, but if it is `AbstractSciMLOperator` then the method should move.


test/runtests.jl
```diff
@@ -87,16 +83,14 @@ end
...
-            @time @safetestset "Symbol and integer based indexing of integrators" begin
-                include("downstream/integrator_indexing.jl")
-            end
-            @time @safetestset "Problem Indexing" begin
-                include("downstream/problem_interface.jl")
-            end
-            @time @safetestset "Solution Indexing" begin
-                include("downstream/solution_interface.jl")
-            end
+        @time @safetestset "Symbol and integer based indexing of integrators" begin
```
ChrisRackauckas, line 86: Someone needs to find the time to change all of these to 

```julia
@time @safetestset "Symbol and integer based indexing of integrators" include("downstream/integrator_indexing.jl")
```

This is what we agreed on as the right way, and the formatter is fine with it, but we just never got around to it. These kinds of clean up PRs would be a good time to just throw the regex on it.

ChrisRackauckas:
Those operators were deprecated and their documentation was deleted long ago. I think it's time to just remove them.

PR #576, opened 2023-12-30
Deprecate separate lb, ub arguments to IntegralProblem

lxvm:
Fixes https://github.com/SciML/Integrals.jl/issues/198
Fixes https://github.com/SciML/Integrals.jl/issues/208

Specifically, this pr deprecates the following method:
```julia
@deprecate IntegralProblem(f::AbstractIntegralFunction,
    lb::Union{Number,AbstractVector{<:Number}},
    ub::Union{Number,AbstractVector{<:Number}},
    p = NullParameters(); kwargs...) IntegralProblem(f, (lb, ub), p; kwargs...)
```

@ChrisRackauckas this pr could also be a good time to decide on the promotion behavior for the endpoints of the integration domain. I actually think promotion should be handled at solve time since it could interfere with AD, e.g. using ForwardDiff and differentiating just the upper limit of integration.

## Checklist

### DIFF
--- a/src/problems/basic_problems.jl
+++ b/src/problems/basic_problems.jl
@@ -426,8 +426,8 @@ IntegralProblem(f,lb,ub,p=NullParameters(); nout=nothing, batch=nothing, kwargs.
 - f: the integrand, callable function `y = f(u,p)` for out-of-place (default) or an
   `IntegralFunction` or `BatchIntegralFunction` for inplace and batching optimizations.
 - domain: an object representing an integration domain, i.e. the tuple `(lb, ub)`.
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
> this pr could also be a good time to decide on the promotion behavior for the endpoints of the integration domain. I actually think promotion should be handled at solve time since it could interfere with AD, e.g. using ForwardDiff and differentiating just the upper limit of integration.

Yes that's how we do it with the other problem types so it makes sense to do the same thing on IntegralProblem.

ChrisRackauckas:
I think this just needs to update the related tests and its good to go.

PR #583, opened 2024-01-02
fix: fix several bugs related to remake and symbolic indexing

AayushSabharwal:
## Checklist


test/remake.jl
```diff
@@ -0,0 +1,118 @@
+using ModelingToolkit, SymbolicIndexingInterface
```
isaacsas, line 1: Can you add JumpSystems to JumpProblems into this test to? (Which requires JumpProcesses?)

Note that JumpProblem is defined in JumpProcesses and has some remake related dispatches there that might need to be updated (basically they just forward all updates to the wrapped prob, i.e. `jump_problem.prob` which is usually a discrete or ODE problem).

AayushSabharwal: Resolved in https://github.com/SciML/SciMLBase.jl/pull/583/commits/d4d5466a90de0f586567e4d8aaa61c5518f9e4d1

ChrisRackauckas:
Add some inference tests as well.

ChrisRackauckas:
What's next here?

AayushSabharwal:
Nothing, it just needs [this](https://github.com/SciML/ModelingToolkit.jl/pull/2409) to be tagged first, which I assume will happen with MTKv9

ChrisRackauckas:
Looks like there's a circle where tests can't pass?

AayushSabharwal:
MTK tests should pass regardless of this PR

ChrisRackauckas:
What's the status of this one?

AayushSabharwal:
This PR needs https://github.com/SciML/ModelingToolkit.jl/pull/2409 to be tagged

AayushSabharwal:
TODO: Symbolic remake won't work with `SymbolCache` as the system as user-provided symbols. I think now that SII exists, `process_p_u0_symbolic` should be able to be implemented in SciMLBase and be independent of where the symbols come from?

AayushSabharwal:
CI won't pass here until https://github.com/SciML/ModelingToolkit.jl/pull/2409 is merged and tagged. I also have plans to remove `process_p_u0_symbolic` from MTK and just handle it in SciMLBase using SII. This would allow for symbolic initialization using DSLs (`SymbolCache`) as well.

isaacsas:
Is there any update on this or the PRs this requires before it can be merged?

PR #603, opened 2024-01-25
fix: remove implicit MTK dependency, other bug fixes, MTKv9 compatibility

AayushSabharwal:
Contrary to the name of the branch, this PR should be able to be merged without MTKv9

## Checklist

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -54,6 +54,7 @@ function remake(prob::ODEProblem; f = missing,
     tspan = missing,
     p = missing,
     kwargs = missing,
+    interpret_symbolicmap = true,
     _kwargs...)
     if tspan === missing
         tspan = prob.tspan
@@ -68,11 +69,26 @@ function remake(prob::ODEProblem; f = missing,
         if u0 === missing
             u0 = prob.u0
         end
-        if (eltype(p) <: Pair && !isempty(p)) || (eltype(u0) <: Pair && !isempty(u0)) # one is a non-empty symbolic map
-            has_sys(prob.f) ||
-                throw(ArgumentError("This problem does not support symbolic maps with `remake`, i.e. it does not have a symbolic origin." *
-                                    " Please use `remake` with the `u0`/`p` keyword arguments as vectors of values, paying attention to the order of initial values/parameters."))
+        isu0symbolic = eltype(u0) <: Pair && !isempty(u0)
+        ispsymbolic = eltype(p) <: Pair && !isempty(p) && interpret_symbolicmap
+        if isu0symbolic && !has_sys(prob.f)
+            throw(ArgumentError("This problem does not support symbolic maps with" *
+                " remake, i.e. it does not have a symbolic origin. Please use `remke`" *
+                "with the `u0` keyword argument as a vector of values, paying attention to" *
+                "parameter order."))
+        end
+        if ispsymbolic && !has_sys(prob.f)
+            throw(ArgumentError("This problem does not support symbolic maps with " *
+                "`remake`, i.e. it does not have a symbolic origin. Please use `remake`" *
+                "with the `p` keyword argument as a vector of values (paying attention to" *
+                "parameter order) or pass `interpret_symbolicmap = false` as a keyword argument"))
+        end
+        if isu0symbolic && ispsymbolic
             p, u0 = process_p_u0_symbolic(prob, p, u0)
+        elseif isu0symbolic
+            _, u0 = process_p_u0_symbolic(prob, prob.p, u0)
+        elseif ispsymbolic
+            p, _ = process_p_u0_symbolic(prob, p, prob.u0)
         end
     end
 
@@ -259,6 +275,7 @@ function remake(prob::OptimizationProblem;
     ucons = missing,
     sense = missing,
     kwargs = missing,
+    interpret_symbolicmap = true,
     _kwargs...)
     if p === missing && u0 === missing
         p, u0 = prob.p, prob.u0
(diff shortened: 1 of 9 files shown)


src/remake.jl
```diff
@@ -69,10 +69,10 @@ function remake(prob::ODEProblem; f = missing,
             u0 = prob.u0
         end
         if (eltype(p) <: Pair && !isempty(p)) || (eltype(u0) <: Pair && !isempty(u0)) # one is a non-empty symbolic map
-            hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :ps) ||
+            isempty(parameter_symbols(prob)) &&
```
oscardssmith, line 72: can we not use parameter_symbols here? it can be relatively expensive.

src/remake.jl
```diff
@@ -273,10 +273,10 @@ function remake(prob::OptimizationProblem;
...
                 throw(ArgumentError("This problem does not support symbolic maps with `remake`, i.e. it does not have a symbolic origin." *
                                     " Please use `remake` with the `p` keyword argument as a vector of values, paying attention to parameter order."))
-            hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :states) ||
+            isempty(variable_symbols(prob)) &&
```
oscardssmith, line 279: same here

src/remake.jl
AayushSabharwal: No, there almost always is a `SymbolCache` but it may or may not have symbols of a particular type

AayushSabharwal: Fixed

src/remake.jl
AayushSabharwal: Fixed

src/scimlfunctions.jl
```diff
@@ -2477,8 +2477,9 @@ function ODEFunction{iip, specialize}(f;
 
     _f = prepare_function(f)
 
-    sys = something(sys, SymbolCache(syms, paramsyms, indepsym))
-
+    if sys === nothing && (syms !== nothing  || paramsyms !== nothing || indepsym !== nothing)
```
oscardssmith, line 2480: can these 2 lines be factored out somehow? it feels kind of weird to need to write them so many times.

src/remake.jl
```diff
@@ -69,12 +69,9 @@ function remake(prob::ODEProblem; f = missing,
             u0 = prob.u0
         end
         if (eltype(p) <: Pair && !isempty(p)) || (eltype(u0) <: Pair && !isempty(u0)) # one is a non-empty symbolic map
-            hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :ps) ||
+            has_sys(prob.f) ||
```
ChrisRackauckas, line 72: Let's solve https://github.com/SciML/DifferentialEquations.jl/issues/922 at the same time here. If it doesn't have a system, then we shouldn't care what `p` is an just use it. `p` in a general diffeq is allowed to be a dictionary or Pair thing if people want, we shouldn't care.

But also, let's add a keyword argument `interpret_symbolicmap = true` and change the error message saying that this can be bypassed by setting that to false.

src/scimlfunctions.jl
```diff
@@ -4073,6 +4065,15 @@ function BatchIntegralFunction(f, integrand_prototype; kwargs...)
...
 end
 
+########## Utility functions
+
+function sys_or_symbolcache(sys, syms, paramsyms, indepsym = nothing)
+  if sys === nothing && (syms !== nothing || paramsyms !== nothing || indepsym !== nothing)
+    sys = SymbolCache(syms, paramsyms, indepsym)
+  end
+  return sys
+end
```
ChrisRackauckas, line 4075: The `syms` stuff is deprecated, I don't get why this is still staying around. If anything it should throw a big warning not to use it.

AayushSabharwal: Should I remove the keywords from the constructors or just `@warn`?

ChrisRackauckas: I guess `@warn` since it was accidentally documented, and remove it from the docstrings.

AayushSabharwal: Fixed in https://github.com/SciML/SciMLBase.jl/pull/603/commits/1f88d7fd6b88844d55cfdf91ff3971b31d9897ac

src/remake.jl
```diff
@@ -68,14 +69,26 @@ function remake(prob::ODEProblem; f = missing,
...
-                throw(ArgumentError("This problem does not support symbolic maps with `remake`, i.e. it does not have a symbolic origin." *
-                                    " Please use `remake` with the `p` keyword argument as a vector of values, paying attention to parameter order."))
-            hasproperty(prob.f, :sys) && hasfield(typeof(prob.f.sys), :states) ||
-                throw(ArgumentError("This problem does not support symbolic maps with `remake`, i.e. it does not have a symbolic origin." *
-                                    " Please use `remake` with the `u0` keyword argument as a vector of values, paying attention to state order."))
+        isu0symbolic = eltype(u0) <: Pair && !isempty(u0)
+        ispsymbolic = eltype(p) <: Pair && !isempty(p) && interpret_symbolicmap
+        if isu0symbolic && !has_sys(prob.f)
+            throw(ArgumentError("This problem does not support symbolic maps with" *
+                " remake, i.e. it does not have a symbolic origin. Please use `remke`" *
```
ChrisRackauckas, line 76:
```suggestion
                " remake, i.e. it does not have a symbolic origin. Please use `remake`" *
```

src/remake.jl
```diff
@@ -272,14 +286,26 @@ function remake(prob::OptimizationProblem;
...
```
ChrisRackauckas, line 293:
```suggestion
                " remake, i.e. it does not have a symbolic origin. Please use `remake`" *
```

ChrisRackauckas (approved):

PR #604, opened 2024-01-26
fix deprecated indexing in EnsembleSummary plot recipe

SebastianM-C:
I missed this ones.
@AayushSabharwal is this ok? It seems a bit odd to need `sim.u.u[1]` at first sight.

### DIFF
--- a/src/ensemble/ensemble_solutions.jl
+++ b/src/ensemble/ensemble_solutions.jl
@@ -156,17 +156,17 @@ end
 end
 
 @recipe function f(sim::EnsembleSummary;
-    trajectories = sim.u[1] isa AbstractArray ? eachindex(sim.u[1]) :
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
Yeah that is weird. `VectorOfArray`s shouldn't have a `VectorOfArray` as `u`. I'll investigate this further

SebastianM-C:
I'm not sure if it helps, but I was wondering if this is due to the fact that individual ODESolutions are no longer AbstractArrays and the recipes relied on this to distinguish between scalar and vector cases, as for ensemble you can have scalar differential equations or output functions that return scalars.

AayushSabharwal:
This does seem to be what is required here. I'm not fully on board with the `AbstractVectorOfArray.u` being another `AbstractVectorOfArray` thing that's going on, but if it works it works I guess.

SebastianM-C:
Should this be merged?

cc @ChrisRackauckas

PR #607, opened 2024-02-01
[WIP] Fix Ensemble indexing when `output_func` is used

SebastianM-C:
When `output_func` is used, the resulting type in `sim.u[i]` is no longer <: `AbstractVectorOfArray`, so we  need to avoid assuming that like in #599, #602 and #604

I was thinking of the following output functions as tests:

```julia
function output_func(sol, i)
    (x=sol[1, :], t=sol.t), false
end

function output_func(sol, i)
    last(sol), false
end

function output_func(sol, i)
    [first(sol), last(sol)], false
end
```

For the plot recipes I was thinking that we could test if no warnings are thrown by the `plot` calls. That should catch depwarns too and if there's any error in the recipes, the test would error.

This PR is WIP and not yet ready.

### DIFF
--- a/src/ensemble/ensemble_analysis.jl
+++ b/src/ensemble/ensemble_analysis.jl
@@ -3,7 +3,11 @@ module EnsembleAnalysis
 using SciMLBase, Statistics, RecursiveArrayTools, StaticArraysCore
 
 # Getters
-get_timestep(sim, i) = (sol.u[i] for sol in sim)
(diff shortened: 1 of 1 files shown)


AayushSabharwal (commented):
LGTM. Are these functions tested anywhere, in both cases (`.u isa AbstractVectorOfArray` and otherwise)?

PR #637, opened 2024-02-26
NonlinearProblem is not a DEProblem

avik-pal:
Fixes https://github.com/SciML/NonlinearSolve.jl/issues/154


ChrisRackauckas:
It's used for the DiffEqBase/solve.jl dispatches.

PR #643, opened 2024-03-05
Added trait for checking derivative requirement #711

ParasPuneetSingh:
## Checklist

## Additional context

Added trait for checking derivative requirement #711

### DIFF
--- a/src/alg_traits.jl
+++ b/src/alg_traits.jl
@@ -119,17 +119,6 @@ Defaults to false.
 requiresconstraints(opt) = false
 
 """
-requiresderivative(opt)
-
-Trait declaration for whether an optimizer
-requires derivatives specified in
-`cons` in `OptimizationProblem`.
-
-Defaults to false.
-"""
-requiresderivative(opt) = false
-
-"""
 requiresgradient(opt)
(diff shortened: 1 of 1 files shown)


src/alg_traits.jl
```diff
@@ -118,6 +118,17 @@ Defaults to false.
...
+"""
+requiresderivative(opt)
+
+Trait declaration for whether an optimizer
+requires derivatives specified in
+`cons` in `OptimizationProblem`.
+
+Defaults to false.
+"""
+requiresderivative(opt) = false
```
ChrisRackauckas, line 130: This is not a very explicit name. Did you mean requires gradient?

ChrisRackauckas: well we can't really change traits easily. They get set in stone once they are used downstream because it's highly breaking to change base namings. So if we're even considering changing it in the future, we should fix that now.

Vaibhavdixit02: @ParasPuneetSingh you can remove this now, we'll only need the ones below

Vaibhavdixit02 (approved):

PR #645, opened 2024-03-06
feat: add discrete saving feature to ODESolution

AayushSabharwal:
## Checklist


src/solutions/solution_interface.jl
```diff
@@ -54,10 +54,51 @@ SymbolicIndexingInterface.is_time_dependent(::AbstractTimeseriesSolution) = true
...
+    return SciMLStructures.replace(ps, SciMLStructures.Discrete(), discretes.u[t])
+end
+
+function SymbolicIndexingInterface.parameter_values_at_state_time(A::AbstractTimeseriesSolution, tidx)
+    ps = parameter_values(A)
+    if !isdefined(A, :discretes) || (discretes = A.discretes) === nothing
+        return ps
+    end
+    t = A.t[tidx]
+    idx = searchsortedfirst(discretes.t, t)
```
baggepinnen, line 78: `searchsortedfirst` returns the first index greater than or equivalent to `t`, that means that the case where there is an exact match `t` has to be handled separately, since then you should return `discretes.u[idx]` instead of `discretes.u[idx - 1]`?

src/solutions/solution_interface.jl
```diff
@@ -54,10 +54,51 @@ SymbolicIndexingInterface.is_time_dependent(::AbstractTimeseriesSolution) = true
...
+    if idx == firstindex(discretes.t)
+        error("This should never happen: there is no discrete parameter value before the current time")
```
baggepinnen, line 80: the same for this error, I should be able to get the value exactly at the first index.

src/solutions/solution_interface.jl
AayushSabharwal: Thanks for pointing that out! I'll fix it

AayushSabharwal: Fixed in https://github.com/SciML/SciMLBase.jl/pull/645/commits/1533e0f53d12043128b46954c68319828d709e01

src/solutions/solution_interface.jl
AayushSabharwal: Fixed in https://github.com/SciML/SciMLBase.jl/pull/645/commits/1533e0f53d12043128b46954c68319828d709e01

AayushSabharwal:
Requires https://github.com/SciML/RecursiveArrayTools.jl/pull/372 to merge and https://github.com/SciML/DiffEqBase.jl/pull/1019 to work

src/solutions/ode_solutions.jl
```diff
@@ -224,33 +255,87 @@ function (sol::AbstractODESolution)(t::Number, ::Type{deriv}, idxs::AbstractVect
         error("Incorrect specification of `idxs`")
     end
     error_if_observed_derivative(sol, idxs, deriv)
-    interp_sol = augment(sol.interp([t], nothing, deriv, sol.prob.p, continuity), sol)
-    first(interp_sol[idxs])
+    ps = parameter_values(sol)
+    # NOTE: This is basically SII.parameter_values_at_time but that isn't public API
+    # and once we move interpolation to SII, there's no reason for it to be
+    if is_parameter_timeseries(sol) == Timeseries()
```
baggepinnen, line 261: I had expected a function with the name `is_parameter_timeseries` to return a boolean, answering the question of the function name.

src/solutions/ode_solutions.jl
```diff
@@ -425,12 +517,33 @@ function solution_new_tslocation(sol::ODESolution{T, N}, tslocation) where {T, N
...
 function solution_slice(sol::ODESolution{T, N}, I) where {T, N}
+    new_t = sol.t[I]
+    if sol.discretes === nothing
+        discretes = nothing
+    else
+        mint, maxt = extrema(new_t)
+        masked_discs = map(sol.discretes) do disc
+            mini = searchsortedfirst(disc.t, mint)
+            maxi = searchsortedlast(disc.t, maxt)
+            disc[mini:maxi]
```
baggepinnen, line 537: this implementation is only correct if `I` is a unit range? Should the argument type of `I` be restricted so that this method cannot be called with, say, `I = [1, 2, 4]`?

src/solutions/ode_solutions.jl
AayushSabharwal: Yeah, it's actually a trait function necessary for dispatching which is why it returns a singleton object

baggepinnen: In that case, `parameter_timeseries_trait` would perhaps be better? 

A function starting with `is` but not returning a bool is very confusing

AayushSabharwal: Yeah that work work. I initially chose `is_parameter_timeseries` because we already have `is_timeseries` which (unfortunately) is also a trait function (for the `IsTimeseriesTrait`, which `is_parameter_timeseries` also leveranges)

src/solutions/ode_solutions.jl
AayushSabharwal: Fixed in https://github.com/SciML/SciMLBase.jl/pull/645/commits/bb65e2dd37fd793d86686dfd14c1452acff4aa5f

ChrisRackauckas (approved):

PR #666, opened 2024-04-14
Add SecondOrderBVProblem

ErikQQY:
Add a new problem type ```SecondOrderBVProblem```, for those BVPs have the form:

$$
y''(t)=f(t,y(t), y'(t))
$$

with bc:

$$
g(y(a), y(b), y'(a), y'(b))=0
$$

In this case we should define a ```SecondOrderBVProblem``` by:

```julia
function test(ddu, du, u, p, t)
    ϵ = 0.1
    ddu[1] = u[2]
    ddu[2] = (-u[1]*du[2] - u[3]*du[3])/ϵ
    ddu[3] = (du[1]*u[3] - u[1]*du[3])/ϵ
end

function bc(res, du, u, p, t)
    res[1] = u[1][1]
    res[2] = u[end][1]
    res[3] = u[1][3] + 1
    res[4] = u[end][3] - 1
    res[5] = du[1][1]
    res[6] = du[end][1]
end
u0 = [1.0, 1.0, 1.0]
tspan = (0.0, 1.0)
prob = SecondOrderBVProblem(test, bc, u0, tspan)
```

 also implemented the (MIRKN)Mono-Implicit Runge-Kutta Nystrom method for this kind of problem, will send a PR once this gets merged.


ChrisRackauckas (approved):

ErikQQY:
Seems there are still some errors need to be fixed

src/scimlfunctions.jl
```diff
@@ -3751,7 +3751,11 @@ function BVPFunction{iip, specialize, twopoint}(f, bc;
     end
 
     bciip = if !twopoint
-        isinplace(bc, 4, "bc", iip)
+        try
+            isinplace(bc, 4, "bc", iip)
+        catch e
+            isinplace(bc, 5, "bc", iip)
+        end
```
ErikQQY, line 3758: Don't know if it's the best way of fixing the issue(bcs are now in the form of ```bc(res, du, u, p, t)```)

ChrisRackauckas: no try catch, do this by having a separate function type. It should be the same as the dynamic ODE function

ErikQQY: Agree, this would be a better solution

ErikQQY: Done

PR #670, opened 2024-04-15
Remove check in `check_error!`

### LINKED ISSUE #669: `check_error!` returns `Success` but doesn't set it in `sol.retcode`
**Describe the bug 🐞**

If I run an integrator to the end with `step!` and call `check_error` I get `ReturnCode.Success` (`sol.retcode` is `ReturnCode.Default`). The docstring for `check_error!` states:

> Same as `check_error` but also set solution's return code

However it doesn't set the solution's return code, but leaves it on `ReturnCode.Default`. I see `solve!` [handles this case specifically](https://github.com/SciML/OrdinaryDiffEq.jl/blob/c04c54bac08405100c2433ac4c172916dc594e28/src/solve.jl#L561-L564) and therefore doesn't see this issue.

Not sure what the best fix is. The return code is not set due to this if condition being false:

https://github.com/SciML/SciMLBase.jl/blob/d947bed3aa31cdf54791930cd8272306d6bd0f20/src/integrator_interface.jl#L646

Should that condition perhaps be removed? Or are integration interface users expected to convert Default to Success themselves?

```julia

using OrdinaryDiffEq, SciMLBase

f(u, p, t) = 1.01 * u
u0 = 1 / 2
tspan = (0.0, 1.0)
prob = ODEProblem(f, u0, tspan)
integrator = init(prob, Tsit5(); reltol = 1e-8, abstol = 1e-8)
step!(integrator, tspan[end])

integrator.sol.retcode  # Default
check_error(integrator)  # Success
SciMLBase.check_error!(integrator) # Success
integrator.sol.retcode  # still Default
```

visr:
Fixes #669.

I'm not sure why this check was there. Was it just to save work? Why should the postamble not run if it is successful?

### DIFF
--- a/test/integrator_tests.jl
+++ b/test/integrator_tests.jl
@@ -1,6 +1,7 @@
 using SciMLBase
-mutable struct DummySolution
-    retcode::Any
+
+struct DummySolution
+    retcode::SciMLBase.ReturnCode.T
 end
 
 SciMLBase.solution_new_retcode(::DummySolution, code) = DummySolution(code)
@@ -73,3 +74,8 @@ end
 @test integrator.sol.retcode == ReturnCode.Default
 @test SciMLBase.check_error!(integrator) == ReturnCode.Success
 @test integrator.sol.retcode == ReturnCode.Success
+
+let
+    integrator = DummyIntegrator()
+    @test 0 == @allocated SciMLBase.check_error!(integrator)
+end
--- a/test/integrator_tests.jl
+++ b/test/integrator_tests.jl
@@ -3,6 +3,8 @@ mutable struct DummySolution
     retcode::Any
 end
 
+SciMLBase.solution_new_retcode(::DummySolution, code) = DummySolution(code)
+
 mutable struct DummyIntegrator{Alg, IIP, U, T} <: SciMLBase.DEIntegrator{Alg, IIP, U, T}
     uprev::U
     tprev::T
@@ -46,6 +48,9 @@ function SciMLBase.done(integrator::DummyIntegrator)
     integrator.t > 10
 end
 
+SciMLBase.check_error(::DummyIntegrator) = ReturnCode.Success
+SciMLBase.postamble!(::DummyIntegrator) = nothing
+
 integrator = DummyIntegrator()
 @test step_dt!(integrator, 1.5) == 2
 @test step_dt!(integrator, 1.5, true) == 1.5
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Does this allocation get elided?

visr:
Running this on the example from #669 I don't see any allocations, nor do I see it in a profile.

```
@btime SciMLBase.check_error!($integrator);
9.510 ns (0 allocations: 0 bytes)
```

I decided to give AllocCheck a try. It just triggered this which looks unrelated:

```
using AllocCheck
@check_allocs g(i, code) = SciMLBase.solution_new_retcode(i.sol, code)
g(integrator, ReturnCode.Default)

ERROR: @check_allocs function encountered 1 errors (1 allocations / 0 dynamic dispatches).

Allocating runtime call to "jl_get_pgcstack" in unknown location
```

Running AllocCheck on `check_error!` does give some allocations from `postamble!` calling `resize!(integrator.sol.t, integrator.saveiter)`.

ChrisRackauckas:
I mean in the context of a normal solve. Elision should happen in isolation but I want to make sure the compiler doesn't hit some limit say on Lorenz.

visr:
I saw allocations coming from `postamble!`. Since that isn't what I'm interested in, I put that back in the if block.

Now the number of allocations for Lorentz is all from setup, just like in this example:

https://docs.sciml.ai/DiffEqDocs/stable/tutorials/faster_ode_example/#Example-Accelerating-a-Non-Stiff-Equation:-The-Lorenz-Equation

test/integrator_tests.jl
```diff
@@ -62,3 +67,9 @@ for (uprev, tprev, u, t) in intervals(DummyIntegrator())
 end
 @test eltype(collect(intervals(DummyIntegrator()))) ==
       Tuple{Vector{Float64}, Float64, Vector{Float64}, Float64}
+
+@test integrator.sol.retcode == ReturnCode.Default
+@test check_error(integrator) == ReturnCode.Success
+@test integrator.sol.retcode == ReturnCode.Default
+@test SciMLBase.check_error!(integrator) == ReturnCode.Success
+@test integrator.sol.retcode == ReturnCode.Success
```
ChrisRackauckas, line 75: Add a test that nothing is allocated and this is good to go.

ChrisRackauckas:
Sorry for the delayed review, but add an allocs test and this is good.

test/integrator_tests.jl
visr: Is 14f6bced0e77b3075713d56cd73d4f56448c46b4 ok? Or do you also want to test this on Lorentz using OrdinaryDiffEq?

PR #674, opened 2024-04-18
Add ADTypes 1.0 compat

gdalle:
## Checklist

## Additional context

[ADTypes.jl](https://github.com/SciML/ADTypes.jl) is about to release a breaking v1.0, so the task of updating downstream repos begins.
This is a test PR to see if there is anything to change here. From what I can tell, the only use of ADTypes.jl is through `AbstractADType`, which is still available, so we should be good.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -50,7 +50,7 @@ SciMLBaseRCallExt = "RCall"
 SciMLBaseZygoteExt = "Zygote"
 
 [compat]
-ADTypes = "0.2.5"
+ADTypes = "0.2.5,1.0.0"
 ArrayInterface = "7.6"
 ChainRules = "1.58.0"
 ChainRulesCore = "1.18"


gdalle:
It seems making ADTypes.jl compatible only with Julia 1.10 and up might be hugely breaking for the SciML ecosystem at the moment?

gdalle:
@ChrisRackauckas thoughts on how to proceed here?

ChrisRackauckas:
I think it's fine to merge since this is just the interface, and thus it won't actually be used until the downstream users (currently Optimization.jl and NonlinearSolve.jl) allow it. This then isolates the problem to the downstream places where it's actually used, which should be fine.

gdalle:
I guess this deserves a minor version bump?

gdalle:
Hi! Now that https://github.com/SciML/ADTypes.jl/issues/46 is done, if ADTypes v1 doesn't break CI for SciMLBase, would it be possible to release v2.34?

PR #681, opened 2024-05-01
autospecialize rather than full specialize by default for the various `NonlinearProblems`

oscardssmith:
This doesn't change anything yet, but is a step in the direction of making NonlinearSolve not specialize on the user given function which will reduce compilation time on first nonlinear solve.


ChrisRackauckas:
might as well

oscardssmith:
are you sure the CI failures here weren't real? They looked plausibly related

avik-pal:
They do seem related

ChrisRackauckas:
Revert

oscardssmith:
https://github.com/SciML/SciMLBase.jl/pull/682

oscardssmith:
I'm really confused how this was causing problems since from looking at the code, AutoSpecialize should only change things for in place ODEFunctions.

PR #683, opened 2024-05-02
Chore: remove mentions of SDESolution

DhairyaLGandhi:
## Checklist

## Additional context

`SDESolution` doesn't seem to exist anymore. The more appropriate type seems to be `RODESolution`. This PR removes mentions of SDESolution and replaces them with `RODESolution`. Further, one of the adjoints was incorrect. It was definining an adjoint over SDEProblem, but instead tries returning SDESolution in the forwards pass. Not sure how this passed tests though


AayushSabharwal:
Could you add a test that differentiates through an `SDEProblem` solve? That way we can verify the new gradients work and ensure they continue to do so

PR #689, opened 2024-05-06
Feat: adjoints through observable functions

DhairyaLGandhi:
## Checklist

## Additional context

Currently, ADing through observables errors, however this allows us to AD through the observable function via symbolic indexing and accumulate and return grads against `sol`

```julia
julia> gs3 = gradient(sol) do sol
    sum(sol[sys.w])
end
((u = [[0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0]  …  [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0], [0.0, 1.0, 1.0, 1.0]], u_analytic = nothing, errors = nothing, t = nothing, k = nothing, prob = (f = nothing, u0 = nothing, tspan = nothing, p = ([0.0, 2990.0, 0.0],), kwargs = nothing, problem_type = nothing), alg = nothing, interp = nothing, dense = nothing, tslocation = nothing, stats = nothing, alg_choice = nothing, retcode = nothing, resid = nothing, original = nothing),)
```

This needs handling as part of when the observable symbol is in a collection (vector/ tuple/ ...), and also for various ADs like ReverseDiff and Enzyme. 

Ideally, this would be handled by removing all the adjoints related to `getindex` and let AD do the heavy lifting for us. But this is faster to implement in its current form.


ext/SciMLBaseZygoteExt.jl
```diff
@@ -182,15 +191,15 @@ end
...
-end
+# @adjoint function literal_getproperty(sol::AbstractTimeseriesSolution,
+#         ::Val{:u})
+#     function solu_adjoint(Δ)
+#         zerou = zero(sol.prob.u0)
+#         _Δ = @. ifelse(Δ === nothing, (zerou,), Δ)
+#         (build_solution(sol.prob, sol.alg, sol.t, _Δ),)
+#     end
+#     sol.u, solu_adjoint
+# end
```
ChrisRackauckas, line 202: Why is this removed?

ext/SciMLBaseZygoteExt.jl
```diff
@@ -122,6 +130,7 @@ end
 
 @adjoint function Base.getindex(VA::ODESolution{T}, sym::Union{Tuple, AbstractVector}) where T
     function ODESolution_getindex_pullback(Δ)
+        @show typeof(Δ)
```
ChrisRackauckas, line 133:
```suggestion
```

ext/SciMLBaseZygoteExt.jl
DhairyaLGandhi: It was returning the `ODESolution` as the adjoint. It is also an issue because it shortcuts the gradients through parameters and instead replaces it with the `sol.prob`, whereas we need to accumulate the gradients here.

DhairyaLGandhi:
Needs https://github.com/JuliaDiff/ChainRules.jl/pull/793

ext/SciMLBaseZygoteExt.jl
ChrisRackauckas: Can you add a unit test in the downstream set which shows this is fine?

ChrisRackauckas:
Add your unit tests as a new downstream testset.

ext/SciMLBaseZygoteExt.jl
DhairyaLGandhi: Happy to. In fact, that's why I asked if anything was relying on this behavior previously. Could you suggest what kind of test you have in mind?

ChrisRackauckas:
https://github.com/SciML/SciMLSensitivity.jl/blob/32f5ae7529a1957661b153f0ca9eff7e4caf0c5a/test/reversediff_output_types.jl#L14 this would hit it.

test/downstream/observables_autodiff.jl
```diff
@@ -0,0 +1,34 @@
+using ModelingToolkit, OrdinaryDiffEq
+using Zygote
+
+@parameters σ ρ β
```
ChrisRackauckas, line 4:
```suggestion
using ModelingToolkit, OrdinaryDiffEq
using Zygote
using ModelingToolkit: t_nounits as t, D_nounits as D

@parameters σ ρ β
```

DhairyaLGandhi:
Note that with SciMLSensitivity.jl#dg/ss (and https://github.com/SciML/SciMLStructures.jl/pull/18) https://github.com/SciML/SciMLSensitivity.jl/blob/32f5ae7529a1957661b153f0ca9eff7e4caf0c5a/test/reversediff_output_types.jl#L14 looks like:

```julia
julia> gs = gradient(u0 -> loss(u0), u0)
([-0.7779831009550049, 0.40028226620020263],)
```

DhairyaLGandhi:
I've added a DAE example in the tests, but switched it off until we get SciMLSensitivity updated as well. The DC motor example fails to initialize currently. If there's a different test case, I can also hook that in.

ext/SciMLBaseZygoteExt.jl
DhairyaLGandhi: I don't think this is what the error is referring to. I am missing a branch https://github.com/DhairyaLGandhi/RecursiveArrayTools.jl/tree/dg/noproj which removes an extra projection rule.

DhairyaLGandhi: It does refer to projecting to a `VectorOfArray`, and that rule wasn't defined for `Tangent`. Removing it gets us the expected results. If we want to project back to a `VectorOfArray` type, then that needs to be handled elsewhere.

Project.toml
```diff
@@ -68,6 +68,7 @@ Logging = "1.10"
 Makie = "0.20"
 Markdown = "1.10"
 ModelingToolkit = "8.75, 9"
+ModelingToolkitStandardLibrary = "2.7"
```
ChrisRackauckas, line 71: This should be gated int Downstream

DhairyaLGandhi: I've added bounds to `test/downstream/Project.toml` in https://github.com/SciML/SciMLBase.jl/pull/689/commits/940ea7820b6108c745bd30aee51a1cd1eb1d7461, should I remove anything from the regular test environment or do i need to declare these in both places?

DhairyaLGandhi: https://github.com/SciML/SciMLBase.jl/pull/689/commits/d061ce47f10d99f78850e08b820727c6228b3c8b does that

DhairyaLGandhi:
@ChrisRackauckas SciMLSensitivity test pass with https://github.com/SciML/SciMLBase.jl/pull/689/commits/d061ce47f10d99f78850e08b820727c6228b3c8b (latest commit), but the Core (Downstream) tests get cancelled before anything runs. Is that because the Core (Python) tests fail for unrelated reasons?

ext/SciMLBaseZygoteExt.jl
DhairyaLGandhi: now that we have restored the adjoint, I believe this can be resolved

gdalle:
So what happens here is:

- The latest ADTypes (v1.2) is installed in the test environment
- But for the Python test group, the environment resolution forces the previous ADTypes (v0.2)
- I think the following message refers to ADTypes
```
1 dependency precompiled but a different version is currently loaded. Restart julia to access the new version
```
- So the latest ADTypes is provided to the Python test group even though compatibility forbids it at the moment.

DhairyaLGandhi:
Both CI/ Python and CI/ Downgrade seem to be failing on master as well.

gdalle:
The problem I mentioned has not been fixed. It's not a problem with ADTypes per se, it's a problem with environment stacking

DhairyaLGandhi:
Is there anything left to be done in this PR?

PR #693, opened 2024-05-16
improve dtmin and unstable detection

oscardssmith:
Followup to https://github.com/SciML/SciMLBase.jl/pull/692. It turns out that OrdinaryDiffEq sets the minimum dt to `eps(t)` which is relatively sensible, but that means that we should be failing with an error if we don't accept a step and we're already at the minimum dt.

This definitely needs some tests, but I think those tests need to be in OrdinaryDiffEq so they will be added as a separate PR once this merges.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -634,7 +634,7 @@ function check_error(integrator::DEIntegrator)
             return ReturnCode.Unstable
         end
     end
-    bigtol = max(opts.reltol, opts.abstol)
+    bigtol = max(max(opts.reltol), max(opts.abstol))
     if isdefined(integrator, :EEst) && integrator.EEst * bigtol < .1
         if opts.unstable_check(integrator.dt, integrator.u, integrator.p,
                                           integrator.t)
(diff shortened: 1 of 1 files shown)


src/integrator_interface.jl
```diff
@@ -584,53 +584,68 @@ function check_error(integrator::DEIntegrator)
...
-    end
-    if integrator.opts.unstable_check(integrator.dt, integrator.u, integrator.p,
-        integrator.t)
-        bigtol = max(integrator.opts.reltol, integrator.opts.abstol)
-        # only declare instability if the dt is very small
-        # or we have at least one digit of accuracy in the solution
-        if integrator.dt<eps(integrator.t) || isdefined(integrator, :EEst) && integrator.EEst * bigtol < .1
-            if integrator.opts.verbose
+    end
+    bigtol = max(opts.reltol, opts.abstol)
```
ChrisRackauckas, line 637:
```suggestion
    bigtol = max(max(opts.reltol), max(opts.abstol))
```

? The array tol case seems to require more

oscardssmith: good point.

oscardssmith:
Re-running CI (now that it's morning so codecov will hopefully be happier and now that the OrdinaryDiffEq test that I think was the only real failure is fixed).

PR #703, opened 2024-06-05
fix: fix `u0` dependent on `p` not working with `remake`

### LINKED ISSUE #672: `remake` doesn't evaluate interdependencies between `u0` and `p`
### Made a comment about this in `Catalyst.jl`:

"Was just about to make an issue related to this, it's a particular problem when using `remake` to update the initial conditions of an `ODEProblem(rs, ...; remove_conserved = true)`. 

- This is because the Γ conservation law constants are defined as parameters, which have symbolic dependencies on the initial conditions.
- Perhaps an overload is necessary for the internal `SciMLBase.remake` function here: 
```julia
function _updated_u0_p_symmap(prob, u0, ::Val{true}, p, ::Val{true})
    isu0dep = any(symbolic_type(v) !== NotSymbolic() for (_, v) in u0)
    ispdep = any(symbolic_type(v) !== NotSymbolic() for (_, v) in p)

    if !isu0dep && !ispdep
        return remake_buffer(prob, state_values(prob), u0),
        remake_buffer(prob, parameter_values(prob), p)
    end
    if !isu0dep
        u0 = remake_buffer(prob, state_values(prob), u0)
        return _updated_u0_p_symmap(prob, u0, Val(false), p, Val(true))
    end
    if !ispdep
        p = remake_buffer(prob, parameter_values(prob), p)
        return _updated_u0_p_symmap(prob, u0, Val(true), p, Val(false))
    end

    varmap = merge(u0, p)
    u0 = anydict(k => symbolic_type(v) === NotSymbolic() ? v : symbolic_evaluate(v, varmap)
    for (k, v) in u0)
    p = anydict(k => symbolic_type(v) === NotSymbolic() ? v : symbolic_evaluate(v, varmap)
    for (k, v) in p)
    return remake_buffer(prob, state_values(prob), u0),
    remake_buffer(prob, parameter_values(prob), p)
end
```
   - Here the main issue I've found are that the `ispdep` check doesn't find that Γ in the parameters is a symbolic type, because it is evaluated to numerical values in the ODEProblem. The only place where the symbolic expressions for Γ are exposed are in `defaults` field of the `ODESystem`.
   - This obstructs the ostensible allowance for interdependencies between u0 and p, seen with the `varmap = merge(u0, p)` line. 
   - Currently my workaround is calling `symbolic_evaluate` prior to calling `remake` using the symbolic expressions from the defaults map:
   ```julia
   function compute_Γ(tunable_var_value_map, constraint_syms, sys_defs::T) where T

    #* Merge the default system definitions with the new symbolic map
    merged_defs = merge(sys_defs, Dict(tunable_var_value_map))

    NTuple{4, eltype(tunable_var_value_map)}(constraint => symbolic_evaluate(merged_defs[constraint], merged_defs) for constraint in constraint_syms)
end
```

I can make a separate issue for this in this repo (not sure if it pertains most here, or `ModelingToolkit.jl` or `SciMLBase.jl`) if this is truly a problem and I'm not missing anything."

_Originally posted by @jonathanfischer97 in https://github.com/SciML/Catalyst.jl/issues/806#issuecomment-2059826318_

I should mention that I'm using `ModelingToolkit.jl` version 8.75 rather than 9.0, due to compat with `Catalyst.jl`, if that makes a difference.

AayushSabharwal:
Close #672 

## Checklist

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -475,7 +475,10 @@ function _updated_u0_p_symmap(prob, u0, ::Val{true}, p, ::Val{false})
     isdep = any(symbolic_type(v) !== NotSymbolic() for (_, v) in u0)
     isdep || return remake_buffer(prob, state_values(prob), u0), p
 
-    temp_state = ProblemState(; p = p)
+    # FIXME: need to provide `u` since the observed function expects it.
+    # This is sort of an implicit dependency on MTK. The values of `u` won't actually be
+    # used, since any state symbols in the expression were substituted out earlier.
+    temp_state = ProblemState(; u = state_values(prob), p = p)
     u0 = anydict(k => symbolic_type(v) === NotSymbolic() ? v : getu(prob, v)(temp_state)
     for (k, v) in u0)
     return remake_buffer(prob, state_values(prob), u0), p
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Needs a test

PR #721, opened 2024-06-17
refactor: use `getu`/`setu` for all indexing

AayushSabharwal:
@TorkelE I moved your symbolic indexing tests here, added some more, and replaced `symbol_indexing.jl`. I've ensured everything in `symbol_indexing.jl` is still tested somewhere. Tests pass using MTK master

## Checklist


AayushSabharwal:
Eventually I'll move these to SII's downstream testset

oscardssmith:
this looks great! one question is now that all these getindexes just call getu, could they all be merged into a single method?

AayushSabharwal:
I'm happy to give it a shot, but defining these methods with a `Union` doesn't feel right to me. Semantically the fact that problems, integrators and solutions all support `getindex` is completely separate (they're three different interfaces) and basically just a coincidence.

Also, `setindex!` for problems uses `___internal_setindex!` for reasons I'm not aware of, so we'd just end up merging time-independent solutions and `DEIntegrator`.

TorkelE:
The tests have been quite heavily modified, so I cannot really tell if they are the same or no. If you say they are the same I trust you.

ChrisRackauckas:
How is this going?

AayushSabharwal:
The added tests pass locally last I checked, but the core CI is failing due to an error I haven't been able to reproduce yet (even with `--check-bounds=yes` and `--depwarn=yes`). That code hasn't been touched, so the sudden failure is weird.

PR #736, opened 2024-07-20
Skip special remakes on non-ODEFunction

ChrisRackauckas:
These all assume ODEFunction

@AayushSabharwal

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -104,7 +104,7 @@ function remake(prob::ODEProblem; f = missing,
 
     if f isa AbstractODEFunction && !(f isa ODEFunction)
         _f = f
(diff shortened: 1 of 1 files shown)


src/remake.jl
```diff
@@ -102,7 +102,9 @@ function remake(prob::ODEProblem; f = missing,
 
     iip = isinplace(prob)
 
-    if f === missing
+    if f isa AbstractODEFunction && !(f isa ODEFunction)
+        _f = f
+    elseif f === missing
```
ChrisRackauckas, line 107:
```suggestion
    elseif f === missing && prob.f isa ODEFunction
```

PR #748, opened 2024-07-24
Add Catalyst as a downstream test

isaacsas:
I'm not sure I've done this right, but I'd like to propose adding Catalyst as a downstream test. We have very comprehensive tests with V14, including for all kinds of symbolic indexing and plotting courtesy of @TorkelE, and I think some of the issues with stuff inadvertently getting broken might get caught if Catalyst was tested on such PRs to SciMLBase (and MTK, but Catalyst is already downstream there). It would also be very helpful to us in trying to diagnose sudden Catalyst breakages to be able to see if it had already broken in a merged SciMLBase PR.


isaacsas:
Note that Catalyst is currently capping SciMLBase at 2.44 on master due to the scalar plotting issue, but we will uncap it once that has been fixed and a release made.

PR #749, opened 2024-07-25
factor denseplot kwarg calculation to function

isaacsas:
This allows JumpProcesses to dispatch this function for SSAStepper, and hence get piecewise linear plotting when not saving every jump.


isaacsas:
Format failures seem to come from a number of untouched files, so I'm going to ignore them here and assume a separate PR will handle reformatting the repo.

PR #754, opened 2024-07-31
handle depwarn ambiguity

ChrisRackauckas:


ChrisRackauckas:
@thazhemadam I don't think this is doing formatter suggestions

AayushSabharwal:
#756 is a better fix

thazhemadam:
> @thazhemadam I don't think this is doing formatter suggestions

It will, with https://github.com/SciML/SciMLBase.jl/pull/757.

PR #762, opened 2024-08-08
Add `strip_solution`

jClugstor:
## Checklist

## Additional context

I know `error(...)` is discouraged, but I couldn't find one that fit, and wasn't sure if I should make one.

This also doesn't completely strip solutions made from implicit methods, since the caches still need to be dealt with.


ChrisRackauckas (approved):

ChrisRackauckas:
Setup downstream tests

src/solutions/ode_solutions.jl
```diff
@@ -604,3 +604,17 @@ function sensitivity_solution(sol::ODESolution, u, t)
...
     return @set sol.interp = interp
 end
+
+function strip_solution(sol::ODESolution)
+    if has_lazy_interpolation(sol.alg)
+        error("The algorithm $(sol.alg) uses lazy interpolation, which is incompatible with
+        solution stripping.")
+    end
+
+    interp = strip_interpolation(sol.interp)
```
thomvet, line 614: @jClugstor : Just randomly reviewing this. It seems strange to me that the `interp` variable isn't used afterwards? Should it go into line 618?

ChrisRackauckas: Yes that looks like a typo.

ChrisRackauckas:
@jClugstor what's the next steps for this?

src/solutions/ode_solutions.jl
jClugstor: Yeah, it's a typo, thanks for pointing it out .

jClugstor:
This will need to check if the solution was made using an implicit method that has caches, if it was I'll need to strip the cache. Should I make a trait for SciMLAlgs e.g. 'has_cache' ?

PR #766, opened 2024-08-19
Add `strip_cache` for DECache

jClugstor:
## Checklist

- [ x] Appropriate tests were added

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -615,6 +615,6 @@ function strip_solution(sol::ODESolution)
 
     ODESolution(sol.u, sol.u_analytic, sol.errors,
         sol.t, sol.k, sol.discretes, nothing, nothing,
-        sol.interp, sol.dense, sol.tslocation, sol.stats,
+        interp, sol.dense, sol.tslocation, sol.stats,
         sol.alg_choice, sol.retcode, sol.resid, sol.original)
 end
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Create a test that catches this.

ChrisRackauckas:
why is this needed here? Isn't it just internal to OrdinaryDiffEq?

jClugstor:
I wasn't sure if there were other interpolation types in SciMLBase that would have caches. 
Is it better to just put 
```julia
function strip_cache(cache::OrdinaryDiffEqCache)
    # most caches don't have jac_config or grad_config
    cache
end
```
in to OrdinaryDiffEq?

ChrisRackauckas:
You define a special strip_interpolation in OrdianryDiffEq, and that will have a function in there to strip caches, and then that is defined as just a helper which uses knowledge of how caches in OrdinaryDiffEq work. I don't see that as generic at all though.

jClugstor:
I was thinking something like this:
```julia
function SciMLBase.strip_interpolation(id::InterpolationData)
    cache = id.cache
    
    cache = strip_cache(cache)

    InterpolationData(nothing, id.timeseries,
        id.ts,
        id.ks,
        id.alg_choice,
        id.dense,
        cache,
        id.differential_vars,
        id.sensitivitymode)
end

function strip_cache(cache::OrdinaryDiffEqCache)
    # most caches don't have jac_config or grad_config
    cache
end
```
and then I can define `strip_cache` specifically for the cache types that need `jac_config` and `grad_config` to be taken out.

PR #770, opened 2024-08-26
Fix solution stripping for ODESolution

jClugstor:
## Checklist

- [ x] Appropriate tests were added
  
## Additional context
Adds a test for stripping ODESolution objects of any function information.


jClugstor:
It feels hacky to rely on `typeof(sol).parameters` to build the new `ODESolution`, but it seems it's needed?

src/solutions/ode_solutions.jl
```diff
@@ -613,7 +613,9 @@ function strip_solution(sol::ODESolution)
 
     interp = strip_interpolation(sol.interp)
 
-    ODESolution(sol.u, sol.u_analytic, sol.errors,
+    bigT = typeof(sol).parameters[1]
+    bigN = typeof(sol).parameters[2]
```
ChrisRackauckas, line 617: This is not a good idea, it will not infer.

ChrisRackauckas: Instead use `@set sol.interp = interp`

jClugstor: Where does `@set` come from? When I do 
```julia
 @set sol.interp = interp
   
    ODESolution(sol.u, sol.u_analytic, sol.errors,
        sol.t, sol.k, sol.discretes, nothing, nothing,
        sol.interp, sol.dense, sol.tslocation, sol.stats,
        sol.alg_choice, sol.retcode, sol.resid, sol.original)
```
It still gives a method error.

jClugstor: Would something like this be better?
```julia
N = sol.u === nothing ? 2 : ndims(eltype(sol.u)) + 1
T = eltype(eltype(sol.u))
```

jClugstor: Oh I see, from Accessors

jClugstor:
Ok, this should be it now :-)

PR #773, opened 2024-08-28
Update optimization_problems.jl

johannahaffner:
This is inconsistent with the description below, which states that the array must be sized to match the number of constraints.

I also cannot think of a reason why the constraint expressions should match the geometry of the minimisation variable (and many cases where this would not be true), so I am assuming that this is an erroneous half-sentence that can be removed.
  
## Additional context

This is a little confusing, it also contradicts the documentation lines 74-77 and 81 and following.


Vaibhavdixit02:
You are right, probably a vestige of some copy paste, thanks!

PR #775, opened 2024-08-30
add `alg_interpretation` trait

prbzrg:
We need this for https://github.com/SciML/SciMLSensitivity.jl/pull/1096

### DIFF
--- a/src/alg_traits.jl
+++ b/src/alg_traits.jl
@@ -243,8 +243,13 @@ as the maximum order of the algorithm.
 function alg_order(alg::AbstractODEAlgorithm)
     error("Order is not defined for this algorithm")
 end
+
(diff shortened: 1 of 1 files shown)


src/alg_traits.jl
```diff
@@ -243,3 +243,5 @@ as the maximum order of the algorithm.
 function alg_order(alg::AbstractODEAlgorithm)
     error("Order is not defined for this algorithm")
 end
+
+function alg_interpretation end
```
ChrisRackauckas, line 247: needs a docstring

prbzrg: There wasn't any docstring for it in `StochasticDiffEq`. What should we put?

PR #776, opened 2024-08-30
allow different number of out-of-place args for isinplace

ranocha:
## Checklist

## Additional context

This helps us at https://github.com/SKopecz/PositiveIntegrators.jl

### DIFF
--- a/src/alg_traits.jl
+++ b/src/alg_traits.jl
@@ -245,16 +245,17 @@ function alg_order(alg::AbstractODEAlgorithm)
 end
 
 """
-Integral interpolation for the SDE solver algorithm. SDEs solutions depend on the chosen definition of the stochastic integral. In the Ito calculus, 
-the left-hand rule is taken, while Stratonovich calculus uses the right-hand rule. Unlike in standard Riemannian integration, these integral rules do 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
What's this for?

ranocha:
For PositiveIntegrators.jl, we need to partition the ODE right-hand side into (positive) production and (negative) destruction terms. Up to now, we have used two functions for this that can either be in-place (like `production!(production_terms, u, p, t)`) or out-of-place (like `production_terms = production(u, p, t)`). Thus, we could use `isinplace` as it is. To improve the performance of some applications, we want to use a single function with in-place form `rhs!(production_terms, destruction_terms, u, p, t)` and out-of-place form `production_terms, destruction_terms = rhs(u, p, t)`. With the current version of `isinplace`, it's hardcoded that the out-of-place form has exactly one argument less than the in-place form. Thus, we cannot use it as it is right now for our new desired dispatches.

ranocha:
The downgrade action fails already on `master`

ranocha:
Is that something we can support with SciMLBase.jl, @ChrisRackauckas? The only other option we have is to replicate code in PositiveIntegrators.jl.

ChrisRackauckas (approved):

ranocha:
Thanks!

PR #790, opened 2024-09-19
Fix interpolation for stripped solutions

jClugstor:
## Checklist

- [x ] Appropriate tests were added
- [x ] Any code changes were done in a way that does not break public API
- [x ] All documentation related to code changes were updated
- [x ] The new code follows the
- [x ] Any new documentation only uses public API
  
## Additional context
Fixes interpolation for stripped solution objects, i.e. solution objects from `strip_solution`.

I'm not sure about doing `struct ODENullFunction <: SciMLBase.AbstractODEFunction{false} end`, but the inner constructor for ODEProblem doesn't allow `f` to be nothing. And the alternative seems to be changing the functors for ODEProblem. This also allows for stripped_solutions to keep some problem information. It might be good to add taking out the whole problem as an option as well?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
-version = "2.53.1"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Please update the downstream tests with this.

PR #808, opened 2024-10-07
fix: use `split = false` system for remake autodiff tests

AayushSabharwal:
## Checklist


ChrisRackauckas:
why split=false?

AayushSabharwal:
Two reasons:
1. The tests were written assuming a parameter vector. They actually replaced the `MTKParameters` object with an `Array` via `remake`. The test working was pure coincidence with MTK internals and because the `MTKParameters` object only had a tunables vector.
2. https://github.com/SciML/SciMLSensitivity.jl/issues/1130

PR #809, opened 2024-10-07
feat: support symbolic indexing of a subset of the system

AayushSabharwal:
Draft while I add more tests

## Checklist

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -615,7 +615,7 @@ function solution_slice(sol::ODESolution{T, N}, I) where {T, N}
     @reset sol.u_analytic = sol.u_analytic === nothing ? nothing : sol.u_analytic[I]
     @reset sol.t = sol.t[I]
     @reset sol.k = sol.dense ? sol.k[I] : sol.k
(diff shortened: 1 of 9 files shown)


src/solutions/ode_solutions.jl
```diff
@@ -180,6 +184,58 @@ function SymbolicIndexingInterface.is_parameter_timeseries(::Type{S}) where {
...
+    T8,
+    T9,
+    T10,
+    T11,
+    T12,
+    T13,
+    T14,
+    T15,
+    T16} where {
+    T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16 <: SavedSubsystem}
```
ChrisRackauckas, line 203: this seems a bit fragile. Why not a trait function?

ChrisRackauckas:
This should roll out to all solution types?

src/solutions/ode_solutions.jl
AayushSabharwal: Yeah I guess that is a better option

AayushSabharwal:
Yes. I'll add the rest.

PR #813, opened 2024-10-09
refactor: remove Tables.jl impl, fall back to RecursiveArrayTools

### LINKED ISSUE #798: Names of unknowns have a `(t)` appendix but DataFrame column names not
**A nice to have: consistency between `unknown` names and `DataFrame` column names**

The names of unknowns/variables of an `ODESystem` come with a `(t)` appendix. However, when transforming the solution to a `DataFrame` the `(t)` is gone. This way one cannot directly use the name of a  variable as an index.
It would be helpful to make naming consistent.

```julia
using ModelingToolkit
using DataFrames
using OrdinaryDiffEq

const t = ModelingToolkit.t_nounits
const D = ModelingToolkit.D_nounits

@variables x(t)=1.0
eqn = [D(x) ~ 0.0]

@named sys = ODESystem(eqn,t,[x],[],tspan=(0, 10.0))
sys = structural_simplify(sys)
prob = ODEProblem(sys)
sol = solve(prob)

dfsys = DataFrame(sol)

String(Symbol(unknowns(sys)[1]))
names(dfsys)[2]

```

**Package Versions**

```julia
Status `~/.julia/environments/v1.10/Project.toml`
  [13f3f980] CairoMakie v0.12.11
⌃ [a93c6f00] DataFrames v1.6.1
⌃ [0c46a032] DifferentialEquations v7.13.0
  [86223c79] Graphs v1.11.2
  [23992714] MAT v0.10.7
⌃ [961ee093] ModelingToolkit v9.40.0
  [769b91e5] Neuroblox v0.4.2 `../../../Projects/neuroblox/codes/Neuroblox.jl`
  [bac558e1] OrderedCollections v1.6.3
⌃ [1dea7af3] OrdinaryDiffEq v6.87.0
⌃ [91a5bcdd] Plots v1.40.1
⌃ [c3e4b0f8] Pluto v0.19.45
```

Thank you!

AayushSabharwal:
Close #798 

## Checklist


AayushSabharwal:
I think the `isrowtable` tests should have been removed with RAT@3, since the relevant documentation says:

> Requirements for defining isrowtable include:
> - `Tables.rows(x) === x`, i.e. the table object itself is a Row iterator
> - If the table object is mutable, it should support:
>   - `push!(x, row)`: allow pushing a single row onto table
>   - `append!(x, rows)`: allow appending set of rows onto table
> - If table object is mutable and indexable, it should support:
>  - `x[i] = row`: allow replacing of a row with another row by index

The third point is not valid for `AbstractDiffEqArray`

ChrisRackauckas:
good point

ChrisRackauckas:
There's no PR to RecursiveArrayTools though?

AayushSabharwal:
It already has the interface, and doesn't define `isrowtable`

AayushSabharwal:
In fact the code I removed here is almost identical to that in RAT

PR #844, opened 2024-11-04
Clarify nlprob docs for ODEFunction

termi-official:
I think I get the idea about what needs to be done next. To make sure we are all on the same page, let me try to clarify the docs here.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -300,7 +300,7 @@ the usage of `f`. These include:
   variables are some compatible vectors set by the specific solver.
   Note that other implicit techniques, like for example some fully-implicit Runge-Kutta methods,
   need to solve different nonlinear systems.
-  Generally the inner nonlinear function of the nonlinear problem is in general of the form `g(z,p') = 0`
+  The inner nonlinear function of the nonlinear problem is in general of the form `g(z,p') = 0`
   where `p'` is a struct with all information about the specific nonlinear problem at hand to solve
   for a specific time discretization. For example, for the nonlinear problem stated in the first
   paragraph in this docstring, `p'` would hold `(dt, γ, inner_tmp, outer_tmp, t, p)`, such that 


src/scimlfunctions.jl
```diff
@@ -289,10 +290,16 @@ the usage of `f`. These include:
...
-- `nlprob`: a `NonlinearProblem` that solves `f(u, t, p) = u_tmp`
-  where the nonlinear parameters are the tuple `(t, u_tmp, p)`.
-  This will be used as the nonlinear problem inside an implicit solver by specifying `u, u_tmp` and `t`
-  such that solving this function produces a solution to the implicit step of your solver.
+- `nlprob`: some `AbstractNonlinearProblem` to define custom nonlinear problems to be used for
+  implicit time discretizations. This allows to use extra structure of the ODE function (e.g.
+  multi-level structure). The rationale here is essentially that implicit ODE integration algorithms
+  need do solve the a nonlinear problems, usually of the form `dt⋅f(γ⋅z+inner_tmp,p,t) + outer_tmp = z`,
+  during time integration, where users want to use extra structure of f during the nonlinear solve.
+  Here `z` is the stage solution vector, `p` is the parameter vector of the ODE problem, `t` is
```
ChrisRackauckas, line 298: how are dt, inner_tmp, outer_tmp, and z defined? in `p`? Where in `p` is it assumed? Or does it match some interface?

Aren't you missing a function that updates the prob based on those values?

termi-official: Good point. Tried to resolve this in the latest commit.

ChrisRackauckas: I don't see how that addresses it. So okay, how do I update `inner_tmp`?

termi-official: I do not see how that logic changes in OrdinaryDiffEqNonlinearSolve.jl. From my current understanding the specific ode integrator at hand passes some struct with e.g. `inner_tmp` into the `nlprob` (e.g. via remake?) to update these during setup and remake.

termi-official: Okay that struct is currently just a tuple. I specifically refer to https://github.com/SciML/OrdinaryDiffEq.jl/blob/1d05b16b9bc52c31f5ec27e347b77d9e5306bee1/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl#L46-L47 where the tmp variable comes from the nonlinear solver struct in https://github.com/SciML/OrdinaryDiffEq.jl/blob/1d05b16b9bc52c31f5ec27e347b77d9e5306bee1/lib/OrdinaryDiffEqNonlinearSolve/src/type.jl#L216-L225

ChrisRackauckas: Okay is that the interface? If you don't build the nlprob such that it is true will its integrations break?

termi-official: > Okay is that the interface?

From my understanding. Can you confirm or clarify if I got some detail wrong @oscardssmith ?

> If you don't build the nlprob such that it is true will its integrations break?

I am pretty sure that it will break the integration. Especially if we stick to the tuple instead of a struct and change the solver to something which requires a slightly different tuple, which is compatible by accident.

ChrisRackauckas: Tuple can be quite limiting though. Why would we not have a functional interface rather than specifying a data type?

oscardssmith: We probably should have a better interface (a `NamedTuple` at least would be a step up)

oscardssmith: I'm not sure a functional interface buys us much here since the only things the user is expected to do is get the elements out...

ChrisRackauckas: MTK cannot easily generate a NamedTuple. An interface via a function would make it easy to generate via SII

termi-official: > [...] functional interface [...]

Functional interface as in _non-defective interface_ or as in _function programming_?

Regarding the tuples, I think we are on the same page then. I see two main advantage of having structs to control the dispatches for the initialization of different nonlinear solver classes (over having tuples and hard-coded if-else blocks on types):
1. Being extensible for custom solvers for specific problem classes
2. Being less error prone to changes, e.g. when there might be some decision in the future to add another possibility to the values of `method`.

PR #857, opened 2024-11-11
feat: add fields to `OverrideInit`, better `nlsolve_alg` handling 

AayushSabharwal:
This replaces `default_nlsolve` in OrdinaryDiffEq, and will have methods in NonlinearSolve which can then be leveraged by all libraries for initialization.

## Checklist

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -182,7 +182,11 @@ function get_initial_values(prob, valp, f, alg::OverrideInit,
         initdata.update_initializeprob!(initprob, valp)
     end
 
-    nlsol = solve(initprob, nlsolve_alg; abstol = alg.abstol)
+    if alg.abstol !== nothing
+        nlsol = solve(initprob, nlsolve_alg; abstol = alg.abstol)
+    else
+        nlsol = solve(initprob, nlsolve_alg)
+    end
 
     u0 = initdata.initializeprobmap(nlsol)
     if initdata.initializeprobpmap !== nothing
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -68,6 +68,15 @@ function Base.showerror(io::IO, e::OverrideInitMissingAlgorithm)
         "OverrideInit specified but no NonlinearSolve.jl algorithm provided. Provide an algorithm via the `nlsolve_alg` keyword argument to `get_initial_values`.")
 end
 
+struct OverrideInitNoTolerance <: Exception
+    tolerance::Symbol
+end
+
+function Base.showerror(io::IO, e::OverrideInitNoTolerance)
+    print(io,
+        "Tolerances were not provided to `OverrideInit`. `$(e.tolerance)` must be provided as a keyword argument to `get_initial_values` or as a keyword argument to the `OverrideInit` constructor.")
+end
+
 """
 Utility function to evaluate the RHS of the ODE, using the integrator's `tmp_cache` if
 it is in-place or simply calling the function if not.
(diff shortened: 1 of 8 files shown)


src/initialization.jl
```diff
@@ -171,15 +171,16 @@ function get_initial_values(prob, valp, f, alg::OverrideInit,
...
+    if nlsolve_alg === nothing && state_values(initprob) !== nothing
         throw(OverrideInitMissingAlgorithm())
     end
 
     if initdata.update_initializeprob! !== nothing
         initdata.update_initializeprob!(initprob, valp)
     end
 
-    nlsol = solve(initprob, nlsolve_alg)
+    nlsol = solve(initprob, nlsolve_alg; abstol = alg.abstol)
```
ChrisRackauckas, line 183: it should default to using the tolerance from the solver if not supplied elsewhere.

src/initialization.jl
```diff
@@ -100,7 +100,8 @@ Check if the algebraic constraints are satisfied, and error if they aren't. Retu
 the `u0` and `p` as-is, and is always successful if it returns. Valid only for
 `ODEProblem` and `DAEProblem`. Requires a `DEIntegrator` as its second argument.
 """
-function get_initial_values(prob::ODEProblem, integrator, f, alg::CheckInit,
+function get_initial_values(
+        prob::AbstractDEProblem, integrator::DEIntegrator, f, alg::CheckInit,
```
ChrisRackauckas, line 104: This doesn't handle DAEProblem?

src/initialization.jl
```diff
@@ -109,7 +110,7 @@ function get_initial_values(prob::ODEProblem, integrator, f, alg::CheckInit,
 
     algebraic_vars = [all(iszero, x) for x in eachcol(M)]
     algebraic_eqs = [all(iszero, x) for x in eachrow(M)]
-    (iszero(algebraic_vars) || iszero(algebraic_eqs)) && return
+    (iszero(algebraic_vars) || iszero(algebraic_eqs)) && return u0, p, true
     update_coefficients!(M, u0, p, t)
     tmp = _evaluate_f_ode(integrator, f, isinplace, u0, p, t)
     tmp .= ArrayInterface.restructure(tmp, algebraic_eqs .* _vec(tmp))
```
ChrisRackauckas, line 116: `integrator.opts.internalnorm`: This doesn't handle Sundials or ODEInterface?

src/initialization.jl
AayushSabharwal: There's an `AbstractDAEProblem` dispatch below

ChrisRackauckas: But then this one should just be for ODEProblem?

AayushSabharwal: SDEProblem dispatches here too

src/alg_traits.jl
```diff
@@ -295,3 +295,34 @@ initialization.
...
+end
+
+"""
+    $(TYPEDSIGNATURES)
+
+Get the `reltol` for the solve stored in the cache. Required for caches of algorithms for
+which `cache_stores_tolerances(alg) == true`.
+"""
+function cache_get_reltol(cache)
+    error("Cache does not store the `reltol`.")
```
ChrisRackauckas, line 327: I thought we talked about passing it?

AayushSabharwal: Didn't you also mention having a trait and throwing an error message? Right now it takes a keyword argument and falls back to this if not provided.

AayushSabharwal: That said I'm more than happy to remove this trait since solvers have to individually opt-in to doing initialization anyway.

ChrisRackauckas: Let's simplify and just make it required.

AayushSabharwal:
A lot of our tests still use codecov, apparently 😅

AayushSabharwal:
This is good to go. Downstream failures are BVDiffEq failing to precompile.

ChrisRackauckas (approved):

PR #860, opened 2024-11-13
fix nlprob to match the initialization system

oscardssmith:
this adds the docs from https://github.com/SciML/SciMLBase.jl/pull/844 and adds all the extra pieces needed to make this the same as what the initialization system uses.

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -289,11 +289,6 @@ the usage of `f`. These include:
   based on the sparsity pattern. Defaults to `nothing`, which means a color vector will be
   internally computed on demand when required. The cost of this operation is highly dependent
   on the sparsity pattern.
-- `nlprob`: a `NonlinearProblem` that solves `f(u, t, p) = u_tmp`
-  where the nonlinear parameters are the tuple `(t, u_tmp, p)`.
-  This will be used as the nonlinear problem inside an implicit solver by specifying `u, u_tmp` and `t`
-  such that solving this function produces a solution to the implicit step of your solver.
-
 ## iip: In-Place vs Out-Of-Place
 
 `iip` is the optional boolean for determining whether a given function is written to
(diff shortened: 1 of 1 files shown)


oscardssmith:
I matched the initialization system

oscardssmith:
oh @ChrisRackauckas, the reason you were confused is that I forgot to add the new file.

ChrisRackauckas:
no import?

oscardssmith:
fixed

src/ODE_nlsolve.jl
```diff
@@ -0,0 +1,51 @@
...
+    the parameter object of the original problem. If absent (`nothing`),
+    this will not be called and the parameters of the problem being
+    solved will be returned as-is.
+    """
+    nlprobpmap::NLProbPmap
+
+    function ODE_NLProb(nlprob::I, update_nlprob!::J, nlprobmap::K,
+            nlprobpmap::L) where {I, J, K, L}
+        return new{I, J, K, L}(nlprob, update_nlprob!, nlprobmap, nlprobpmap)
+    end
```
ChrisRackauckas, line 49: Is this needed?

src/scimlfunctions.jl
```diff
@@ -424,7 +419,7 @@ struct ODEFunction{iip, specialize, F, TMM, Ta, Tt, TJ, JVP, VJP, JP, SP, TW, TW
     colorvec::TCV
     sys::SYS
     initialization_data::ID
-    nlprob::NLP
+    nlprob_data::NLP
```
ChrisRackauckas, line 422: We should probably type restrict both of these?

src/ODE_nlsolve.jl
oscardssmith: probably not.

src/scimlfunctions.jl
oscardssmith: yeah, I was copying `initialization_data`. Should we restrict both?

ChrisRackauckas: yes

src/ODE_nlsolve.jl
ChrisRackauckas: delete it?

ChrisRackauckas (approved):

ChrisRackauckas:
Issues with include order seen in tests.

oscardssmith:
yeah. the issue is that the fields can't be type restricted since there otherwise is no order the files can be ordered in that avoids circular dependency

PR #862, opened 2024-11-14
AutoSpecialize for NonlinearProblem

oscardssmith:
since `NonlinearProblem` doesn't have a `t` field, most of the time, we can just use `Void` for all of the functions of inplace ones without a performance penalty.

### DIFF
--- a/src/problems/nonlinear_problems.jl
+++ b/src/problems/nonlinear_problems.jl
@@ -104,6 +104,9 @@ function IntervalNonlinearProblem(f, tspan, p = NullParameters(); kwargs...)
     IntervalNonlinearProblem(IntervalNonlinearFunction(f), tspan, p; kwargs...)
 end
 
+
+_default_nl_specialize(p) = sizeof(p)==0 || ismutable(p) ? AutoSpecialize : FullSpecialize
+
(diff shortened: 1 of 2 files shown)


oscardssmith:
interestingly, even with this, NonlinearSolve is still 90% compilation time...

ChrisRackauckas:
@avik-pal test it a bit?

avik-pal:
> interestingly, even with this, NonlinearSolve is still 99% compilation time...

Do you mean the current precompile workload? I think it manually sets the specialization so it has no effect.

Also we would need the corresponding DiffEqBase update for this to work I think

oscardssmith:
I changed the workload to make `p=nothing` and `AutoSpecialize` (since it is going to be hard to precompile meaningfully for non-null parameters). What DiffEqBase changes would be needed?

avik-pal:
https://github.com/SciML/DiffEqBase.jl/pull/1101/files

oscardssmith:
That very specifically shouldn't be needed. The key observation of this PR is that you don't actually need to FunctionWrap in place functions since you can just do a dynamic dispatch instead.

PR #876, opened 2024-11-23
Add MaxNumSub return code

ErikQQY:
Part of https://github.com/SciML/BoundaryValueDiffEq.jl/issues/257

Add `MaxNumSub` return code to indicate the maximum number of subintervals are exceeded during the adaptivity process and throw proper warning


src/retcodes.jl
```diff
@@ -149,6 +149,26 @@ EnumX.@enumx ReturnCode begin
...
+
+    A failure exit state of the solver. If this return code is given, then the
+    solving process was unsuccessful and exited early because during the solver's
+    adaptivity, mesh length exceeded the `max_num_subintervals` either set by default or specified
+    by users in the solver.
+
+    ## Common Reasons for Seeing this Return Code
+
+      - This commonly occurs in BVP solving if the original mesh are too coarse or
+        the tolerance are too stringent. It is recommended that in such cases, one tries to increase the default `max_num_subintervals`
```
ChrisRackauckas, line 163: Isn't this maxiters?

ErikQQY: No, `MaxIters` is used for the exit because the default iteration number exceeded, say, number of Newton iterations, but for cases when solving ill-conditioned problems with adaptivity(we redistribute or halve the mesh), the mesh would become so large that we need to warn users instead of just throw `Failure`. Although the reasons for `MaxIters` and `MaxNumSub` are similar, tolerance is stringent or nonstiff solvers on stiff problems, etc, I believe these failure statuses are two different things.

PR #891, opened 2024-12-11
feat: add new `remake(::AbstractSciMLFunction)`, fix some `remake` bugs.

AayushSabharwal:
## Checklist

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -122,19 +122,23 @@ end
 by keyword arguments. For stochastic functions (e.g. `SDEFunction`) the `g` keyword argument
 is used to override `func.g`. For split functions (e.g. `SplitFunction`) the `f2` keyword
 argument is used to override `func.f2`, and `f` is used for `func.f1`. If
-`f isa AbstractSciMLFunction`, properties of `f` will override those of `func` (but not ones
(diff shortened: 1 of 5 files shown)


AayushSabharwal:
Formatting issues are on master, will make a separate PR to fix them to avoid confusing the intent of this one.

src/remake.jl
```diff
@@ -99,6 +99,105 @@ function remake(
...
+in `a`, except if they are `nothing`. Keys not present in `a` are assumed to have
+a value of `nothing`.
+"""
+function _similar_namedtuple_merge_ignore_nothing(a::NamedTuple, b::NamedTuple)
+    ks = fieldnames(typeof(b))
+    return NamedTuple{ks}(ntuple(Val(length(ks))) do i
+        something(get(b, ks[i], nothing), get(a, ks[i], nothing), Some(nothing))
+    end)
+end
+
```
ChrisRackauckas, line 116: isn't this just `(;a..., b...)`?

AayushSabharwal: No, because `nothing` values in `b` will override values in `a`. This function makes it so that if a key has the value `nothing` in `b`, it uses the value from `a`.

PR #896, opened 2024-12-22
Make error check more compatible with SciML interface

termi-official:
This will require some changed in many downstream packages. The main issue for now is that `SciMLBase.last_step_failed` never returns true if the integrator is adaptive.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -169,6 +169,13 @@ function set_proposed_dt!(i::DEIntegrator, dt)
 end
 
 """
+    get_tdir(i::DEIntegrator)
+
+Get the time direction of the integrator. Should return 1 or -1 with the same type as the time type of the integrator.
+"""
+get_tdir(i::DEIntegrator) = i.tdir
+
+"""
     savevalues!(integrator::DEIntegrator,
       force_save=false) -> Tuple{Bool, Bool}
 
@@ -406,6 +413,15 @@ function get_sol(integrator::DEIntegrator)
     return integrator.sol
 end
 
+"""
+    get_retcode(integrator::DEIntegrator)
+
+Get the return code of the integrator.
+"""
+function get_retcode(integrator::DEIntegrator)
+    return integrator.sol.retcode
+end
+
 ### Addat isn't a real thing. Let's make it a real thing Gretchen
 
 function addat!(a::AbstractArray, idxs, val = nothing)
@@ -564,6 +580,15 @@ end
 
 last_step_failed(integrator::DEIntegrator) = false
 
+# Standard error message for classical PID type controllers
+function controller_message_on_dtmin_error(integrator::DEIntegrator)
+    if isdefined(integrator, :EEst)
+       return ", and step error estimate = $(integrator.EEst)"
+    else
+        return ""
+    end
+end
+
 """
     check_error(integrator)
 
@@ -571,11 +596,11 @@ Check state of `integrator` and return one of the
 [Return Codes](https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#retcodes)
 """
 function check_error(integrator::DEIntegrator)
-    if integrator.sol.retcode ∉ (ReturnCode.Success, ReturnCode.Default)
-        return integrator.sol.retcode
+    if get_retcode(integrator) ∉ (ReturnCode.Success, ReturnCode.Default)
+        return get_retcode(integrator)
     end
     opts = integrator.opts
-    verbose = opts.verbose
+    verbose = hasproperty(opts, :verbose) && opts.verbose
     # This implementation is intended to be used for ODEIntegrator and
     # SDEIntegrator.
     if isnan(integrator.dt)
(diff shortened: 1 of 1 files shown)


src/integrator_interface.jl
```diff
@@ -168,6 +168,13 @@ function set_proposed_dt!(i::DEIntegrator, dt)
     error("set_proposed_dt!: method has not been implemented for the integrator")
 end
 
+"""
+    get_tdir(i::DEIntegrator)
+
+Get the time direction of the integrator. Should return 1 or -1 with the same type as the time type of the integrator.
+"""
+get_tdir(i::DEIntegrator) = i.tdir
```
termi-official, line 176:
```suggestion
function get_tdir(i::DEIntegrator)
    if hasproperty(i, :tdir)
        i.tdir
    else
        error("get_tdir: method has not been implemented for the integrator")
    end
```

src/integrator_interface.jl
```diff
@@ -406,6 +413,15 @@ function get_sol(integrator::DEIntegrator)
...
 end
 
+"""
+    get_retcode(integrator::DEIntegrator)
+
+Get the return code of the integrator.
+"""
+function get_retcode(integrator::DEIntegrator)
+    return integrator.sol.retcode
+end
```
termi-official, line 423: To quickly elaborate on this, I essentially have problems where "time sub-integrators" advance parts of the solution and I could not find a sane way to define this distributed solution stuff into the current solution interface. Technically I do not even need every subintegrator to hold a solution. They just need to communicate return codes between each other. Having this interface here would be a compromise.

src/integrator_interface.jl
```diff
@@ -564,18 +580,27 @@ end
...
 last_step_failed(integrator::DEIntegrator) = false
 
+# Standard error message for classical PID type controllers
+function controller_message_on_dtmin_error(integrator::DEIntegrator)
+    if isdefined(integrator, :EEst)
+       return ", and step error estimate = $(integrator.EEst)"
+    else
+        return ""
+    end
+end
```
termi-official, line 590: Some voodoo time adaption algorithms actually work without defining an error estimate, so I would like to remove the dependency that the integrator itself needs to carry the error estimate and would like to propose in the future that there is either some kind of ControllerCache or the contoller itself carries the estimate.

src/integrator_interface.jl
```diff
@@ -597,42 +622,35 @@ function check_error(integrator::DEIntegrator)
     # We also exit if the ODE is unstable according to a user chosen callback
     # but only if we accepted the step to prevent from bailing out as unstable
     # when we just took way too big a step)
-    step_accepted = !hasproperty(integrator, :accept_step) || integrator.accept_step
-    if !opts.force_dtmin && opts.adaptive
-        if abs(integrator.dt) <= abs(opts.dtmin) &&
-           (!step_accepted || (hasproperty(opts, :tstops) ?
-             integrator.t + integrator.dt < integrator.tdir * first(opts.tstops) :
-             true))
+    step_rejected = last_step_failed(integrator)
```
termi-official, line 625: Right now this does not work, as `last_step_failed` is always false if the integrator is adaptive. If everyone agrees that this is closer to the semantics that we really want, then I will go through all subpackages and make the corresponding change.

src/integrator_interface.jl
ChrisRackauckas: Most integrators don't set the return code until the end though?

src/integrator_interface.jl
ChrisRackauckas: Yeah... that's going to be rough because there needs to be a tie-in in every stepper for how to calculate and store EEst, if it's needed. I'm not opposed though, because yes some methods like a priori estimates don't need the EEst and so it shouldn't spend the time to calculate it.

src/integrator_interface.jl
ChrisRackauckas: I think this could be reasonable.

src/integrator_interface.jl
termi-official: Correct. But for operator splitting methods I have essentially a tree of integrators, each solving subproblems on small time intervals. The easiest example is essentially to have an additivley split right hand side for some ODE: 

$d_t u = f(u,p,t) = f_1(u,p,t) + f_2(u,p,t)$

which I want to integrate from [0,T]. Operator splitting method now definine two ODEs

$d_t u^* = f_1(u^*,p,t)$

and

$d_t u^{**} = f_2(u^{**},p,t)$

and they intrgrate each of these in a specific order on subintervals on $[0,T]$. The simplest rule is to solve them alternatingly on intervals with fixed dt. E.g. the first step would be

1. Set $u^*(0) = u_0$
2. Solve $d_t u^* = f_1(u^*,p,t)$ on $[0,dt]$ (with your algorithm of choice)
3. Set $u^{**}(0) = u^*(dt)$
4. Solve $d_t u^{**} = f_2(u^{**},p,t)$ on $[0,dt]$ (with your algorithm of choice)


In my implementation right now I have three separate integrators. One custom time integrator for the coordination of which subintegrator needs to integrate its associated problem next (+ time step controll of the time intervals to solve the subproblems on), and another two integrators for the respective subproblems. To check whether a solve worked or not the outer time integrator essentially checks the inner integrators return code after each solve of the subproblems. Does this explain it?

Probably not the best architecture, but at least it is functional.

PR #936, opened 2025-02-18
Showing initialization information for `AbstractDEProblem`

vyudu:
Towards https://github.com/SciML/ModelingToolkit.jl/issues/3268

### DIFF
--- a/src/problems/problem_utils.jl
+++ b/src/problems/problem_utils.jl
@@ -26,9 +26,13 @@ function Base.summary(io::IO, prob::AbstractDEProblem)
         prob.tspan isa Function ?
         "Unknown" : (prob.tspan === nothing ?
          "Nothing" : typeof(prob.tspan[1])),
-        no_color, ". In-place: ",
-        type_color, isinplace(prob),
-        no_color)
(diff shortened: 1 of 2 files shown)


src/problems/problem_utils.jl
```diff
@@ -26,9 +26,13 @@ function Base.summary(io::IO, prob::AbstractDEProblem)
         prob.tspan isa Function ?
         "Unknown" : (prob.tspan === nothing ?
          "Nothing" : typeof(prob.tspan[1])),
-        no_color, ". In-place: ",
-        type_color, isinplace(prob),
-        no_color)
+        no_color, 
+        ". In-place: ", type_color, isinplace(prob), no_color) 
+    println(io)
+    print(io, "Initialization status: ", type_color, initialization_status(prob), no_color)
```
ChrisRackauckas, line 32: O think we only need to print this if the initialization problem is not nothing

vyudu: Done, any other info I should add?

ChrisRackauckas (approved):

PR #943, opened 2025-02-26
fix: fix remake autodiff tests and Zygote adjoint

AayushSabharwal:


AayushSabharwal:
This should fix our Downstream, DiffEqBase and StochasticDiffEq CI. MTKStdlib needs some fixes which I'll get to, NonlinearSolve has "type piracy" between its own sub packages which is arguably fine but it would still be nice to make CI green. SciMLSensitivity has... problems. I've added Core6 and Core7 to CI here. I think Core6 should pass now, the type instability in `late_binding_update_u0_p` was fixed.

AayushSabharwal:
Okay Downstream testset failure is because of hashconsing. Specifically, `WeakValueDict` has a `lock` call which Zygote can't AD through.

AayushSabharwal:
On the surface though, this is due to calling `getproperty(::MTK.AbstractSystem, ::Symbol)`. We _could_ just define a zero adjoint for that?

ChrisRackauckas:
Yes it's non-numeric and can be skipped as a zerograd.

ext/SciMLBaseZygoteExt.jl
```diff
@@ -48,6 +48,9 @@ end
             grz = pullback(getter, sym, VA.u[j], VA.prob.p, VA.t[j])[2](Δ)
             du = [k == j ? grz[2] : zero(VA.u[1]) for k in 1:length(VA.u)]
             dp = grz[3] # pullback for p
+            if dp === nothing
+                dp = parameter_values(VA)
```
ChrisRackauckas, line 52: shouldn't this be zero?

AayushSabharwal:
I have the fixes ready locally. Needs https://github.com/SciML/ModelingToolkit.jl/pull/3422.

ext/SciMLBaseZygoteExt.jl
AayushSabharwal: Fixed

ext/SciMLBaseChainRulesCoreExt.jl
```diff
@@ -15,52 +16,28 @@ function ChainRulesCore.rrule(
         j::Integer)
     function ODESolution_getindex_pullback(Δ)
         i = symbolic_type(sym) != NotSymbolic() ? variable_index(VA, sym) : sym
-        if i === nothing
+        du, dprob = if i === nothing
             getter = getobserved(VA)
             grz = rrule_via_ad(config, getter, sym, VA.u[j], VA.prob.p, VA.t[j])[2](Δ)
-            du = [k == j ? grz[2] : zero(VA.u[1]) for k in 1:length(VA.u)]
-            dp = grz[3] # pullback for p
+            du = [k == j ? grz[3] : zero(VA.u[1]) for k in 1:length(VA.u)]
```
ChrisRackauckas, line 22: 😅 oh no.

ext/SciMLBaseZygoteExt.jl
```diff
@@ -40,32 +40,36 @@ import SciMLStructures
...
-    function ODESolution_getindex_pullback(Δ)
-        i = symbolic_type(sym) != NotSymbolic() ? variable_index(VA, sym) : sym
-        du, dprob = if i === nothing
-            getter = getobserved(VA)
-            grz = pullback(getter, sym, VA.u[j], VA.prob.p, VA.t[j])[2](Δ)
-            du = [k == j ? grz[2] : zero(VA.u[1]) for k in 1:length(VA.u)]
-            dp = grz[3] # pullback for p
-            dprob = remake(VA.prob, p = dp)
-            du, dprob
-        else
```
ChrisRackauckas, line 53: Is this because the ChainRulesCore one is fixed? If so, just delete instead of comment.

src/solutions/save_idxs.jl
```diff
@@ -105,6 +105,20 @@ struct SavedSubsystem{V, T, M, I, P, Q, C}
     partition_count::C
 end
 
+SavedSubsystem(indp, pobj, ::Nothing) = nothing
+
+function SavedSubsystem(indp, pobj, saved_idxs::Vector{Int})
+    isempty(saved_idxs) && return nothing
+    isempty(variable_symbols(indp)) && return nothing
```
ChrisRackauckas, line 112: won't this cause inference issues?

src/solutions/save_idxs.jl
```diff
@@ -357,29 +371,32 @@ corresponding to the state variables and a `SavedSubsystem` to pass to `build_so
...
+function get_save_idxs_and_saved_subsystem(prob, save_idxs::Vector{Int})
+    save_idxs, SavedSubsystem(prob, parameter_values(prob), save_idxs)
+end
+function get_save_idxs_and_saved_subsystem(prob, save_idx::Int)
+    save_idx, SavedSubsystem(prob, parameter_values(prob), save_idx)
+end
 function get_save_idxs_and_saved_subsystem(prob, save_idxs)
-    if save_idxs === nothing
-        saved_subsystem = nothing
+    if !(save_idxs isa AbstractArray) || symbolic_type(save_idxs) != NotSymbolic()
```
ChrisRackauckas, line 382: Won't this be an unnecessary allocation in the scalar case? You only want to do this when symbolic scalar.

src/solutions/save_idxs.jl
```diff
@@ -357,29 +371,32 @@ corresponding to the state variables and a `SavedSubsystem` to pass to `build_so
...
-        if !(save_idxs isa AbstractArray) || symbolic_type(save_idxs) != NotSymbolic()
-            _save_idxs = [save_idxs]
+        _save_idxs = save_idxs
+    end
+    saved_subsystem = SavedSubsystem(prob, parameter_values(prob), _save_idxs)
+    if saved_subsystem !== nothing
+        _save_idxs = get_saved_state_idxs(saved_subsystem)
+        if isempty(_save_idxs)
+            # no states to save
+            save_idxs = Int[]
```
ChrisRackauckas, line 392: Just error early? This case is a bit odd. Saving nothing? Must be a user issue.

ext/SciMLBaseZygoteExt.jl
AayushSabharwal: Oops, yeah.

src/solutions/save_idxs.jl
AayushSabharwal: I think the `isempty(saved_idxs)` case can be removed. I'll try and see if I can figure out the second case from type information.

src/solutions/save_idxs.jl
AayushSabharwal: The user could still be saving discrete variables

PR #944, opened 2025-02-27
Fix: make isclock robust to non TimeDomain argument

pitx-perf:
Tentative to fix: https://github.com/SciML/SciMLBase.jl/issues/945

## Checklist


AayushSabharwal (approved):
Thanks for the quick fix. This looks good if tests pass. It seems the issue is because Moshi uses the function `Moshi.Data.isa_variant` to check if a given object is the specified ADT variant, but the function is not robust to arguments that _aren't_ a variant of the ADT.

PR #946, opened 2025-02-27
Support Moshi wildcards via dispatch

### LINKED ISSUE #945: isclock is not robust to non TimeDomain argument
**Describe the example**

From https://github.com/SciML/SciMLBase.jl/pull/922
isclock(nothing) throws an error while it returned false previously.

isclock(nothing)

```julia

```

**Error & Stacktrace ⚠️**

```julia

```

**Not Working Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia

```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia

```

  - Output of `versioninfo()`

```julia

```

**Working Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia

```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia

```

  - Output of `versioninfo()`

```julia

```

**Additional context**

visr:
Fixes #945
Reported upstream as https://github.com/Roger-luo/Moshi.jl/issues/43.

This should restore the old behavior. One possible exception is `is_discrete_time_domain`, which was defined as not being continuous. So that previously returned true for non-clocks, which seems undesirable to me. Though I'd be happy to preserve that behavior if it is expected anywhere.

### DIFF
--- a/src/clock.jl
+++ b/src/clock.jl
@@ -40,22 +40,28 @@ discrete-time systems that assume a fixed sample time, such as PID controllers a
 filters.
 """ SolverStepClock
 
-isclock(c) = isa(c, TimeDomain) && @match c begin
+isclock(c::TimeDomain) = @match c begin
     PeriodicClock() => true
     _ => false
 end
 
-issolverstepclock(c) = isa(c, TimeDomain) && @match c begin
+issolverstepclock(c::TimeDomain) = @match c begin
     SolverStepClock() => true
     _ => false
 end
 
-iscontinuous(c) = isa(c, TimeDomain) && @match c begin
+iscontinuous(c::TimeDomain) = @match c begin
     ContinuousClock() => true
     _ => false
 end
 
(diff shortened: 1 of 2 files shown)


visr:
Ha I missed #944, which already fixes #945. I just rebased this on top of that because it's still worth fixing this behavior on master, which is unrelated to the Moshi change:

```jl
julia> is_discrete_time_domain(:banana)
true
```

And it is probably a bit better to use dispatch here instead of `isa`.

AayushSabharwal:
While I agree it does make sense that `is_discrete_time_domain` return `false` for non-time-domain arguments, it has been the way it is for a long time. I'm concerned this may break things that somehow rely on the weird behavior. MTK CI isn't running right now - https://github.com/SciML/ModelingToolkit.jl/pull/3382 will fix that and once it passes I'll approve this.

AayushSabharwal (approved):

PR #965, opened 2025-03-22
Remove deprecated operator code

ChrisRackauckas:


test/diffeqoperator.jl
```diff
@@ -1,39 +0,0 @@
-using SciMLBase
-using LinearAlgebra
-
-@testset "DiffEqOperator" begin
```
oscardssmith, line 4: should these tests be updated to use the new operators rather than deleted?

ChrisRackauckas: No, SciMLOperators.jl already has tests and they are much more extensive than this, superset it already IIRC. So I am not worried about copying these because I believe the first version already did that.

mateuszbaran:
Uh... I completely missed the deprecation notice. What should I change `AbstractDiffEqOperator` in https://github.com/JuliaManifolds/ManifoldDiffEq.jl to?

Issue #967, opened 2025-03-23
Cannot precompile


PhyX-Meow:
I got a lot of error messages like this when `using DifferentialEquation`
```
┌ DiffEqBase
│  WARNING: Method definition islinear(Any) in module SciMLOperators at /home/wuli/.julia/packages/SciMLOperators/2UPBq/src/interface.jl:311 overwritten in module SciMLBase at /home/wuli/.julia/packages/SciMLBase/pypmI/src/operators/operators.jl:7.
│  ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
└
```
```
Julia Version 1.11.4
Commit 8561cc3d68d (2025-03-10 11:36 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 9700X 8-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, generic)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)
Environment:
  LD_LIBRARY_PATH = .:../lib:./lib
```

ChrisRackauckas:
This was fixed by https://github.com/SciML/SciMLBase.jl/pull/965 . It was the result of moving a dispatch to the proper library to fix a type piracy that Aqua could now detect. You should be able to update in like 10 minutes and the warnings should go away. Because it's a type piracy it's hard to do this in a nicer way, but hopefully the Aqua infrastructure should make this not possible to add in the future.

PR #968, opened 2025-03-23
Remove fallbacks from SciMLOperators

ErikQQY:
Fix: #967 

Since the fallbacks are moved to SciMLOperators in https://github.com/SciML/SciMLOperators.jl/pull/260, we can safely remove these lines in SciMLBase


ChrisRackauckas:
This was in https://github.com/SciML/SciMLBase.jl/pull/965.

PR #969, opened 2025-03-23
Fix Dynamical Remake by making _func_cache optional

ChrisRackauckas:


src/remake.jl
```diff
@@ -187,7 +187,7 @@ function remake(
 
             props = @delete props.f2
 
-            if !ismissing(forig)
+            if !ismissing(forig) && hasproperty(forg, :_func_cache)
```
ChrisRackauckas, line 190:
```suggestion
            if !ismissing(forig) && hasproperty(forig, :_func_cache)
```

Issue #962, opened 2025-03-20
Verbosity System Description


jClugstor:
The current way to specify verbosity for solvers is through the `Bool` keyword argument `verbose`, which just causes the solver to print out more information if it's `true`. We can have a lot more flexibility by adding an `AbstractVerbosity` abstract type and concrete subtypes for every SciMLProblem type. 

For example we'll have a `BVPVerbosity <: AbstractVerbosity` that will have toggle-able Bool fields for everything that we would want to print during a BVP solve. 

We can have constructors so that `BVPVerbosity(true)` has every field `true`, `BVPVerbosity(false)` has every field false, and `BVPVerbosity(nothing)` is the default which has some things true and other things false.

This will allow the user to have much greater control over what is printed during the solution of any problem.

jClugstor:
Related to https://github.com/SciML/BoundaryValueDiffEq.jl/pull/290#issuecomment-2721686975

ChrisRackauckas:
Some things to read, especially from R's global warning system, which I think has done the best for user-level toggling of messaging:

* https://stackoverflow.com/questions/11239428/how-to-change-warning-setting
* https://stat.ethz.ch/R-manual/R-devel/library/base/html/warning.html
* https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/warnings
* https://adv-r.hadley.nz/conditions.html

The idea is that we want to get something with similar-ish functionality, but without relying on globals, which is not thread safe and has performance issues, instead allowing this to be at a function level and uniform across SciML. So let's keep that goal in mind.

The basic kwargs on everything should be:

* `verbose: highest level toggle on all printing and warning defaults. Defaults to Verbosity.Default, Allowed: None, Default, All`
* `error_control: toggles handling on error control and adaptivity checking algorithms, such as convergence issues and correctness guarantees (e.x. `dt < dtmin`). Defaults to Verbosity.Warn. Allowed Verbosity: None, Warn, Error.`
* `performance: toggles handling on performance issues, such as mismatched input/output types in out of place functions and other known performance traps. Defaults to Verbosity.Warn. Allowed Verbosity: None, Warn, Error.`
* `numerical: toggles handling of potential numerical issues, such as large condition numbers and detection of other potential floating point issues. Defaults to Verbosity.None. Allowed: None, Edge, All`.

Edge = only prints the info on edge cases, like high condition numbers and changes to QR pivoting.
[...]

PR #973, opened 2025-03-27
Add verbosity specifier types 

jClugstor:
## Checklist

## Additional context

Provides the types and constructors needed for https://github.com/SciML/SciMLBase.jl/issues/962. 

Each type has a kwarg constructor, one that takes a single Bool and sets every field to that Bool, one that takes `nothing` and sets default values for every field, and the default constructor is available, so that `@set` etc. works. 

These will be useful when the types have more fields and more toggles for what can be printed during the solves.

### DIFF
--- a/docs/src/interfaces/Verbosity.md
+++ b/docs/src/interfaces/Verbosity.md
@@ -1,9 +1,57 @@
 # SciML Verbosity
-The SciML verbosity system is designed to make it easy for users to specify what messages are logged, and at what level they are logged at, during the solution process. 
+The SciML verbosity system is designed to make it easy for users to specify what messages are logged and at what level they are logged at during the solution process. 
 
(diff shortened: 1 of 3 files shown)


src/problems/bvp_problems.jl
```diff
@@ -474,3 +474,40 @@ struct BVPAliasSpecifier <: AbstractAliasSpecifier
         end
     end
 end
+
+@doc doc"""
+
+Determines what is printed when solving a `BVPProblem`.
```
ErikQQY, line 480:
```suggestion
Determines what is printed when solving a `BVProblem`.
```

src/problems/bvp_problems.jl
```diff
@@ -474,3 +474,40 @@ struct BVPAliasSpecifier <: AbstractAliasSpecifier
...
+### Constructors
+
+```julia
+BVPVerbosity(verbose::Bool): sets all fields to `verbose`
+BVPVerbosity(; verbose = nothing): kwarg constructor
+BVPVerbosity(::Nothing): uses default options
+```
+"""
+struct BVPVerbosity <: AbstractVerbositySpecifier
+    verbose::Any
```
ErikQQY, line 494: `verbose` is either a `Bool` or `Nothing` right, why use `Any` here?

PR #976, opened 2025-03-28
move downstream tests downstream

oscardssmith:
This moves all of the heavy SciML tests to downstream.


oscardssmith:
I'm also taking inspiration from https://github.com/SciML/LinearSolve.jl/pull/592 and moving Downgrade CI to 1.10 so it has a prayer of passing.

oscardssmith:
@ChrisRackauckas do you understand the LTS python segfault or the downstream test failures? They seem unrelated, but it's somewhat hard to say.

ChrisRackauckas:
Downstream is https://github.com/FluxML/Zygote.jl/pull/1564

PR #978, opened 2025-03-31
fix: propagate kwargs to remake_initialization_data

vyudu:


AayushSabharwal (changes requested):
This is breaking. The method in MTK doesn't have kwargs, so this will never dispatch to it.

vyudu:
Right yeah I change the MTK function here
https://github.com/SciML/ModelingToolkit.jl/pull/3516

AayushSabharwal:
It's still breaking, because this version of SciMLBase won't work with older versions of MTK

PR #995, opened 2025-04-16
refactor: allow new clocks to be defined in downstream packages

AayushSabharwal:


src/clock.jl
```diff
@@ -71,13 +74,51 @@ function first_clock_tick_time(c, t0)
...
 
-struct IndexedClock{I}
-    clock::TimeDomain
+# public
+"""
+    $(TYPEDEF)
+
+A struct representing the operation of indexing a clock to obtain a subset of the time
+points at which it ticked. The actual list of time points depends on the time interval
+for which the clock was running, and can be obtained via `canonicalize_indexed_clock`
```
baggepinnen, line 83:
```suggestion
points at which it ticked. The actual list of time points depends on the tick instances 
on which the clock was ticking, and can be obtained via `canonicalize_indexed_clock`
```

baggepinnen: a clock may not have a fixed interval

baggepinnen (approved):

PR #1019, opened 2025-05-12
fix `isinplace` inference and add inference tests

oscardssmith:
This also removes `NoMethodsError`, `TooManyArgumentsError`, and `TooFewArgumentsError` and fixes inference for some `*Function` constructors`

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -4339,7 +4339,8 @@ function BVPFunction{iip, specialize, twopoint}(f, bc;
         if iip_f
             jac = update_coefficients! #(J,u,p,t)
         else
-            jac = (u, p, t) -> update_coefficients!(deepcopy(jac_prototype), u, p, t)
+            jac_prototype_copy = deepcopy(jac_prototype)
+            jac = (u, p, t) -> update_coefficients!(jac_prototype_copy, u, p, t)
         end
     end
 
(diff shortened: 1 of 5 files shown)


src/scimlfunctions.jl
```diff
@@ -4199,7 +4199,7 @@ IntervalNonlinearFunction(f::IntervalNonlinearFunction; kwargs...) = f
 struct NoAD <: AbstractADType end
 
 (f::OptimizationFunction)(args...) = f.f(args...)
-OptimizationFunction(args...; kwargs...) = OptimizationFunction{true}(args...; kwargs...)
+OptimizationFunction(f, args...; kwargs...) = OptimizationFunction{isinplace(f, 3)}(f, args...; kwargs...)
```
ChrisRackauckas, line 4202: no, OptimizationFunction always has a scalar return so its iip is determined not by `f` but by the other functions such as the jac and cons

src/utils.jl
```diff
@@ -246,66 +101,14 @@ form is disabled and the 2-argument signature is ensured to be matched.
...
-            iip_dispatch
-        else
-            # Equivalent to, if oop_dispatch exists, treat as oop
-            # Otherwise, it's iip
-            !oop_dispatch
-        end
+        hasmethod(f, ntuple(_->Any, outofplace_param_number)) && return false
+        hasmethod(f, ntuple(_->Any, inplace_param_number)) && return true
     end
+    throw(FunctionArgumentsError(fname, f))
```
ChrisRackauckas, line 111: This error message loses a lot of information. Can we in the error path do a method check and throw the more informative error message?

src/scimlfunctions.jl
oscardssmith: fair enough. Will revert.

oscardssmith:
I just remembered the tricky part here. This query isn't quite right (see https://github.com/JuliaLang/julia/issues/56279 for details):

```
julia> f(a::Int, b) = true

julia> hasmethod(f, (Any, Any))
false # because !(Any<:Int)
```

src/utils.jl
oscardssmith: IMO the better method is to just put the expected `nargs` in the exception and then we can make the exception printing in charge of that logic. That way none of the complicated bits are in the main function.

PR #1022, opened 2025-05-12
fix function inference

oscardssmith:
pulled out the uncontroversial part from https://github.com/SciML/SciMLBase.jl/pull/1019

### DIFF
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -948,7 +948,7 @@ dt: the time step
 
 ```julia
 ImplicitDiscreteFunction{iip,specialize}(f;
-                                analytic = __has_analytic(f) ? f.analytic : nothing, 
+                                analytic = __has_analytic(f) ? f.analytic : nothing,
                                 resid_prototype = __has_resid_prototype(f) ? f.resid_prototype : nothing)
 ```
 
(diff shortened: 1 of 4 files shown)


src/scimlfunctions.jl
```diff
@@ -4199,7 +4199,10 @@ IntervalNonlinearFunction(f::IntervalNonlinearFunction; kwargs...) = f
 struct NoAD <: AbstractADType end
 
 (f::OptimizationFunction)(args...) = f.f(args...)
-OptimizationFunction(args...; kwargs...) = OptimizationFunction{true}(args...; kwargs...)
+function OptimizationFunction(f, args...; kwargs...)
+    isinplace(f, 2, outofplace_param_number=2)
```
ChrisRackauckas, line 4203: This isn't needed though?

src/scimlfunctions.jl
```diff
@@ -4251,8 +4254,9 @@ end
 (f::MultiObjectiveOptimizationFunction)(args...) = f.f(args...)
 
 # Convenience constructor
-function MultiObjectiveOptimizationFunction(args...; kwargs...)
-    MultiObjectiveOptimizationFunction{true}(args...; kwargs...)
+function MultiObjectiveOptimizationFunction(f, args...; kwargs...)
+    isinplace(f, 2, outofplace_param_number=2)
```
ChrisRackauckas, line 4258:
```suggestion
    isinplace(f, 3, outofplace_param_number=2)
```

src/scimlfunctions.jl
oscardssmith: The purpose of this is to ensure that the function has the correct number of arguments, which I believe we want to do.

PR #1023, opened 2025-05-15
refactor: rewrite `OverrideInit` to be more Zygote compatible

AayushSabharwal:
https://github.com/SciML/ModelingToolkit.jl/pull/3585#issuecomment-2883919162

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -295,13 +295,18 @@ function get_initial_values(prob, valp, f, alg::OverrideInit,
     end
 
     if initdata.initializeprobmap !== nothing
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
I wouldn't say it's specific to Zygote, but it's also general good practice right? In instances where you cannot determine provenance easily (say you overrode variables with differing types), then also we have to create these copies.

AayushSabharwal:
Do any other AD frameworks run into this issue? Because if not, it is by definition specific to Zygote regardless of whether this is a code smell or not.

I'm not trying to say Zygote is broken but it would be good to understand this better so we don't keep running into it.

PR #1025, opened 2025-05-16
Make `get_initial_values` differentiate

DhairyaLGandhi:
## Checklist

## Additional context
`nlsol` has 2 incompatible graphs that assign to the same variable. This way we can use a simple indirection to disambiguate between them. For some reason this causes the `Ref` with the extra field access in `initprob`, so using `initdata` instead.


AayushSabharwal (approved):
This fixes the Zygote AD issue

PR #1036, opened 2025-05-26
Accumulation for ODEProblem

DhairyaLGandhi:
## Checklist

## Additional context

Sister PR for https://github.com/FluxML/Zygote.jl/pull/1574 which would have much narrower dispatch for `getfield`.

### DIFF
--- a/ext/SciMLBaseZygoteExt.jl
+++ b/ext/SciMLBaseZygoteExt.jl
@@ -304,14 +304,8 @@ end
   val = getfield(x, f)
   function back(Δ)
     Zygote.accum_param(__context__, val, Δ) === nothing && return
-    if isimmutable(x)
-      dx = (; Zygote.nt_nothing(x)..., pair(Val(f), Δ, x)...)
-      (_project(x, dx), nothing)
-    else
-      dx = Zygote.grad_mut(__context__, x)
-      dx[] = (; dx[]..., pair(Val(f), Zygote.accum(getfield(dx[], f), Δ))...)
-      return (dx[],nothing)
-    end
+    dx = (; Zygote.nt_nothing(x)..., pair(Val(f), Δ, x)...)
+    (_project(x, dx), nothing)
   end
   Zygote.unwrap(val), back
 end


ext/SciMLBaseZygoteExt.jl
```diff
@@ -299,4 +300,23 @@ end
...
+    # error()
+    Zygote.accum_param(__context__, val, Δ) === nothing && return
+    if isimmutable(x)
+      error()
+      dx = (; Zygote.nt_nothing(x)..., pair(Val(f), Δ, x)...)
+      (_project(x, dx), nothing)
+    else
+      dx = Zygote.grad_mut(__context__, x)
+      dx[] = (; dx[]..., pair(Val(f), Zygote.accum(getfield(dx[], f), Δ))...)
+      return (dx,nothing)
```
AayushSabharwal, line 315: What's the point of this block if the first line is `error()`?

ext/SciMLBaseZygoteExt.jl
```diff
@@ -299,4 +300,23 @@ end
...
+    end
+  end
+  Zygote.unwrap(val), back
+end
+Zygote.accum(::Tuple{}, ::NamedTuple{}) = ()
```
AayushSabharwal, line 320: Isn't this type piracy?

src/initialization.jl
```diff
@@ -258,50 +258,58 @@ function get_initial_values(prob, valp, f, alg::OverrideInit,
...
-            SciMLBase.successful_retcode(nlsol) && normresid <= abstol
-        else
-            SciMLBase.successful_retcode(nlsol)
-        end
-    end
-
+    nlsol2 = prob.f.initialization_data.initializeprob
     if initdata.initializeprobmap !== nothing
-        u0 = initdata.initializeprobmap(choose_branch(nlsol))
+        u02 = initdata.initializeprobmap(nlsol2)
```
AayushSabharwal, line 265: This doesn't make sense - we solve initialization, but calculate the new `u0` from the unsolved problem?

AayushSabharwal (changes requested):

ext/SciMLBaseZygoteExt.jl
DhairyaLGandhi: Removed in [a03bb02](https://github.com/SciML/SciMLBase.jl/pull/1036/commits/a03bb02cc059c88aa27f04aafe6fa00893888585)

ext/SciMLBaseZygoteExt.jl
ChrisRackauckas: It is 😅

DhairyaLGandhi: https://github.com/FluxML/Zygote.jl/pull/1574 is definitely the safer bet.

Why are problem types `mutable`? Do they need to be?

PR #1055, opened 2025-06-11
Update observables + AD tests with MTK v10

DhairyaLGandhi:
## Checklist

## Additional context

There seem to be some variable map changes in latest MTK, so updating tests to reflect that.


test/downstream/observables_autodiff.jl
```diff
@@ -33,15 +33,15 @@ sol = solve(prob, Tsit5())
...
     du = [du_ for _ in sol.u]
     @test du == gs.u
 
     # Observable in a vector
     gs, = gradient(sol) do sol
         sum(sum.(sol[[sys.w, sys.x]]))
     end
-    du_ = [0.0, 1.0, 1.0, 2.0]
+    du_ = [1.0, 1.0, 2.0, 0.0]
     du = [du_ for _ in sol.u]
```
ChrisRackauckas, line 45:
```suggestion
    du = [du_ for _ in sol.u]
```

Just make this correct instead.

ChrisRackauckas:
```
    du_ = [1.0, 1.0, 1.0, 0.0]
    du = [du_ for _ in sol[[D(x), x, y, z]]]
```

the order of this is not based on the tunable order, it's based on the sort order of `u`. That's why it broke. I am saying if you fix the order by checking it symbolically, then you should change the `du_` being tested to the order of the symbolic index.

DhairyaLGandhi:
There seems to be a new failure in SciMLSensitivity/Core1 that I haven't seen earlier. But that is independent of this PR, since this only touches different files, from a different test set

DhairyaLGandhi:
@ChrisRackauckas `solution_interface.jl` and `modelingtoolkit_remake.jl` seem to have non AD related changes, so I was curious if there were any missing changes in this PR that you had in mind.

PR #1074, opened 2025-07-26
fix: fix type-instability of `get_root_indp`

AayushSabharwal:


test/runtests.jl
```diff
@@ -60,6 +60,9 @@ end
         @time @safetestset "Clocks" begin
             include("clock.jl")
         end
+        @time @safetestset "JET" begin
```
ChrisRackauckas, line 63: Needs to go into a QA session.

ChrisRackauckas (approved):

RomeoV:
Makes sense. Just noting that this currently stops trimming (or at least JET being happy), but I'm happy to just use this branch for now. If others are looking for an alternative dirty fix that also seems to work, I've had success by pirating `get_root_indp` like so:
```julia
import NonlinearSolve.SciMLBase.get_root_indp
NonlinearSolve.SciMLBase.get_root_indp(prob::NonlinearLeastSquaresProblem) = prob.p
```

AayushSabharwal:
I haven't found a way to test this in a way that actually reflects how it is used and works across Julia versions.

PR #1075, opened 2025-07-26
Update Downgrade.yml

ChrisRackauckas:


.github/workflows/Downgrade.yml
```diff
@@ -15,15 +15,20 @@ jobs:
...
+          version: ${{ matrix.julia-version }}
+      - uses: julia-actions/julia-downgrade-compat@v2
 #        if: ${{ matrix.version == '1.6' }}
         with:
           skip: Pkg,TOML
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+        with:
+          FORCE_LATEST_COMPATIBLE_VERSION: false
+          CHECK_BOUNDS: false
```
ChrisRackauckas, line 33:
```suggestion
```

.github/workflows/Downgrade.yml
```diff
@@ -15,15 +15,19 @@ jobs:
...
-      - uses: julia-actions/julia-downgrade-compat@v1
+          version: ${{ matrix.julia-version }}
+      - uses: julia-actions/julia-downgrade-compat@v2
 #        if: ${{ matrix.version == '1.6' }}
         with:
           skip: Pkg,TOML
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+        with:
+          FORCE_LATEST_COMPATIBLE_VERSION: false
```
ChrisRackauckas, line 32:
```suggestion
```

ChrisRackauckas:
@StefanKarpinski does the minimum resolver respect Julia version for the standard libraries?

StefanKarpinski:
It should

.github/workflows/Downgrade.yml
```diff
@@ -15,15 +15,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['min']
+        downgrade_mode: ['alldeps']
+        julia-version: ['1.10', '1']
```
ChrisRackauckas, line 19:
```suggestion
        julia-version: ['1.10']
```

ChrisRackauckas:
It doesn't seem to prevent LinearAlgebra v1.10 on Julia v1.11 🤷 , but if we only use the downgrade CI on LTS it seems like that's a fine solution. Thanks @StefanKarpinski looks like it's fully in play now! 30 PRs to make.

PR #1081, opened 2025-07-28
`nothing` is a perfectly valid nonlinear solver algorithm

ChrisRackauckas:
it's the default

### DIFF
--- a/test/downstream/initialization.jl
+++ b/test/downstream/initialization.jl
@@ -242,8 +242,7 @@ end
     integ = init(prob; initializealg = NoInit())
 
     @testset "Errors without `nlsolve_alg`" begin
-        @test_throws SciMLBase.OverrideInitMissingAlgorithm SciMLBase.get_initial_values(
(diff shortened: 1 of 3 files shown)


test/downstream/initialization.jl
```diff
@@ -242,8 +242,7 @@ end
     integ = init(prob; initializealg = NoInit())
 
     @testset "Errors without `nlsolve_alg`" begin
-        @test_throws SciMLBase.OverrideInitMissingAlgorithm SciMLBase.get_initial_values(
-            prob, integ, fn, SciMLBase.OverrideInit(), Val(false))
+        SciMLBase.get_initial_values(prob, integ, fn, SciMLBase.OverrideInit(), Val(false))
```
ChrisRackauckas, line 245:
```suggestion
        SciMLBase.get_initial_values(prob, integ, fn, SciMLBase.OverrideInit(), Val(false);  abstol, reltol)
```

PR #1097, opened 2025-08-08
add PreallocationTools.DiffCache as cache for SplitODEProblem

thomvet:
Addresses https://github.com/SciML/OrdinaryDiffEq.jl/issues/2719
Needs: https://github.com/SciML/PreallocationTools.jl/pull/131 and https://github.com/SciML/DiffEqBase.jl/pull/1188


ChrisRackauckas:
If we do the dependency decrease, we might as well just make it be a DualCache?

ChrisRackauckas (approved):

src/scimlfunctions.jl
```diff
@@ -2618,9 +2618,10 @@ end
 
 (f::SplitFunction)(u, p, t) = f.f1(u, p, t) + f.f2(u, p, t)
 function (f::SplitFunction)(du, u, p, t)
-    f.f1(f._func_cache, u, p, t)
+    tmp = PreallocationTools.get_tmp(f._func_cache, u)
```
ChrisRackauckas, line 2621:
```suggestion
    tmp = get_tmp(f._func_cache, u)
```

src/scimlfunctions.jl
```diff
@@ -2666,11 +2667,11 @@ end
 
 (f::SDDEFunction)(args...) = f.f(args...)
 (f::SplitSDEFunction)(u, p, t) = f.f1(u, p, t) + f.f2(u, p, t)
-
 function (f::SplitSDEFunction)(du, u, p, t)
-    f.f1(f._func_cache, u, p, t)
+    tmp = PreallocationTools.get_tmp(f._func_cache)
```
ChrisRackauckas, line 2671:
```suggestion
    tmp = get_tmp(f._func_cache)
```

PR #1098, opened 2025-08-11
Move general utility functions and errors from DiffEqBase to SciMLBase 

jClugstor:
## Checklist

## Additional context

We want to move the solve dispatches for Nonlinear problems to NonlinearSolveBase from DiffEqBase. In order to do that we need to move some of the things from DiffEqBase to SciMLBase so that NonlinearSolveBase can use them.


jClugstor:
@ChrisRackauckas this moves the general utility functions used for the solve dispatches to SciMLBase, the extensions needed for those, and most of the errors that were in DiffEqBase before. 

I have a branch of DiffEqBase that imports and uses these from SciMLBase, which also means that for example `DiffEqBase.anyeltypedual` will still work even though it's defined in SciMLBase. I think that should make this non-breaking and allow us to gradually move to using SciMLBase instead of DiffEqBase for these.

Project.toml
```diff
@@ -29,19 +29,27 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 RuntimeGeneratedFunctions = "7e49a35a-f44a-4d26-94aa-eba1b4ca6b47"
 SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
 SciMLStructures = "53ae85a6-f571-4167-b2af-e1d143709226"
+Static = "aedffcd0-7271-4cad-89d0-dc628f76c6d3"
```
ChrisRackauckas, line 32: oh shit, that shouldn't be there... why is static needed? Just noticed this too late. This should not be in SciMLBase!

jClugstor: `reduce_tup` is used for the Dual type checking, `anyeltypedual`.

PR #1110, opened 2025-08-15
Put utilities for solve in SciMLBase

jClugstor:
## Checklist

## Additional context

This time for real and without Static


ext/SciMLBaseForwardDiffExt.jl
```diff
@@ -0,0 +1,410 @@
...
+But given the properties of automatic differentiation requiring that differentiation of parameters
+implies differentiation of state, we assume any dual parameters implies differentiation of state
+and then attempt to upconvert `u0` to match that dual-ness. Because this changes types, this needs
+to be specified at compiled time and thus cannot have a Bool-based opt out, so in the future this
+may be extended to use a preference system to opt-out with a `UPCONVERT_DUALS`. In the case where
+upconversion is not done automatically, the user is required to upconvert all initial conditions
+themselves, for an example of how this can be confusing to a user see
+https://discourse.julialang.org/t/typeerror-in-julia-turing-when-sampling-for-a-forced-differential-equation/82937
+"""
+@generated function anyeltypedual(x, ::Type{Val{counter}}) where {counter}
```
ChrisRackauckas, line 103: Why is this needed in SciMLBase?

jClugstor: `anyeltypedual` is used in `promote_u0`, which is used by `get_concrete_problem` for NonlinearProblem and also differential equations. So it's needed for both. So this either needs to be in SciMLBase, or it needs to be in both DiffEqBase and NonlinearSolveBase.

ChrisRackauckas: why would nonlinearproblem need it? It should use implicit diff, and it requires p as an array for that just like linearproblem?

PR #1117, opened 2025-08-25
Move solve for EnsembleProblem and `_concrete_solve_adjoint` to SciMLBase

jClugstor:
## Checklist

## Additional context
Currently `solve` for Ensemble and WeightedEnsemble problems is in DiffEqBase. However there can be ensembles of NoninearProblems as well as DiffEqProblems. So this is part of the process of removing the DiffEqBase dependency from NonlinearSolve.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -516,3 +516,40 @@ set_mooncakeoriginator_if_mooncake(x::SciMLBase.ADOriginator) = x
     end
     q
 end
+
+####
+# Catch undefined AD overload cases
+
+const ADJOINT_NOT_FOUND_MESSAGE = """
+                                  Compatibility with reverse-mode automatic differentiation requires SciMLSensitivity.jl.
+                                  Please install SciMLSensitivity.jl and do `using SciMLSensitivity`/`import SciMLSensitivity`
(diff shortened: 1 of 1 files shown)


jClugstor:
@ChrisRackauckas I think this needs to go in, then this https://github.com/SciML/DiffEqBase.jl/pull/1201 to remove it from DiffEqBase. Then I think both https://github.com/SciML/DiffEqBase.jl/pull/1197 and https://github.com/SciML/NonlinearSolve.jl/pull/669 will be very close to finished.

jClugstor:
@ChrisRackauckas for some reason when `solve(prob::EnsembleProblem...` is defined in both SciMLBase and DiffEqBase it causes some Dual type check tests to error, which is what the DiffEqBase core test failures are from. Maybe something to do with the method being overridden? 

Anyway, the PR to remove these from DiffEqBase is here https://github.com/SciML/DiffEqBase.jl/pull/1201, but the tests for both of these won't pass without the other.

jClugstor:
@ChrisRackauckas should I split this PR up or is this ok?

ChrisRackauckas:
Downstream isn't setup with `!isdefined` yet?

Issue #1207, opened 2026-01-15
Mooncake does not support SymbolicIndexingInterface AD


ChrisRackauckas-Claude:
## Summary

When using `DifferentiationInterface.gradient` with `AutoMooncake()` backend on functions that involve `SymbolicIndexingInterface` operations (like `sol[sys.x]` symbolic indexing), the tests fail.

## Context

This was discovered during the migration from direct AD backend usage to DifferentiationInterface in PR #1206. The following test patterns fail with Mooncake:

- Symbolic indexing gradients: `sol -> sum(sol[lorenz1.x])`
- Vector symbolic indexing: `sol -> sum(sum.(sol[[lorenz1.x, lorenz2.x]]))`
- Tuple symbolic indexing: `sol -> sum(sum.(collect.(sol[(lorenz1.x, lorenz2.x)])))`
- Time series symbolic indexing: `sol -> sum(sum.(sol[[lorenz1.x, lorenz2.x], :]))`
- Observable function AD: `sol -> sum(sol[sys.w])`
- BatchedInterface AD

## Affected Files

- `test/downstream/adjoints.jl`
- `test/downstream/observables_autodiff.jl`

## Current Status

Tests are marked as `@test_broken` in PR #1206 until this is resolved.

## Related

This may be related to how Mooncake handles the ChainRules/rrules defined in SymbolicIndexingInterface or SciMLBase's extensions.

ChrisRackauckas-Claude:
## Investigation Results

After attempting to port the ChainRulesCore/Zygote rules to Mooncake, I've found that this is a more fundamental issue than simply missing rules.

### The Problem

Mooncake is a **source-to-source AD** system, which means it needs to compile derivative code for the entire function being differentiated. When differentiating through `sol[x]` where `sol` is an `ODESolution`:

1. Mooncake tries to compile rules for the entire `ODESolution` type hierarchy
2. `ODESolution` contains deeply nested generic types (MTKParameters, ODEFunction, ModelingToolkit wrappers, etc.)
3. This causes a **stack overflow during rule compilation**, not during execution

### Attempted Solutions

1. **Adding `@is_primitive` declarations** - Marking `getindex(ODESolution, sym)` as primitive doesn't help because Mooncake still needs to understand how to handle the `ODESolution` tangent type
2. **Custom `rrule!!` implementations** - The rules were written but Mooncake fails during compilation before ever reaching them
[...]

PR #1212, opened 2026-01-16
Add @from_rrule for ODESolution symbolic indexing in Mooncake extension

ChrisRackauckas-Claude:
## Summary

- Adds `@from_rrule` macro to import the ChainRulesCore rrule for `getindex(::ODESolution, sym)` to Mooncake
- Enables `sol[sym]` patterns to work with Mooncake's reverse-mode AD by reusing the existing ChainRulesCore implementation

## Motivation

This is a first step toward addressing #1207 (Mooncake AD backend incompatibility with SymbolicIndexingInterface).

The existing ChainRulesCore extension already has a proper rrule for symbolic indexing. By using Mooncake's `@from_rrule` macro, we can import this rule for use with Mooncake without duplicating the implementation.

## Limitations

- This only handles `sol[sym]` (2-argument getindex). The 3-argument version `sol[sym, j::Integer]` uses `RuleConfig` for AD nesting via `rrule_via_ad`, which may need additional work.
- The deeper issue of Mooncake needing custom tangent types for the complex `ODESolution` type hierarchy may still require additional work.

## Test plan

- [ ] Verify extension loads without error: `using SciMLBase, Mooncake`
- [ ] Run existing Mooncake tests to check for regressions
- [ ] Test symbolic indexing gradients with Mooncake (currently marked as `@test_broken` in `test/downstream/adjoints.jl`)

### DIFF
--- a/ext/SciMLBaseMooncakeExt.jl
+++ b/ext/SciMLBaseMooncakeExt.jl
@@ -1,7 +1,9 @@
 module SciMLBaseMooncakeExt
 
 using SciMLBase, Mooncake
-using SciMLBase: ADOriginator, ChainRulesOriginator, MooncakeOriginator
+using SciMLBase: ADOriginator, ChainRulesOriginator, MooncakeOriginator,
+    ODESolution, RODESolution, ODEProblem, SDEProblem, EnsembleSolution,
+    NonlinearProblem, IntervalNonlinearProblem
 import Mooncake: rrule!!, CoDual, zero_fcodual, @is_primitive,
     @from_rrule, @zero_adjoint, @mooncake_overlay, MinimalCtx,
     NoPullback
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
## Testing Results

I tested whether this PR fixes issue #1207 by running the Zygote tests with Mooncake on Julia 1.11.8.

### What the PR Does
The PR adds a single rule in `ext/SciMLBaseMooncakeExt.jl`:
```julia
@from_rrule(
    MinimalCtx,
    Tuple{typeof(getindex), ODESolution, Any},
    true,
)
```
This imports the existing ChainRulesCore rrule for `sol[sym]` into Mooncake.

### Test Results (Julia 1.11.8)

| Test Case | Zygote | Mooncake |
|-----------|--------|----------|
| Simple ODE (no MTK): `sol[1]` | ✓ | ✓ |
| Simple ODE (no MTK): `sol.u` | ✓ | ✓ |
| MTK ODE: `sol[lorenz1.x]` | ✓ | ✗ (StackOverflow) |
| MTK ODE: `sol[[lorenz1.x, lorenz2.x]]` | ✓ | ✗ (StackOverflow) |

### Root Cause
The ChainRulesCore rrule for `getindex(::ODESolution, sym)` **works correctly** and can be called directly. The problem is:

1. Mooncake's source-to-source AD requires compiling tangent types for `ODESolution`
2. ModelingToolkit-produced `ODESolution` objects contain deeply nested types (MTKParameters, RuntimeGeneratedFunctions, GeneratedFunctionWrapper, etc.)
3. Compiling tangent types for these deeply nested structures causes **StackOverflowError**

The `@from_rrule` only covers the `getindex` operation itself - it doesn't prevent Mooncake from trying to compile rules for the ODESolution's tangent type, which triggers the overflow.

### Conclusion
This PR does **not** fix issue #1207 for ModelingToolkit-based ODEs. The tests marked as `@test_broken` for Mooncake in `test/downstream/adjoints.jl` and `test/downstream/observables_autodiff.jl` still fail with StackOverflowError.

As noted in issue #1207, full support would require:
- Tangent type definitions for `ODESolution` and nested MTK types
- Struct construction/access rules for gradient propagation  
- Potentially upstream Mooncake changes to handle deeply nested types

PR #1237, opened 2026-02-19
fix: remove ODEProblem constructor rrules causing Zygote Jacobian accumulation

ChrisRackauckas-Claude:
## Summary

- Remove the two `ChainRulesCore.rrule` definitions for `ODEProblem` and `ODEProblem{iip, T}` constructors in `SciMLBaseChainRulesCoreExt.jl`
- These rrules trigger a known Zygote bug with mutable struct tangent caching, causing `Zygote.jacobian` (and any repeated pullback calls) to accumulate gradients cumulatively instead of computing independent gradients

## Root Cause

`ODEProblem` is a `mutable struct`. Zygote stores mutable struct tangents in a `Context.cache::IdDict` keyed by object identity. When a field is accessed via `getfield`, Zygote accumulates into a shared `Ref{Any}` in this cache. Zygote's native constructor pullback (`Jnew` in `src/lib/lib.jl:321`) resets this cache after each backward pass call. However, when a ChainRulesCore constructor rrule exists, `Jnew` is bypassed and the cache is never reset. On the second `pb()` call, the old tangent from the first call persists and gets summed with the new one.

This is acknowledged in Zygote source at `src/lib/lib.jl:313`: `# TODO captured mutables + multiple calls to 'back'`

## Minimal Reproducer (pure Zygote, no ODE)

```julia
using Zygote, ChainRulesCore
mutable struct S; x::Vector{Float64}; end
ChainRulesCore.rrule(::Type{S}, x) = S(x), ȳ -> (NoTangent(), ȳ.x)
f(x) = S(x).x .^ 2

Zygote.jacobian(f, [1.0, 2.0])
# Returns [2 0; 2 4] instead of [2 0; 0 4]
```

## ODE Reproducer (before this fix)

```julia
using Zygote, OrdinaryDiffEq, SciMLSensitivity
f(x) = solve(ODEProblem((u,p,t)->p.*u, x, (0.,1.), [1.,2.]),
             Tsit5(), abstol=1e-6, reltol=1e-3).u[end]
Zygote.jacobian(f, [1.0, 2.0])
# Returns [e 0; e e²] instead of [e 0; 0 e²]
```

This was reported at https://discourse.julialang.org/t/automatic-differentiation-of-the-flow-of-an-ode-differentiationinterface-scimlsensitivity/135730

## Fix

Remove the constructor rrules. These rrules only did trivial pass-through of tangent fields (`ȳ.f, ȳ.u0, ȳ.tspan, ...`), which is exactly what Zygote's native constructor tracing does via `__new__`/`__splatnew__`. Without them, Zygote uses `Jnew` which properly resets the mutable tangent cache between pullback calls.

## Test plan

- [x] `Zygote.jacobian` gives correct diagonal Jacobian for ODE flow
- [x] `DifferentiationInterface.jacobian` with `AutoZygote()` matches `ForwardDiff`
- [x] Individual `Zygote.gradient` calls remain correct
- [x] Repeated `pullback()` calls give independent (non-accumulated) results
- [x] 3-element system shows correct diagonal Jacobian (no lower-triangular pattern)
- [x] `Pkg.test("SciMLBase")` passes (all 4413 tests)

### DIFF
--- a/ext/SciMLBaseChainRulesCoreExt.jl
+++ b/ext/SciMLBaseChainRulesCoreExt.jl
@@ -65,25 +65,12 @@ function ChainRulesCore.rrule(::typeof(getindex), VA::ODESolution, sym)
     return VA[sym], ODESolution_getindex_pullback
 end
 
-function ChainRulesCore.rrule(::Type{ODEProblem}, args...; kwargs...)
-    function ODEProblemAdjoint(ȳ)
-        return (NoTangent(), ȳ.f, ȳ.u0, ȳ.tspan, ȳ.p, ȳ.kwargs, ȳ.problem_type)
-    end
-
-    return ODEProblem(args...; kwargs...), ODEProblemAdjoint
-end
-
-function ChainRulesCore.rrule(
-        ::Type{
-            <:ODEProblem{iip, T},
-        }, args...; kwargs...
-    ) where {iip, T}
-    function ODEProblemAdjoint(ȳ)
-        return (NoTangent(), ȳ.f, ȳ.u0, ȳ.tspan, ȳ.p, ȳ.kwargs, ȳ.problem_type)
-    end
-
-    return ODEProblem(args...; kwargs...), ODEProblemAdjoint
-end
+# NOTE: Constructor rrules for ODEProblem were removed. ODEProblem is a mutable struct,
+# and Zygote's mutable struct tangent cache is not reset between repeated pullback calls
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Yeah mutable structs capture a ref in the pullback when using `rrule`s. The constructor adjoint needs to account for that. It shouldn't affect subsequent `gradient` calls, only when the pullback is reused.

ChrisRackauckas:
Can you follow up with what would be faster here then?

DhairyaLGandhi:
Sure

PR #1250, opened 2026-02-25
Add supports_solve_rng trait for ensemble RNG forwarding guard

isaacsas:


isaacsas:
@ChrisRackauckas can we merge this?

PR #1261, opened 2026-03-01
Unify ODE solution types: DAESolution, RODESolution as ODESolution aliases

ChrisRackauckas-Claude:
## Summary

- Expands `ODESolution` to carry the union of all fields from the previously separate types: `du` (derivatives, from DAESolution), `W` (noise process, from RODESolution), and `seed` (random seed, from RODESolution)
- `DAESolution` and `RODESolution` are now `const` aliases for `ODESolution`, eliminating redundant struct definitions
- All `build_solution` dispatches create `ODESolution` directly, with problem-type dispatch determining which fields are populated
- `calculate_solution_errors!` is unified into a single implementation that branches on problem type (`AbstractDAEProblem`, `AbstractRODEProblem`/`AbstractSDDEProblem`, `AbstractDDEProblem`, or standard ODE)
- Removes ~290 lines of duplicated code across `dae_solutions.jl` and `rode_solutions.jl` (duplicate `getproperty`, `solution_new_retcode`, `solution_new_tslocation`, `solution_slice`, `ConstructionBase` methods, callable methods, etc.)
- `isdenseplot` now checks `sol.prob` type instead of `sol isa AbstractRODESolution`

## Breaking changes

- `ODESolution` struct has new fields (`du`, `W`, `seed`) which changes its type parameters
- Code dispatching on the concrete types `DAESolution` or `RODESolution` will now match any `ODESolution` (since they are aliases)
- Code dispatching on `AbstractDAESolution` or `AbstractRODESolution` with concrete solution values will no longer match (since the concrete type is always `ODESolution <: AbstractODESolution`)
- The abstract types `AbstractDAESolution`, `AbstractDDESolution`, `AbstractRODESolution` are unchanged and remain in the hierarchy for interface purposes

## Test plan

- [x] All SciMLBase unit tests pass
- [ ] Downstream solver packages (OrdinaryDiffEq, StochasticDiffEq, DASKR, Sundials) need testing for compatibility

### DIFF
--- a/src/solutions/dae_solutions.jl
+++ b/src/solutions/dae_solutions.jl
@@ -1,89 +1,8 @@
 """
-$(TYPEDEF)
-
-Representation of the solution to an differential-algebraic equation defined by an DAEProblem.
-
-## DESolution Interface
-
-For more information on interacting with `DESolution` types, check out the Solution Handling
-page of the DifferentialEquations.jl documentation.
-
-<https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/>
-
-## Fields
-
-  - `u`: the representation of the DAE solution. Given as an array of solutions, where `u[i]`
-    corresponds to the solution at time `t[i]`. It is recommended in most cases one does not
-    access `sol.u` directly and instead use the array interface described in the Solution
-    Handling page of the DifferentialEquations.jl documentation.
-  - `du`: the representation of the derivatives of the DAE solution.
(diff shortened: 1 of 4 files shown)


oscardssmith:
Is the improvement just that there would be less duplication?

isaacsas:
Seems reasonable to me, but I don't use DAESolution or RODESolution, so don't have a great idea of how disruptive this would be.

PR #1287, opened 2026-03-25
Gate DomainError hint to only trigger within SciML solver context

### LINKED ISSUE #606: Importing `SciMLBase` has side-effects on `Base.Math.throw_complex_domainerror`
Open a fresh julia session in a clean (temporary) environment. Then:

```julia
log(-1)
```

```
ERROR: DomainError with -1.0:
log was called with a negative real argument but will only return a complex result if called with a complex argument. Try log(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math ./math.jl:33
 [2] _log(x::Float64, base::Val{:ℯ}, func::Symbol)
   @ Base.Math ./special/log.jl:301
 [3] log
   @ Base.Math ./special/log.jl:267 [inlined]
 [4] log(x::Int64)
   @ Base.Math ./math.jl:1578
 [5] top-level scope
   @ REPL[1]:1
```

Then,

```julia
using Pkg
Pkg.add("SciMLBase")
using SciMLBase

log(-1)
```

```
ERROR: DomainError with -1.0:
log was called with a negative real argument but will only return a complex result if called with a complex argument. Try log(Complex(x)).
DomainError detected in the user `f` function. This occurs when the domain of a function is violated.
For example, `log(-1.0)` is undefined because `log` of a real number is defined to only output real
numbers, but `log` of a negative number is complex valued and therefore Julia throws a DomainError
by default. Cases to be aware of include:

* `log(x)`, `sqrt(x)`, `cbrt(x)`, etc. where `x<0`
* `x^y` for `x<0` floating point `y` (example: `(-1.0)^(1/2) == im`)

Within the context of SciML, this error can occur within the solver process even if the domain constraint

<snip>

Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math ./math.jl:33
 [2] _log(x::Float64, base::Val{:ℯ}, func::Symbol)
   @ Base.Math ./special/log.jl:301
 [3] log
   @ Base.Math ./special/log.jl:267 [inlined]
 [4] log(x::Int64)
   @ Base.Math ./math.jl:1578
 [5] top-level scope
   @ REPL[13]:1
```

Because `log` is being called outside of the context of any SciML package, I would expect the normal error message to be thrown. 

One thing that confuses me a bit is that, by behaviour, the `Base.Math.throw_complex_domainerror` function appears to have been piratically redefined:

```julia
Base.Math.throw_complex_domainerror(:moo, 1)
```
```
ERROR: DomainError with 1:
moo was called with a negative real argument but will only return a complex result if called with a complex argument. Try moo(Complex(x)).
DomainError detected in the user `f` function. This occurs when the domain of a function is violated.
For example, `log(-1.0)` is undefined because `log` of a real number is defined to only output real
numbers, but `log` of a negative number is complex valued and therefore Julia throws a DomainError
by default. Cases to be aware of include:

<snip>
```

... and yet Julia only claims that one method exists, with the standard implementation:
```julia
julia> methods(Base.Math.throw_complex_domainerror)
# 1 method for generic function "throw_complex_domainerror" from Base.Math:
 [1] throw_complex_domainerror(f::Symbol, x)
     @ math.jl:32
```

Even using e.g. `@code_lowered Base.Math.throw_complex_domainerror(:moo, 1)` shows the standard implementation, not what actually gets called.

Observed in Julia 1.10, and `SciMLBase` 2.21.0.

**Additional context**

This error caused a little confusion for me, as I assumed that it was coming from within a SciML function call -- in fact it was just a call to `log` in some other code I had written. But I happened to have an indirect dependency on SciMLBase, causing the above to occur.

ChrisRackauckas-Claude:
## Summary

- Adds `_has_sciml_in_stacktrace()` helper that inspects the exception backtrace to check if SciMLBase code is present (e.g. `SciMLFunction` wrappers in `scimlfunctions.jl`)
- The DomainError error hint now only displays the SciML-specific message when the error originates from within a SciML solver context
- Standalone calls like `log(-1)` at the REPL or `logit(2)` via LogExpFunctions no longer show the misleading SciML hint

Fixes #606

## How it works

The error hint handler calls `catch_backtrace()` to get the current exception's stacktrace, then checks if any frame's file path starts with SciMLBase's package directory (via `pkgdir`). When a solver calls user code, SciMLBase's `SciMLFunction` wrappers are always on the stack. When `log(-1)` is called standalone, they're not.

The other two hints (MethodError for NullParameters, FunctionWrappersWrappers error) don't need this gating since they already check for SciML-specific types.

## Test plan

- [x] `log(-1.0)` at REPL with SciMLBase loaded → no SciML hint
- [x] `logit(2)` via LogExpFunctions with SciMLBase loaded → no SciML hint
- [x] `log(u)` inside `solve(ODEProblem(...), Tsit5())` → SciML hint correctly shown
- [x] `Pkg.test()` passes

### DIFF
--- a/test/error_hint_tests.jl
+++ b/test/error_hint_tests.jl
@@ -1,18 +1,12 @@
 using SciMLBase, Test
 
-function showerror_string(e)
-    buf = IOBuffer()
-    showerror(buf, e)
-    return String(take!(buf))
-end
-
 @testset "DomainError hint only in SciML context" begin
     # DomainError outside any SciML context should NOT show the SciML hint
     @testset "standalone log(-1) does not trigger hint" begin
         output = try
             log(-1.0)
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("DomainError", output)
         @test !occursin("DomainError detected in the user", output)
@@ -22,7 +16,7 @@ end
         output = try
             sqrt(-1.0)
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("DomainError", output)
         @test !occursin("DomainError detected in the user", output)
@@ -35,7 +29,7 @@ end
         output = try
             odefun(-1.0, nothing, 0.0)
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("DomainError", output)
         @test occursin("DomainError detected in the user", output)
@@ -48,7 +42,7 @@ end
         output = try
             odefun(du, [-1.0], nothing, 0.0)
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("DomainError", output)
         @test occursin("DomainError detected in the user", output)
@@ -59,7 +53,7 @@ end
         output = try
             throw(DomainError(-1, "some other message"))
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("DomainError", output)
         @test !occursin("DomainError detected in the user", output)
@@ -71,7 +65,7 @@ end
         output = try
             SciMLBase.NullParameters() + 1
         catch e
-            showerror_string(e)
+            sprint(showerror, e)
         end
         @test occursin("NullParameters", output)
     end


test/error_hint_tests.jl
```diff
@@ -0,0 +1,78 @@
...
+        @test occursin("DomainError", output)
+        @test occursin("DomainError detected in the user", output)
+    end
+
+    # Non-complex DomainError should never show the hint regardless of context
+    @testset "unrelated DomainError does not trigger hint" begin
+        output = try
+            throw(DomainError(-1, "some other message"))
+        catch e
+            showerror_string(e)
```
devmotion:
```suggestion
            sprint(showerror, e)
```

PR #1311, opened 2026-04-10
refactor: accept arbitrary kwargs in some MTK-only API functions

AayushSabharwal:
## Checklist

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -428,16 +428,24 @@ function SciMLBase.remake(
 end
 
 """
-    remake_initialization_data(sys, scimlfn, u0, t0, p, newu0, newp)
+    $TYPEDEF
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
we generally want to go the other direction? Why are we loosening this? Is there another validation step?

AayushSabharwal:
Requirements tend to come up later that make changes very difficult. For example, Catalyst requested the ability to pass `warn_initialize_determined` to `remake`, which we can't do without this. All keyword arguments that will be passed will be documented here, but ensuring that implementors accept arbitrary kwargs makes it so that we can pass additional information without breaking everything.

ChrisRackauckas:
Can it be open ended with validation via some context struct?

AayushSabharwal:
That works too. I'll make the changes.

PR #1313, opened 2026-04-10
Add Mooncake+ForwardDiff extension for symbolic indexing AD

ChrisRackauckas-Claude:
## Summary

Adds Mooncake support for `getindex(sol, sym)` symbolic indexing on `AbstractTimeseriesSolution`, addressing part of #1207.

## Problem

Mooncake currently cannot differentiate through `sol[sys.x]` or `sol[sys.w]` because the SymbolicIndexingInterface dispatch chain uses:
- Hash-consed symbols (SymbolicUtils' `BasicSymbolic` interning)
- Task-local values (`COMPARE_FULL` in hashconsing)
- `IdDict{Any, Any}` for caching
- Atomic pointers (`atomic_pointerref` in hashcons)

None of which are differentiable infrastructure. Each layer requires a different fix and Mooncake's compiler hits walls trying to differentiate through atomic intrinsics, task storage, and hash-consing — which all lead to the same place: looking up a symbol's index.

## Fix

Add `rrule!!` primitives for `getindex(::AbstractTimeseriesSolution, sym)` and `getindex(::AbstractTimeseriesSolution, sym, j)` that bypass the entire dispatch chain:

- **State variables**: scatter the output gradient directly into `sol.u[k][i]`
- **Observables**: compute `∂obs/∂u` via ForwardDiff at each saved time point and accumulate into the u-fdata of the solution

This mirrors how `SciMLBaseChainRulesCoreExt` handles the same operation for Zygote (using `rrule_via_ad` through the observed function).

## Why a combined Mooncake+ForwardDiff extension

The observable case requires nested AD on the observed function. ForwardDiff is the simplest tool for this and is already a SciMLBase weakdep. Splitting into a combined extension follows the standard Julia pattern for multi-package dependencies.

## Verified

```julia
using DifferentiationInterface, ADTypes
using ModelingToolkit, OrdinaryDiffEq, Mooncake, ForwardDiff

@parameters σ ρ β
@variables x(t) y(t) z(t) w(t)
eqs = [D(D(x)) ~ σ*(y-x), D(y) ~ x*(ρ-z) - y, D(z) ~ x*y - β*z, w ~ x+y+z+2β]
@mtkcompile sys = System(eqs, t)
prob = ODEProblem(sys, [...], (0.0, 100.0))
sol = solve(prob, Tsit5())

# State variable
gs = DifferentiationInterface.gradient(sol -> sum(sol[sys.x]), AutoMooncake(), sol)
# gs.fields.u[k] = [0.0, 1.0, 0.0, 0.0]  ✓

# Observable
gs = DifferentiationInterface.gradient(sol -> sum(sol[sys.w]), AutoMooncake(), sol)
# gs.fields.u[k] = [1.0, 1.0, 1.0, 0.0]  ✓ (matches Zygote)
```

## Test plan
- [x] Verified state variable indexing returns correct gradient
- [x] Verified observable indexing matches expected `∂(x+y+z+2β)/∂(D(x),x,y,z) = (0,1,1,1)` per timestep
- [ ] CI passes
- [ ] Update `test/downstream/observables_autodiff.jl` to enable Mooncake tests (separate PR — Mooncake's `Tangent` uses `gs.fields.u` not `gs.u`)

## Limitations not yet handled
- Parameter gradients via observables (the rrule scatters into `sol.u` only)
- Initialization observable indexing (`isol[w]`)
- DAE observable indexing (likely works but untested)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -64,6 +64,7 @@ SciMLBaseMakieExt = "Makie"
 SciMLBaseMeasurementsExt = "Measurements"
 SciMLBaseMonteCarloMeasurementsExt = "MonteCarloMeasurements"
 SciMLBaseMooncakeExt = "Mooncake"
+SciMLBaseMooncakeForwardDiffExt = ["Mooncake", "ForwardDiff"]
 SciMLBasePartialFunctionsExt = "PartialFunctions"
 SciMLBasePyCallExt = "PyCall"
 SciMLBasePythonCallExt = "PythonCall"
--- a/ext/SciMLBaseMooncakeForwardDiffExt.jl
+++ b/ext/SciMLBaseMooncakeForwardDiffExt.jl
@@ -1,0 +1,85 @@
+module SciMLBaseMooncakeForwardDiffExt
+
+using SciMLBase
+using SciMLBase: ChainRulesOriginator, MooncakeOriginator, AbstractTimeseriesSolution,
+    getobserved
+using SymbolicIndexingInterface: symbolic_type, NotSymbolic, variable_index,
+    parameter_values
+import Mooncake
+using Mooncake: CoDual, zero_fcodual, @is_primitive, MinimalCtx,
+    NoPullback, NoFData, NoRData, fdata, zero_tangent
+using ForwardDiff: ForwardDiff
+
+# These rules override the state-only versions in SciMLBaseMooncakeExt with
+# full observable support via ForwardDiff. They are dispatched via the same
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Closing in favor of #NEW — that one uses Mooncake's own `build_rrule` to recurse into the observable function (mirroring how `SciMLBaseChainRulesCoreExt` uses `rrule_via_ad` for Zygote), so it doesn't need ForwardDiff or a combined extension.

PR #1319, opened 2026-04-12
Fix nested ForwardDiff tag mismatch in UJacobianWrapper for Rosenbrock solvers

AdityaPandeyCN:
## Checklist

## Additional context
Fix for https://github.com/SciML/OrdinaryDiffEq.jl/issues/3381
When NonlinearSolve computes a Jacobian over an ODE solve,` p `enters as D`ual{NLTag}` and Rosenbrock's internal ForwardDiff seeds `u `as `Dual{OrdEqTag, Dual{NLTag}, CS}`. UJacobianWrapper passes these mismatched types to the user function. p[1]*u[1] produces `Dual{NLTag} `outer due to tag precedence, but du expects `Dual{OrdEqTag}` outer, so the assignment crashes.
This promote p to match `eltype(u)` in UJacobianWrapper before calling f

### DIFF
--- a/ext/SciMLBaseForwardDiffExt.jl
+++ b/ext/SciMLBaseForwardDiffExt.jl
@@ -8,7 +8,7 @@ import SciMLBase:
     AbstractTimeseriesSolution, NonlinearProblem, NonlinearLeastSquaresProblem,
     ODEProblem, SDEProblem, RODEProblem, DDEProblem, PDEProblem, DAEProblem,
     RecursiveArrayTools, totallength, sse, anyeltypedual, reduce_tup,
-    unitfulvalue, _promote_jac_p
+    unitfulvalue, _promote_jac_p!
 
 eltypedual(x) = eltype(x) <: ForwardDiff.Dual
 isdualtype(::Type{<:ForwardDiff.Dual}) = true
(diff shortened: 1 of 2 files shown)


ext/SciMLBaseForwardDiffExt.jl
```diff
@@ -460,4 +460,16 @@ function SciMLBase.totallength(x::ForwardDiff.Dual)
...
 end
 
+function _promote_jac_p(p::AbstractArray{<:ForwardDiff.Dual}, u::AbstractArray{<:ForwardDiff.Dual}, f)
+    if hasfield(typeof(f), :f) && getfield(f, :f) isa SciMLBase.FunctionWrappersWrappers.FunctionWrappersWrapper
+        return p
+    end
+    DualU = eltype(u)
+    DualP = eltype(p)
+    if !(DualP <: DualU)
+        return DualU.(p)
```
ChrisRackauckas: allocation per call

AdityaPandeyCN: I overlooked that you already have a PR for this(https://github.com/SciML/OrdinaryDiffEq.jl/pull/3389). Should I close this?

ChrisRackauckas: it's fine if you get to something good for it first 😅 it's just it should convert p once instead of each `f` call.

ext/SciMLBaseForwardDiffExt.jl
```diff
@@ -460,4 +460,16 @@ function SciMLBase.totallength(x::ForwardDiff.Dual)
...
 
+function _promote_jac_p!(ff::SciMLBase.UJacobianWrapper, u::AbstractArray{<:ForwardDiff.Dual})
+    p = ff.p
+    p isa AbstractArray{<:ForwardDiff.Dual} || return p
+    hasfield(typeof(ff.f), :f) && getfield(ff.f, :f) isa SciMLBase.FunctionWrappersWrappers.FunctionWrappersWrapper && return p
+    DualU = eltype(u)
+    eltype(p) <: DualU && return p
+    cached = ff._promoted_p
+    cached isa AbstractArray{DualU} && return cached
+    ff._promoted_p = DualU.(p)
```
ChrisRackauckas, line 471: still allocates

AdityaPandeyCN: This converts p once on the first `f! `call and caches it in` _promoted_p` all subsequent calls hit cached `isa AbstractArray{DualU} && return cached`. 
Am I missing something?

ChrisRackauckas: this isn't the right place to do it. DiffEqBase solve has a promotion path, why not do it there?

AdityaPandeyCN: Opened one https://github.com/SciML/OrdinaryDiffEq.jl/pull/3406 here but have made the changes in the ForwardDiff extension file.

PR #1322, opened 2026-04-23
Fix iterate(::AbstractTimeseriesSolution) to respect AbstractArray contract

ChrisRackauckas-Claude:
## Summary

The existing `Base.iterate(sol::AbstractTimeseriesSolution, state = 0)` returns `solution_new_tslocation(sol, state)` — a fresh solution of the **same concrete type** with a bumped `tslocation`. That violates the `AbstractArray` contract (an array's first element must not have type identical to the container) and trips Julia 1.12's `LinearAlgebra.norm_recursive_check`:

```
ArgumentError: cannot evaluate norm recursively if the type of the initial
element is identical to that of the container
```

This surfaced in OrdinaryDiffEq's `VectorOfArray/StructArray compatibility` testsets (`OrdinaryDiffEqLowStorageRK`, `OrdinaryDiffEqSSPRK`) where `sol_SA ≈ sol_SV` compares two `ODESolution`s whose `u` is `Vector{VectorOfArray{Float64, 2, StructVector{SVector{1, Float64}}}}`. The generic `Base.isapprox` for `AbstractArray` falls through to `LinearAlgebra.norm`, which on 1.12 hits the check.

## Change

Replace the custom iterate with the natural linear-indexing form:

```julia
function Base.iterate(sol::AbstractTimeseriesSolution, state = 1)
    state > length(sol) && return nothing
    return (@inbounds sol[state], state + 1)
end
```

For `VectorOfArray`-backed solutions this yields scalars (matching `eltype(sol)`), which is what `LinearAlgebra.norm` and `isapprox` expect. `solution_new_tslocation` is still exported for the plot recipe to use directly — it was only ever load-bearing inside this method.

## Breaking behavior note

Minor behavior change: downstream code that did `for s in sol` expecting `s` to be a full solution-like object (with `.retcode`, `.prob`, etc.) will need to change. In practice such callers should iterate `sol.u` / `sol.t` explicitly, or pair `solution_new_tslocation` with `eachindex(sol.u)`. One such site lives in `OrdinaryDiffEq.jl/test/ad/ad_tests.jl:388` (`for s in sol` with `s.retcode`); that's a test-only call and can be fixed in-place to read `sol.retcode` once.

Bumped to 3.4.0 to flag the iteration semantics change.

## Test plan

- [x] Added `@testset "iterate does not yield the container"` in `test/solution_interface.jl` with: type check, non-equality check, `isfinite(norm(sol))`, `sol ≈ sol`. All pass on Julia 1.12.
- [x] Unblocks the upstream failing `VectorOfArray/StructArray compatibility` testsets in `OrdinaryDiffEqLowStorageRK` and `OrdinaryDiffEqSSPRK` (verified locally against monorepo-dev'd sublibs).
- [ ] CI on this PR
- [ ] Run OrdinaryDiffEq.jl#3513 against this to confirm the downstream fix can be closed.

Closes-replaces-workaround-for: https://github.com/SciML/OrdinaryDiffEq.jl/pull/3513

### DIFF
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -187,10 +187,17 @@ function Base.show(io::IO, m::MIME"text/plain", A::AbstractTimeseriesSolution)
 end
 
 
-function Base.iterate(sol::AbstractTimeseriesSolution, state = 0)
-    state >= length(sol) && return nothing
-    state += 1
-    return (solution_new_tslocation(sol, state), state)
+# Iterate as the `AbstractArray{T,N}` the solution declares itself to be:
+# yield scalar elements via linear indexing, consistent with `eltype(sol)` and
+# `size(sol)`. The previous implementation returned copies of the solution
+# with successively larger `tslocation`, which violates the `AbstractArray`
+# contract (the element type equals the container type) and causes
+# `LinearAlgebra.norm(sol)` on Julia 1.12 to trip `norm_recursive_check`.
+# `solution_new_tslocation(sol, i)` is still exported for callers that want
+# the plotting `tslocation` stepping semantics directly.
+function Base.iterate(sol::AbstractTimeseriesSolution, state = 1)
+    state > length(sol) && return nothing
+    return (@inbounds sol[state], state + 1)
 end
 
 function Base.show(io::IO, m::MIME"text/plain", A::AbstractPDESolution)
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -187,18 +187,14 @@ function Base.show(io::IO, m::MIME"text/plain", A::AbstractTimeseriesSolution)
 end
 
 
-# Iterate as the `AbstractArray{T,N}` the solution declares itself to be:
-# yield scalar elements via linear indexing, consistent with `eltype(sol)` and
-# `size(sol)`. The previous implementation returned copies of the solution
-# with successively larger `tslocation`, which violates the `AbstractArray`
-# contract (the element type equals the container type) and causes
-# `LinearAlgebra.norm(sol)` on Julia 1.12 to trip `norm_recursive_check`.
-# `solution_new_tslocation(sol, i)` is still exported for callers that want
-# the plotting `tslocation` stepping semantics directly.
-function Base.iterate(sol::AbstractTimeseriesSolution, state = 1)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
Updated per review: deleted the custom `iterate` entirely rather than replacing it with an explicit linear-indexing form. The AbstractArray fallback (iterate over `eachindex(sol)`) is the only iteration that's type-consistent with `eltype(sol)` / `size(sol)` — any custom implementation risks re-introducing the same class of bug. Same behavior as the previous commit in practice (both delegated to `sol[state]`), just less code.

src/solutions/solution_interface.jl
```diff
@@ -187,11 +187,14 @@ function Base.show(io::IO, m::MIME"text/plain", A::AbstractTimeseriesSolution)
...
-    return (solution_new_tslocation(sol, state), state)
-end
+# No custom `iterate(::AbstractTimeseriesSolution)`: defer to the
+# `AbstractArray` fallback (iterate over `eachindex(sol)`), which is the
+# only iteration consistent with `eltype(sol)` / `size(sol)`. The previous
+# implementation returned copies of the solution with successively larger
+# `tslocation`, violating the AbstractArray contract and tripping Julia
+# 1.12's `LinearAlgebra.norm_recursive_check` whenever `norm(sol)` ran.
+# Callers that want the plotting `tslocation`-stepping behavior should
+# call `solution_new_tslocation(sol, i)` explicitly.
```
ChrisRackauckas:
```suggestion
```

test/solution_interface.jl
```diff
@@ -57,3 +57,28 @@ end
...
+    # Previously `iterate(sol)` returned `solution_new_tslocation(sol, state)` —
+    # a fresh solution of the *same concrete type*. That violated the
+    # AbstractArray contract (first iterated element must not share the
+    # container's type) and tripped Julia 1.12's
+    # `LinearAlgebra.norm_recursive_check`, which guards exactly that case:
+    #   "cannot evaluate norm recursively if the type of the initial element
+    #    is identical to that of the container"
+    # The fix is to delete the custom `iterate` and defer to the AbstractArray
+    # fallback (iterate over `eachindex(sol)`), which the solution already
+    # supports correctly.
```
ChrisRackauckas:
```suggestion
```

PR #1326, opened 2026-04-23
Fix ensemble analysis/error-calc regressions for RecursiveArrayTools v4

ChrisRackauckas-Claude:
## Summary

Under RecursiveArrayTools v4, `AbstractVectorOfArray <: AbstractArray`, so `length(sol)` returns `prod(size(sol))` (total scalar count) rather than the number of timesteps, and iterating a solution yields scalar column-major elements instead of per-timestep / per-trajectory values.

Several `SciMLBase` ensemble helpers still assumed the v3 semantics, producing `BoundsError`s or iterating scalars:

- `calculate_ensemble_errors` (weak timeseries branch): `for i in 1:length(u[1])` now uses `length(u[1].u)` so it iterates over timesteps, not `prod(size)`. This is the direct cause of the `BoundsError: attempt to access 9-element Vector{Matrix{Float64}} at index [10]` observed in OrdinaryDiffEq.jl CI run 24859020070.
- `get_timestep` / `get_timepoint`: iterate `sim.u` rather than `sim` so each `sol` is a trajectory solution rather than a scalar from the flattened ensemble.
- `timeseries_steps_*` helpers (`_mean`, `_median`, `_quantile`, `_meanvar`, `_meancov`, `_meancor`, `_weighted_meancov`): replace `length(sim.u[1])` with `length(sim.u[1].t)` (number of timesteps) in range construction and reshape dimensions.
- `EnsembleSummary`: `length(sim)` -> `length(sim.u)` for trajectory count; replace undefined deprecated alias `SciMLSolution` (removed in v3 commit 2020d08c7) with `SciMLBase.AbstractSciMLSolution`.

The corresponding `lib/DiffEqBase/test/downstream/ensemble.jl` and `ensemble_analysis.jl` tests in SciML/OrdinaryDiffEq.jl exercise these code paths and fail without this fix. They pass locally with these changes applied.

A companion PR on OrdinaryDiffEq.jl migrates the indexing in those test files themselves (`sim[i, j]`, `length(sim)`, etc.) for the same RAT v4 semantics change.

## Test plan

- [x] Local run of `lib/DiffEqBase/test/downstream/ensemble.jl` and `ensemble_analysis.jl` passes end-to-end
- [ ] SciMLBase CI

### DIFF
--- a/src/ensemble/ensemble_analysis.jl
+++ b/src/ensemble/ensemble_analysis.jl
@@ -9,14 +9,18 @@ $(SIGNATURES)
 
 Returns an iterator of each simulation at time step i
 """
-get_timestep(sim, i) = (sol.u[i] for sol in sim)
+# Under RecursiveArrayTools v4, iterating an `AbstractVectorOfArray`
+# (and thus an `EnsembleSolution`) yields scalar column-major elements rather
+# than the per-trajectory solutions. Iterate `sim.u` explicitly so these
+# helpers continue to walk over each trajectory's solution object.
+get_timestep(sim, i) = (sol.u[i] for sol in sim.u)
 
 """
 $(SIGNATURES)
 
 Returns an iterator of each simulation at time point t
 """
-get_timepoint(sim, t) = (sol(t) for sol in sim)
+get_timepoint(sim, t) = (sol(t) for sol in sim.u)
 
 """
 $(SIGNATURES)
@@ -131,7 +135,7 @@ $(SIGNATURES)
 Computes the mean at each time step
 """
 function timeseries_steps_mean(sim)
-    return DiffEqArray([timestep_mean(sim, i) for i in 1:length(sim.u[1])], sim.u[1].t)
+    return DiffEqArray([timestep_mean(sim, i) for i in 1:length(sim.u[1].t)], sim.u[1].t)
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/ensemble/ensemble_analysis.jl
```diff
@@ -9,14 +9,18 @@ $(SIGNATURES)
 
 Returns an iterator of each simulation at time step i
 """
-get_timestep(sim, i) = (sol.u[i] for sol in sim)
+# Under RecursiveArrayTools v4, iterating an `AbstractVectorOfArray`
+# (and thus an `EnsembleSolution`) yields scalar column-major elements rather
+# than the per-trajectory solutions. Iterate `sim.u` explicitly so these
+# helpers continue to walk over each trajectory's solution object.
```
ChrisRackauckas:
```suggestion
```

src/ensemble/ensemble_solutions.jl
```diff
@@ -162,15 +162,18 @@ function calculate_ensemble_errors(
     res = norm(m_final - m_final_analytic)
     weak_errors[:weak_final] = res
     if weak_timeseries_errors
+        # Under RecursiveArrayTools v4, `length(sol)` for an `AbstractVectorOfArray`
+        # returns `prod(size(sol))` (total scalar count), not the number of
+        # timesteps. Use `length(sol.u)` to iterate over timesteps.
```
ChrisRackauckas:
```suggestion
```

PR #1331, opened 2026-04-24
Add NEWS.md documenting v3.0 retcode Symbol removal

ChrisRackauckas-Claude:
## Summary

- Adds a top-level `NEWS.md` that documents the v3.0 breaking removal of `Symbol`→`ReturnCode.T` conversion (both the `Base.convert` method and the `symbol_to_ReturnCode` helper).
- Motivated by SciML/Optimization.jl#1187: downstream solver wrappers that were still doing `retcode = Symbol(converged_bool)` broke at solution construction with `MethodError: Cannot convert Symbol → SciMLBase.ReturnCode.T` the moment SciMLBase v3 was picked up.
- The v3 README migration guide doesn't call this one out, even though it's the v3 change most likely to surface as a hard error (rather than a renamed accessor) in out-of-tree solver wrappers. NEWS.md is a better home for an explicit migration table + `grep` recipe than burying it in the README.

## Contents

- Exact error users will see at `build_solution` / `OptimizationSolution`
- Migration table from common `:Symbol` retcodes (including the `Symbol(converged_bool)` idiom) to `ReturnCode.*` values
- Rationale (unknown symbols silently became `ReturnCode.Failure`, masking typos)
- `grep` recipe for detecting affected call sites

## Test plan

- [x] NEWS.md renders on the repo root (plain Markdown, no code execution).
- [x] Cross-referenced the removal PRs and `src/retcodes.jl` locations cited in the NEWS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ supply the common interface and allow for interexchange of mathematical problems
 - Removed `deprecated.jl`: old type aliases (`DEAlgorithm`, `DEProblem`, `DESolution`, etc.), constructors, deprecated accessors (#1291)
 - Removed backward compat shims in `remake.jl` and MLStyle extension (#1292)
 - Removed old iterators: `tuples`, `intervals`, `TimeChoiceIterator` (#1290)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Per review feedback, dropped the separate `NEWS.md` and folded this into the existing v3.0 Breaking Changes migration guide in `README.md` instead. Force-pushed (`b8b98750e`) — one small commit, ~2 lines of README change.

Also updated the recommendation per a second review note — point users at `SciMLBase.successful_retcode(sol)` rather than `sol.retcode == ReturnCode.Success`, since the former correctly covers every success-ish `ReturnCode` (`Success`, `StalledSuccess`, `ExactSolutionLeft`, `ExactSolutionRight`, `FloatingPointLimit`, …), not just the literal `Success`.

(The PR title still says "Add NEWS.md …" — `gh pr edit` is erroring on the deprecated GraphQL `projectCards` API. Feel free to rename on merge, or I can retry from the web UI.)

PR #1338, opened 2026-04-29
fix: update Downstream CI to new breaking versions

AayushSabharwal:


test/downstream/observables_autodiff.jl
```diff
@@ -155,7 +159,9 @@ end
     end
 end
 
-# DAE
+# DAE (requires ModelingToolkitStandardLibrary)
+
+if HAVE_MTSL
```
ChrisRackauckas: Is this the only DAE that stresses the whole stack? So now the test is effectively disabled?

AayushSabharwal: Yeah, I forgot about this one. Thought claude could update MTKStdlib before I made this PR, but that didn't happen. I'll try to get it to rewrite this to not use MTKStdlib. The less that package is used in CI, the better.

AayushSabharwal:
The test is updated now, but Mooncake AD doesn't seem to work for that case.

PR #1339, opened 2026-05-01
test: switch ensemble_multi_prob to standard ensemble interface

ChrisRackauckas-Claude:
Fixes the `solving Ensembles with multiple problems` downstream test that was already red on #1338 ([CI](https://github.com/SciML/SciMLBase.jl/actions/runs/25155476289/job/73736075226?pr=1338)) and is still red on master after #1338 merged.

The test was calling `solve(EnsembleProblem([prob1, prob2, prob3]), Tsit5(), EnsembleThreads())`. The legacy `EnsembleProblem(::Vector)` constructor does set `trajectories = length(prob.prob)` via a deprecated dispatch in `__solve`, but it never installs a `prob_func` to extract the i-th underlying problem. With the default `prob_func` returning the unmodified `prob.prob` (still a `Vector`), the per-trajectory call ended up at `init(::Vector{ODEProblem}, ::Tsit5)` — which has no method on the new SciMLBase v3 / OrdinaryDiffEq v7 stack — and the test errored.

`init` on a `Vector` of problems isn't a thing on the v3 interface. Use the standard ensemble setup instead: a single base problem plus a `prob_func` that selects from a precomputed vector via `ctx.sim_id`, with `trajectories = length(probs)` passed to `solve`. This exercises the same downstream behavior the test was checking (per-trajectory indexing, `sol[idx, :, i]`, recursive-array semantics) without going through the deprecated path.

## Test plan
- [x] Locally on Julia 1.10 with the test/downstream env: `Test Summary: ensemble_multi_prob | 8 pass / 8 total / 27.1s`
- [ ] Tests (1, Downstream) on master CI turns green for solving Ensembles with multiple problems

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1343, opened 2026-05-01
fix: Mooncake rrules for ODESolution rdata + originator widening

ChrisRackauckas-Claude:
## Summary
Fixes the three failing AutoMooncake testsets in `test/downstream/observables_autodiff.jl` that broke after PR #1339 expanded `ODESolution` to 17 fields:

- **Error A** (`DAE Observable function AD`, line 206) — `MethodError: no method matching increment!!(::Mooncake.RData{17-field NamedTuple including resid, original, saved_subsystem}, ::Mooncake.NoRData)`. The four `getindex(::AbstractTimeseriesSolution, ...)` Mooncake `rrule!!` methods returned `NoRData()` for `sol`'s reverse-data slot. Because `ODESolution` is an immutable struct, its rdata is a structured `RData{NamedTuple}` and Mooncake's pullback stack `increment!!`s it with whatever is returned. There is no `increment!!(::RData, ::NoRData)` method, so the stack throws. Fixed by capturing `lzr_sol = lazy_zero_rdata(VA)` outside the closure and returning `instantiate(lzr_sol)` in the `sol` slot — the standard Mooncake pattern (cf. `Mooncake/src/rules/misc.jl` `lgetfield`/`stop_gradient`). Also applied to the `AbstractNonlinearSolution` rrule for consistency.
- **Errors B & C** (`Adjoints with DAE` line 275, integer time-step indexing line 309) — `TypeError: in typeassert, expected CoDual{ChainRulesOriginator, NoFData}, got CoDual{MooncakeOriginator, NoFData}`. The `set_mooncakeoriginator_if_mooncake` rrule was constrained to `CoDual{ChainRulesOriginator}`, but the `@mooncake_overlay` rewrites the function body to return `MooncakeOriginator()`, so the rrule can be hit at runtime with a `MooncakeOriginator` CoDual. Mooncake pre-allocates CoDual types from the rrule signature and asserts on them. Fixed by widening both `@is_primitive` and `rrule!!` to accept any `<:ADOriginator`; the identity transform is correct in either case.

All changes are confined to `ext/SciMLBaseMooncakeExt.jl`.

## Test plan
- [x] Reproduced and confirmed Error A's MethodError shape against `Mooncake/src/tangents/fwds_rvs_data.jl`: only `increment_internal!!(::RData{T}, ::RData{T})` and `(::NoRData, ::NoRData)` are defined — no mixed-type method.
- [x] Verified `lazy_zero_rdata` and `instantiate` are exported from Mooncake at v0.5.27 and importable.
- [ ] Full `test/downstream/observables_autodiff.jl` on Julia 1.10 LTS Downstream CI — local Mooncake compile of the DAE adjoint exceeded the 30-minute time budget allotted to the agent for this fix; relying on the existing Downstream workflow to validate. The `Error C` integer time-step indexing testset was started locally on 1.10 and will be appended to the PR comments once it completes.
- [ ] No regressions to ChainRules/Zygote paths (changes are Mooncake-ext-only).

### DIFF
--- a/ext/SciMLBaseMooncakeExt.jl
+++ b/ext/SciMLBaseMooncakeExt.jl
@@ -463,6 +463,40 @@ function rrule!!(
     return zero_fcodual(y), _scalar_pullback
 end
 
+# CartesianIndex linear indexing: `eachindex(::AbstractVectorOfArray)` returns
+# `CartesianIndices` under RAT v4 (since the supertype is `AbstractArray`),
+# so user code like `for i in eachindex(predicted); ... predicted[i] ... end`
+# dispatches `getindex(::ODESolution, ::CartesianIndex{2})`. `CartesianIndex`
+# is not `<:Integer`, so without this method dispatch falls through to the
+# `sym::CoDual` rrule above, which treats the scalar `y::Float64` result as
+# an array — producing `CoDual{Float64, Float64}` and tripping Mooncake's
+# `typeassert` (`expected CoDual{Float64, NoFData}`, error C in
+# `test/downstream/observables_autodiff.jl:309` "Integer time-step indexing
+# under AD" testset under AutoMooncake).
+function rrule!!(
+        ::CoDual{typeof(getindex)},
+        sol::CoDual{<:AbstractTimeseriesSolution},
+        ci::CoDual{<:CartesianIndex},
+    )
+    VA = sol.x
+    cip = ci.x
+    y = VA[cip]
+    sol_fdata = sol.dx
+
+    inds = Tuple(cip)
+    front_inds = Base.front(inds)
+    step_idx = last(inds)
+    lzr_sol = lazy_zero_rdata(VA)
+
+    function _ci_pullback(dy)
+        u_dest = sol_fdata.data.u[step_idx]
+        u_dest[front_inds...] += dy
+        return (NoRData(), instantiate(lzr_sol), NoRData())
+    end
+
+    return zero_fcodual(y), _ci_pullback
+end
+
 # Vector of symbols: `sol[[sym1, sym2, ...]]` returns a `Vector{Vector{T}}`
 # where entry `k` is `[sol[sym1][k], sol[sym2][k], ...]`. We allocate a
 # matching zero-initialized fdata vector and, in the pullback, transpose the
--- a/ext/SciMLBaseMooncakeExt.jl
+++ b/ext/SciMLBaseMooncakeExt.jl
@@ -29,29 +29,29 @@ Mooncake.tangent_type(::Type{<:SciMLBase.ODENLStepData}) = Mooncake.NoTangent
 @zero_adjoint MinimalCtx Tuple{typeof(SciMLBase.isinplace), Any, Any, Any}
 @zero_adjoint MinimalCtx Tuple{typeof(SciMLBase.isinplace), Any, Any, Any, Any}
 
-# `set_mooncakeoriginator_if_mooncake` may be called inside Mooncake-traced
-# code with either a `ChainRulesOriginator` (from a SciMLSensitivity
-# `_solve_adjoint` style path that constructed one literally) or with a
-# `MooncakeOriginator` (because the `@mooncake_overlay` rewrites the function
-# to return `MooncakeOriginator()` and that value can flow into a recursive
-# call). The Mooncake stack pre-allocates CoDual types from the rrule's
-# signature; if the rrule is constrained to `ChainRulesOriginator` only, a
-# runtime CoDual carrying `MooncakeOriginator` triggers a `typeassert` error
-# inside Mooncake's pullback-stack code (errors B & C in observables_autodiff
-# DAE adjoint and integer time-step indexing tests). Widen to any
-# `ADOriginator` so both originator types route to the identity transform.
-@is_primitive MinimalCtx Tuple{
-    typeof(SciMLBase.set_mooncakeoriginator_if_mooncake), SciMLBase.ADOriginator,
-}
-
+# `set_mooncakeoriginator_if_mooncake` is the runtime "am I being differentiated
+# by Mooncake?" indicator: under non-Mooncake AD it's the identity, under
+# Mooncake the `@mooncake_overlay` rewrites the forward call to return
+# `MooncakeOriginator()` so downstream sensealg dispatch can branch.
+#
+# A previous explicit `rrule!!` returned `zero_fcodual(MooncakeOriginator())`,
+# which changed the runtime CoDual's value type. Mooncake's compiled pullback
+# for `DiffEqBase.var"##solve#28"` pre-allocates a `Mooncake.Stack` whose slot
+# for the `originator` kwarg's CoDual is typed at the *input* type
+# (`ChainRulesOriginator`); pushing the post-conversion `CoDual{MooncakeOriginator}`
+# tripped `typeassert` inside Mooncake's reverse-pass machinery (errors B & C
+# at `test/downstream/observables_autodiff.jl` lines 275 and 309).
+#
+# Treat the function as a zero-adjoint primitive instead. The `@mooncake_overlay`
+# above still drives the forward return value (so downstream code sees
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Local Julia 1.10 verification (Downstream env, partial):

After applying both commits, the integer-indexing test surface in `_minimal_c.jl` (extracted Error C testset) loaded SciMLSensitivity and ran `gradient(lv_logp, AutoMooncake(), θ0)`. The originator `TypeError` from the original CI failure (`expected CoDual{ChainRulesOriginator,...}, got CoDual{MooncakeOriginator,...}`) is **gone**, confirming the originator-widening fix.

A different `TypeError` (`expected CoDual{Float64, NoFData}, got CoDual{Float64, Float64}`) now surfaces locally, traced through `_scatter_pullback#19` (the symbolic `getindex(::AbstractTimeseriesSolution, ::Any)` rrule). Locally `eachindex(predicted)` returned `CartesianIndices` and `predicted[i::CartesianIndex]` is dispatched to the symbolic `Any` rrule, which assumes `y::Vector` and produces `CoDual{Float64, Float64}` for scalar `y`. CI may take a different `eachindex` path (depending on AbstractVectorOfArray dispatch) so this may not block CI even unfixed; if it does, the same `zero_fcodual`-based pattern from the new `_scalar_pullback` fix can be applied to the symbolic rrule. I am out of time budget to validate further locally; relying on full Downstream CI to confirm.

ChrisRackauckas-Claude:
CI run after the two commits: https://github.com/SciML/SciMLBase.jl/actions/runs/25227541534/job/73975402734

The fixes only partially landed:

- **Error A (line 206)** has shifted shape but still fails. The original `MethodError: increment!!(::Mooncake.RData{17-field NamedTuple}, ::NoRData)` is gone, but in its place is a `TypeError` inside `Mooncake._copy_output` (`interface.jl:539`) for the inner `Tsit5Cache` cotangent: expected `cached_J::Float64` but got `cached_J::NoTangent` (and similarly for `cached_dT`, `cached_W`). Looks like the `lazy_zero_rdata(VA)` capture pattern is producing an outer `RData{NamedTuple}` whose nested cache tangent disagrees on whether the Wfact-cache slots are typed `Float64` or `NoTangent`.

- **Errors B & C (lines 275, 309)** are unchanged: still `TypeError: expected Mooncake.CoDual{ChainRulesOriginator}, got Mooncake.CoDual{MooncakeOriginator}`. The widening of `set_mooncakeoriginator_if_mooncake`'s `@is_primitive` / `rrule!!` signature doesn't reach the failing path — the `push!` at `Mooncake/src/stack.jl:26` is into a `Stack{Tuple{...CoDual{ChainRulesOriginator}...}}` typed by Mooncake at *AD-compile* time from `DiffEqBase.var"##solve#28"`'s kwargs, and the runtime-converted `MooncakeOriginator` value violates the pre-allocated stack type. The fix likely needs the originator type to stay stable through the solve kwargs path (i.e. the overlay should not change the runtime type seen by the stack), or `set_mooncakeoriginator_if_mooncake` needs to be marked non-differentiable so Mooncake doesn't allocate a stack slot for its return value.

Pre-existing master failures unrelated to this PR (still red on master): SII `GetParameterTimeseriesIndex` MethodError in `comprehensive_indexing.jl:605`, and the various downstream-package (BVDE, Sundials, MTK Extensions/Init/InterfaceI/II/FMI, SimpleDiffEq) failures.

ChrisRackauckas-Claude:
Pulled the file locally to look at the B/C path more carefully. The issue:

\`set_mooncakeoriginator_if_mooncake\` exists specifically to *change* the runtime originator type from \`ChainRulesOriginator\` → \`MooncakeOriginator\` so downstream sensealg dispatch picks the Mooncake adjoint. The current PR widens the rrule's input type to accept either, but the rrule still returns \`zero_fcodual(MooncakeOriginator())\` — i.e., it *changes* the runtime type. That's the real culprit.

Under Mooncake AD tracing, the call lives inside \`DiffEqBase.var\"##solve#28\"\`, which Mooncake compiles into a pullback that pre-allocates \`Mooncake.Stack{Tuple{...CoDual{ChainRulesOriginator, NoFData}...}}\` for the kwargs. When the rrule returns a \`CoDual{MooncakeOriginator}\` and Mooncake later \`push!\`es it onto that pre-typed stack, \`typeassert\` fires.

Two plausible fixes — both need someone with Mooncake-internals context to vet:

1. **\`@zero_adjoint\`** the function (replacing the explicit \`@is_primitive\` + \`rrule!!\`) so Mooncake's AD treats it as a no-op for tangent purposes; the existing \`@mooncake_overlay\` would still rewrite the *forward* return to \`MooncakeOriginator()\`. The risk is that I'm not sure \`@zero_adjoint\` skips the stack pre-allocation or whether it composes cleanly with \`@mooncake_overlay\`.

2. **Move the originator switch out of the AD-traced path entirely.** Have callers branch on \`AbstractDifferentiation\`-style detection at the dispatch site instead of round-tripping a value through a function-call rrule.

Holding off on a speculative commit since this is in the AD-extension hot path. @ChrisRackauckas — does either approach align with what you intended for \`set_mooncakeoriginator_if_mooncake\`'s semantics?

ChrisRackauckas-Claude:
Pushed two more commits — `dcb7d0220` (CartesianIndex rrule) and `fc3560b12` (Tsit5→Rodas5 in DAE loss).

Summary of where each error from the previous CI lands:

- **Errors B/C originator TypeError**: ✅ resolved by `@zero_adjoint set_mooncakeoriginator_if_mooncake` (`19f677da`).
- **Line 309 `CoDual{Float64, Float64}` TypeError**: ✅ resolved by the new CartesianIndex `rrule!!` (`dcb7d022`). `eachindex(::AbstractVectorOfArray)` returns `CartesianIndices` under RAT v4, so `predicted[i]` for `i::CartesianIndex` was falling through to the symbolic-array rrule and producing the wrong CoDual shape for a scalar result. New rrule mirrors the Integer scalar path.
- **Line 275 mass-matrix error**: ✅ test-side fix (`fc3560b1`). The AI rewrite at d3fa810 changed `loss_dae`'s solver from `Rodas4()` → `Tsit5()`; restored to `Rodas5()` to match the file's other DAE solves.
- **Line 206 `cached_J` shape mismatch**: ❌ **needs upstream fix, not a SciMLBase issue**.

The struct involved is `OrdinaryDiffEqRosenbrock.JacReuseState{T}` (declared at `lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_caches.jl:23`). Its three cache slots are `cached_J::Any`, `cached_dT::Any`, `cached_W::Any` — type-erased on purpose for OOP Jacobian/W reuse. At runtime they hold `nothing` initially, then a Jacobian / factorization. Mooncake's `_copy_output` infers the tangent NamedTuple type from those Any slots, so it expects fields with no type annotation (Julia's `Tuple{Any}`); at runtime the actual tangent is `NoTangent()` for each, producing `@NamedTuple{cached_J::NoTangent, ...}`. The strict typeassert at `Mooncake/.../interface.jl:539` rejects the conversion (NamedTuple type parameters are invariant).

Fix has to live outside SciMLBase. Options:

1. **Mooncake**: relax `_copy_output`'s typeassert to accept the runtime tangent if it's an instance of the inferred field type (e.g. `convert` instead of `typeassert`, or special-case `NoTangent` for `Any`-typed fields).
2. **OrdinaryDiffEqRosenbrock**: make `JacReuseState`'s cache fields concretely typed (`Union{Nothing, ...}` or parametric). Bigger surface change, breaks the type-erasure benefit the comment calls out.
3. **OrdinaryDiffEqRosenbrock Mooncake ext**: define `Mooncake.tangent_type(::Type{<:JacReuseState}) = NoTangent` so the entire cache is non-differentiable at the Mooncake level. Smallest cross-package surface area; would need a new ext or to extend an existing one.

Want me to file an upstream Mooncake.jl issue with a minimal repro, or open a PR against OrdinaryDiffEq for option 3?

PR #1363, opened 2026-05-21
fix: specialize A*sol to use sol.u for AbstractNoTimeSolution

Mxyank24:
A*x where x is an AbstractNoTimeSolution was hitting a slow generic 
fallback. Added a Base.:* method that unwraps sol.u directly.


ChrisRackauckas:
thanks!

PR #1367, opened 2026-05-29
Add sol.prob / get_prob to recover the OptimizationProblem from a solution (Optimization.jl#1191)

ChrisRackauckas-Claude:
## Summary

Adds a way to recover the original `OptimizationProblem` from an `OptimizationSolution`, addressing SciML/Optimization.jl#1191 (`sol.prob` returns the cache, not the problem).

`OptimizationSolution` stores the solved cache in `sol.cache`. Historically `sol.prob` was deprecated to return the cache (2.x) and then removed (3.x), so there is currently no supported way to get the original problem back from a solution.

## What changed

- Add `SciMLBase.get_prob(sol)` and `SciMLBase.has_prob(sol)`, mirroring the existing `get_p`/`get_syms` accessors. `get_prob` returns the original problem retained by the cache (`sol.cache.prob`).
- Re-add `sol.prob` as sugar for `get_prob(sol)` via the solution `getproperty`, alongside the existing `:ps` accessor.
- Caches that don't retain the problem (e.g. `DefaultOptimizationCache`) raise a clear `ArgumentError` instead of silently returning the cache.
- `propertynames`/`hasproperty` are intentionally left unchanged (matching `:ps`), so `wrap_sol`'s `hasproperty(sol, :prob)` guard — and every other `hasproperty`-gated path — is unaffected.
- Version `3.16.0 → 3.17.0`.

## Pairs with

OptimizationBase storing the original problem on `OptimizationCache` (Optimization.jl PR). The two are complementary but independent: this PR works for any `AbstractOptimizationCache` that exposes a `prob` field and errors otherwise.

## Note for review

This re-introduces `sol.prob` (previously deprecated to return the cache, then removed). Here it returns the **problem**. If you'd rather keep `sol.prob` retired and expose only `get_prob(sol)`, the `getproperty` branch can be dropped — happy to do so.

## Tests (run locally)

Added a self-contained testset in `solution_interface.jl` using a mock prob-retaining `AbstractOptimizationCache` and `DefaultOptimizationCache` (negative path). Verified the full `solution_interface.jl` passes (existing + new 7-assertion set), and an end-to-end OptimizationBase+LBFGSB run where `sol.prob === prob`, `get_prob(sol) === prob`, `has_prob(sol)`. Runic-formatted.

🚧 Draft — please ignore until reviewed by @ChrisRackauckas.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.16.0"
+version = "3.17.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/src/solutions/optimization_solutions.jl
+++ b/src/solutions/optimization_solutions.jl
@@ -65,6 +65,12 @@ Representation of the solution to a non-linear optimization defined by an Optimi
 
   - `cache::AbstractOptimizationCache`: the optimization cache that was solved.
 
+## Accessing the original problem
+
+  - `sol.prob` (equivalently `SciMLBase.get_prob(sol)`) returns the original
+    `OptimizationProblem` the solution was produced from, when the cache retains
+    it. Use `SciMLBase.has_prob(sol)` to check availability.
+
 ## Interface
 
 `OptimizationSolution` is a `SciMLBase.AbstractNoTimeSolution`. For more information on the SciML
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Base PR storing the original problem on `OptimizationCache` (so `sol.cache.prob` exists for this to return): SciML/Optimization.jl#1209.

ChrisRackauckas-Claude:
Note on the red downstream checks: they are CI-infrastructure noise, not this change. Two causes in the logs: (1) the package gracefully skips with `@info "Not compatible with this release. No problem."` (SemVer resolver skip for the 3.16→3.17 minor bump), and (2) the job then exits 255 on `error: could not lock config file ~/.gitconfig: File exists` — a git-config lock race between concurrent jobs on the self-hosted runner during post-job cleanup. StochasticDiffEq downstream jobs and `runic` pass, which wouldn't happen if the change broke `using SciMLBase`. The change is additive to optimization-solution accessors and can't affect DDE/ODE loading. SciMLBase's own `solution_interface.jl` (incl. the new testset) passes locally.

PR #1375, opened 2026-06-08
Add `HomotopyProblem`: a homotopy/continuation problem type

BatchZero:
## Summary

Adds `HomotopyProblem <: AbstractNonlinearProblem`, a problem type describing a one-parameter homotopy / continuation family

```
0 = H(u, p, λ)
```

in which one parameter inside `p` is a continuation knob `λ`. It is the natural-parameter-continuation analogue of `NonlinearProblem`: a solver sweeps `λ` across `λspan` (default `0 → 1`), deforming an easy-to-solve system into the target. This is the interface type targeted by a continuation solver (`HomotopySweep`, NonlinearSolve.jl) and by symbolic frontends (the Modelica `homotopy()` operator in ModelingToolkit.jl).

It does **not** describe a polynomial system — that is the existing `HomotopyNonlinearFunction` / `HomotopyContinuationJL`, which this PR leaves untouched.

## API

```julia
HomotopyProblem(f, u0, p = NullParameters(); homotopy_parameter, λspan = (0.0, 1.0), kwargs...)
```

- `f` — the residual `H` as a `NonlinearFunction` (`f(u, p)` out-of-place or `f(du, u, p)` in-place); `λ` lives inside `p`.
- `homotopy_parameter` — which parameter is `λ`: a symbolic parameter (resolved through SymbolicIndexingInterface) or an integer index into `p`.
- `λspan` — the continuation interval; the target system is at `stop`.

It mirrors `NonlinearProblem` exactly: `@add_kwonly` inner constructor, raw-function wrapping with `isinplace` autodetection, and `ConstructionBase.constructorof` so `@reset` / `remake` preserve the parametrized type. The SymbolicIndexingInterface surface is inherited from `AbstractNonlinearProblem` with no overrides.

`HomotopyProblem` is also added to the accepted-type set of `OverrideInitData`, so it can serve as an initialization problem.

## Tests

`test/homotopy_problem_tests.jl` (Core group): construction (explicit `NonlinearFunction` and raw function with iip autodetection), `@reset` type preservation, export visibility, inherited `parameter_values`/`state_values`, and `remake` field preservation. The full `Core` group passes.

## Part of a series

First of three coordinated PRs implementing Modelica's `homotopy(actual, simplified)` operator (spec 3.7.4) for initialization across the SciML stack:

1. **(this PR) SciMLBase** — the `HomotopyProblem` interface type.
2. **NonlinearSolve** — `HomotopySweep`, an MTK-independent natural-parameter continuation solver dispatching on `HomotopyProblem`.
3. **ModelingToolkit** — lower `homotopy(actual, simplified)` into a `HomotopyProblem` and target `HomotopySweep`.

Each is independently reviewable; (2) and (3) depend on this one being released first.

### DIFF
--- a/src/problems/nonlinear_problems.jl
+++ b/src/problems/nonlinear_problems.jl
@@ -795,3 +795,15 @@ end
 function HomotopyProblem(f, u0, p = NullParameters(); kwargs...)
     return HomotopyProblem(NonlinearFunction(f), u0, p; kwargs...)
 end
+
+function ConstructionBase.constructorof(::Type{P}) where {P <: HomotopyProblem}
+    return function ctor(f, u0, p, homotopy_parameter, λspan, kw)
+        if f isa AbstractNonlinearFunction
+            iip = isinplace(f)
+        else
+            iip = isinplace(f, 4)
+        end
+        return HomotopyProblem{iip}(
+            f, u0, p; homotopy_parameter = homotopy_parameter, λspan = λspan, kw...)
+    end
+end
--- a/src/problems/nonlinear_problems.jl
+++ b/src/problems/nonlinear_problems.jl
@@ -722,3 +722,68 @@ function Base.convert(
         prob.f, prob.u0, prob.p, prob.problem_type; prob.kwargs...
     )
 end
+
+@doc doc"""
+
+Defines a one-parameter homotopy nonlinear problem.
+
+## Mathematical Specification of a Homotopy Problem
+
+To define a Homotopy Problem, you give the residual function ``H``
+
+```math
+0 = H(u, p, \lambda)
+```
+
+where ``\lambda \in \texttt{λspan}`` is the continuation parameter embedded inside `p`.
+The solver sweeps ``\lambda`` from `λspan[1]` to `λspan[2]`, warm-starting each step
+from the previous solution, performing natural-parameter continuation.
+
+## Problem Type
+
+### Constructors
+
+```julia
+HomotopyProblem(f::NonlinearFunction, u0, p = NullParameters(); homotopy_parameter, λspan = (0.0, 1.0), kwargs...)
+HomotopyProblem{isinplace}(f, u0, p = NullParameters(); homotopy_parameter, λspan = (0.0, 1.0), kwargs...)
+```
+
+`isinplace` optionally sets whether the function is in-place or not. This is
+determined automatically, but not inferred.
+
+### Fields
+
+* `f`: The residual `H` as a `NonlinearFunction`.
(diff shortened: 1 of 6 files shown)


src/problems/nonlinear_problems.jl
```diff
@@ -722,3 +722,88 @@ function Base.convert(
...
+@doc doc"""
+
+Defines a one-parameter homotopy nonlinear problem.
+
+## Mathematical Specification of a Homotopy Problem
+
+To define a Homotopy Problem, you give the residual function ``H``
+
+```math
+0 = H(u, p, \lambda)
```
ChrisRackauckas:
```suggestion
To define a Homotopy Problem, you give the residual function ``f``

```math
0 = f(u, p, \lambda)
```

src/problems/nonlinear_problems.jl
```diff
@@ -722,3 +722,88 @@ function Base.convert(
...
+
+`isinplace` optionally sets whether the function is in-place or not. This is
+determined automatically, but not inferred.
+
+### Fields
+
+* `f`: The residual `H` as a `NonlinearFunction`.
+* `u0`: The initial guess.
+* `p`: The parameters; one entry, located by `homotopy_parameter`, is ``\lambda``.
+* `homotopy_parameter`: locator (symbol or index) of ``\lambda`` within `p`.
```
ChrisRackauckas: This requires a very specific structure for `p` which will not always be possible. Is there a reason to no just entirely separate it out? I think it should just be lambda is a separate scalar, rather than trying to treat "`p` is a constant except for one index". That also would make things like automatic differentiation really wonky.

src/problems/nonlinear_problems.jl
BatchZero: Applied, but as part of the larger λ-separation rewrite rather than via the suggestion block — the whole signature section it targeted changed. The docstring now reads `0 = f(u, p, λ)` throughout.

src/problems/nonlinear_problems.jl
BatchZero: Reworked as requested: the `homotopy_parameter` index field is deleted, and λ is now a separate trailing argument — `f(u, p, λ)` out-of-place, `f(du, u, p, λ)` in-place, i.e. the time-dependent argument convention with λ in place of `t`, with in-place autodetection via `isinplace(f, 4)`. `p` passes through unchanged, so any parameter object works (NamedTuple `p` — impossible under the old design — is now in the tests), and ∂f/∂λ is directly available for future tangent/arclength predictors.

PR #1379, opened 2026-06-09
Add better logging for singular cases in integration

Shreyas-Ekanathan:
## Checklist

## Additional context

Uses `u` and the jacobian from the integrator to give more insightful error messages on retcodes `DtLessThanMin` and `Unstable`.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -668,23 +668,15 @@ function check_error(integrator::DEIntegrator)
     # SDEIntegrator.
 
     if isnan(integrator.dt)
-        if verbose isa Bool
-            @warn "NaN dt detected. Likely a NaN value in the state, parameters, or derivative value caused this outcome."
-        else
-            @SciMLMessage("NaN dt detected. Likely a NaN value in the state, parameters, or derivative value caused this outcome.", verbose, :dt_NaN)
-        end
+        @SciMLMessage("NaN dt detected. Likely a NaN value in the state, parameters, or derivative value caused this outcome.", verbose, :dt_NaN)
         return ReturnCode.DtNaN
     end
     if integrator.iter > opts.maxiters
-        if verbose isa Bool
-            @warn "Interrupted. Larger maxiters is needed. If you are using an integrator for non-stiff ODEs or an automatic switching algorithm (the default), you may want to consider using a method for stiff equations. See the solver pages for more details (e.g. https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/#Stiff-Problems)."
-        else
-            @SciMLMessage(
-                "Interrupted. Larger maxiters is needed. If you are using an integrator for non-stiff ODEs or an automatic switching algorithm (the default), you may want to consider using a method for stiff equations. See the solver pages for more details (e.g. https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/#Stiff-Problems).",
-                verbose,
-                :max_iters
-            )
-        end
+        @SciMLMessage(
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
The Downstream CI failure on this PR ("Unitful interpolations" erroring and aborting the rest of the downstream suite, [run 27226679941](https://github.com/SciML/SciMLBase.jl/actions/runs/27226679941)) is caused by a unit bug in the new diagnostic block:

```julia
ref = max(abs(u0[i]), one(eltype(u)))
```

For a Unitful state (e.g. `u0 = 1.0u"g/L"` in `test/downstream/unitful_interpolations.jl`), `one(eltype(u))` is the dimensionless multiplicative identity, so `max(::Quantity{Float64, 𝐌 𝐋⁻³}, ::Float64)` throws:

```
DimensionError: 1.0 and 1.0 g L^-1 are not dimensionally compatible.
 [4] max(x::Unitful.Quantity{Float64, 𝐌 𝐋^-3, ...}, y::Float64)
 [5] check_error(integrator::OrdinaryDiffEqCore.ODEIntegrator{...})
   @ SciMLBase src/integrator_interface.jl:710
```

Verified locally (Julia 1.10.11, OrdinaryDiffEq v7.0.0, Unitful v1.28.0): the unitful test passes on master, fails with this branch with exactly the CI stacktrace, and passes again with the one-character-class fix:

```julia
ref = max(abs(u0[i]), oneunit(eltype(u)))
```

`oneunit` preserves the units (and is identical to `one` for plain floats), and the subsequent `abs(u[i]) > 1.0e6 * ref` comparison is then dimensionally consistent.

Two more things worth addressing while in here:

1. **This block runs on every step, not just on failure.** `check_error!` is called once per iteration of the `solve!` loop (OrdinaryDiffEqCore `src/solve.jl:764`), and the new state/Jacobian analysis sits inside `if !opts.force_dtmin && opts.adaptive` *before* any of the abort conditions are checked. So every adaptive solve now pays `findall(!isfinite, u)` + a blow-up scan + (for Radau-style caches) a fresh `calc_J` on **every accepted step**. The diagnostics should be computed lazily, only inside the branches that actually return `DtLessThanMin`/`Unstable`.
2. `round(u[i], sigdigits = 4)` in the message construction may also need care for exotic number types, but it's only reached when a diagnostic fires.

oscardssmith (approved):

src/integrator_interface.jl
```diff
@@ -702,38 +704,24 @@ function check_error(integrator::DEIntegrator)
...
-                end
-                @SciMLMessage(lazy"dt($(integrator.dt) <= dtmin($(opts.dtmin)), at t=$(integrator.t)$EEst. Aborting. There is either an error in your model specification or the true solution is unstable.", verbose, :dt_min_unstable)
+                @SciMLMessage(lazy"dt($(integrator.dt)) <= dtmin($(opts.dtmin)) at t=$(integrator.t)$EEst. Aborting. There is either an error in your model specification or the true solution is unstable.$diagnostic", verbose, :dt_min_unstable)
             end
             return ReturnCode.DtLessThanMin
         elseif !step_accepted && integrator.t isa AbstractFloat &&
                 abs(integrator.dt) <= abs(eps(integrator.t))
+            diagnostic = log_instability(integrator)
+            EEst = isdefined(integrator, :EEst) ?
+                lazy", step error estimate = $(integrator.EEst)" : ""
```
ChrisRackauckas: No reason to allocate this unless verbosity is on.

src/integrator_interface.jl
ChrisRackauckas: It's not a bool? @jClugstor

jClugstor: That's probably left over from needing backwards compatibility before.

jClugstor:
@Shreyas-Ekanathan `verbose` is now always a `DEVerbosity` so bool comparisons won't work. What you probably want to do here is check if a specific toggle is `Silent`, which you can do using `SciMLLogging.verbosity_to_bool(verbose.toggle)` where toggle is whatever verbosity toggle you want, I think in this case `dt_NaN`?

Shreyas-Ekanathan:
should be fixed now, lmk if there are any other changes needed!

PR #1386, opened 2026-06-13
Splat kwargs as NamedTuple's in remake() to avoid invalidations

JamesWrigley:
Also fixed a test error in 26ee749ff.
Written with help from Claude :robot: 

## Checklist

### DIFF
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -374,9 +374,11 @@ function remake(
     end
 
     prob = if kwargs === missing
+        # Splat as NamedTuples (not the `Pairs` directly) to keep the lowered
+        # kwarg merges off the invalidation-prone `merge(::Any, ::Pairs)` path.
         ODEProblem{iip}(
-            f, newu0, tspan, newp, prob.problem_type; prob.kwargs...,
-            _kwargs...
+            f, newu0, tspan, newp, prob.problem_type;
+            (values(prob.kwargs)::NamedTuple)..., (values(_kwargs)::NamedTuple)...
         )
     else
         ODEProblem{iip}(f, newu0, tspan, newp, prob.problem_type; kwargs...)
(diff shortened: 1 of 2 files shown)


JamesWrigley:
AFAICT all test failures are preexisting/unrelated.

ChrisRackauckas:
Was it already different types based on the kwargs before?

JamesWrigley:
Not sure I fully understand :thinking: For context this is to get around invalidations from OrderedCollections: https://github.com/SciML/NonlinearSolve.jl/pull/975

OrderedCollections defines `merge(::AbstractDict, ::AbstractDict)` methods that will invalidate these calls, so by asserting `kwargs` to be NamedTuple's we avoid the invalidations.

ChrisRackauckas:
I mean, using a namedtuple makes it so the kwargs field is different based on the arguments of the kwarg. This makes every problem a different type, which can make the usage skip the precompiled solver. Was that already the case before changing to namedtuple?

JamesWrigley:
Oh I see what you mean, I think that was already the case before since `Base.Pairs` includes the NamedTuple type in its type signature:
```julia-repl
julia> foo(; kwargs...) = typeof(kwargs)
foo (generic function with 1 method)

julia> foo()
Base.Pairs{Symbol, Union{}, Nothing, @NamedTuple{}}

julia> foo(; x=1, y="foo")
Base.Pairs{Symbol, Any, Nothing, @NamedTuple{x::Int64, y::String}}
```

So this should have the same precompilation behaviour.

JamesWrigley:
After some CI investigation:
- The Catalyst tests should be fixed by https://github.com/SciML/Catalyst.jl/pull/1495.
- Some of the MTK tests are failing because of a compat problem with DiffEqDev -> NLSove. Should be fixed by https://github.com/JuliaNLSolvers/NLsolve.jl/pull/295.

PR #1416, opened 2026-07-03
Raise Mooncake compat floor to 0.5.36 (fix Downgrade resolution)

ChrisRackauckas-Claude:
Replaces the `Mooncake = "0.4, 0.5"` union with a single floor at the latest, `Mooncake = "0.5.36"`, in every Project.toml that declares a Mooncake `[compat]` entry. Dropping the old `0.4, 0.5` union stops the downgrade min-resolution from being able to pin an ancient Mooncake — the "limit to latest Mooncake" fix for the promotion-caused Downgrade `Unsatisfiable` wall.

## Raised floors

| File | Old floor | New floor |
|---|---|---|
| `Project.toml` (root, weakdep) | `0.4, 0.5` | `0.5.36` |
| `test/downstream/Project.toml` (full dep) | `0.4, 0.5` | `0.5.36` |

Root package version patch-bumped `3.31.0 -> 3.31.1` (narrowing a weakdep `[compat]` floor is a patch-level change; no public-API change). `test/downstream/Project.toml` is an unnamed test env, so no version bump there. Only `[compat]` Mooncake lines were touched — the `[deps]`/`[weakdeps]` UUID entries are unchanged.

## Verification — resolves_at_min = TRUE (observed)

Reproduced the downgrade min-resolution with the actual `julia-actions/julia-downgrade-compat` `downgrade.jl` (current `main`, incl. #49/#50/#51 merged-test-deps + `[sources]` handling) on Julia 1.12, cloning `Resolver.jl` exactly as CI does:

- **Root `.` (the CI invocation: `projects="."`, `mode=deps`)** — `Successfully resolved minimal versions for .` (exit 0). Mooncake is a weakdep here, so it enters the manifest only as a `DifferentiationInterface` extension reference, not a version-pinned dep; the resolve succeeds regardless.
- **`test/downstream/Project.toml` (Mooncake is a full `[deps]`)** — `Successfully resolved minimal versions` (exit 0). Mooncake pins to **0.5.36** (was 0.5.25 under the old `0.4, 0.5` floor — the current registry + merged compat constraints from ModelingToolkit etc. already prevented an ancient 0.4.x even before this change).

So on the current tooling the Resolver-capacity `Unsatisfiable` wall does **not** reproduce here; the floor-raise resolves cleanly at minimum versions in both projects and pins Mooncake to the intended 0.5.36 where it is a real dependency. `resolves_at_min = true`; no dependency on #52 for this repo.

---


ChrisRackauckas-Claude:
Closing. Verified locally (ran the action's own downgrade.jl on the identical commit CI failed on) that raising the Mooncake compat floor does NOT fix the Downgrade lane. The failure is `julia-downgrade-compat@v2` merged mode promoting all ~13 weakdep AD backends (Mooncake, Zygote, ForwardDiff, DifferentiationInterface, …) to hard deps and running `Resolver.jl --min` on the joint graph — which is registry-snapshot-dependent (flips UNSAT↔SAT as those packages release) and independent of Mooncake's version. This PR also unnecessarily narrows Mooncake compat. Real fix is julia-actions/julia-downgrade-compat#52 (do not promote weakdeps in merged resolution). Superseded — closing.

PR #1420, opened 2026-07-08
Support FunctionProperties 1.1 wrapper predicates

ChrisRackauckas-Claude:

This supports the FunctionProperties 1.1 compat bump from #1418 by extending the SciMLBase weak extension to forward the new public FunctionProperties predicates through `AbstractSciMLFunction` wrappers.

Changes:
- allow `FunctionProperties = "0.1.7, 1.1"`
- forward `FunctionProperties.islinear` and `FunctionProperties.isautonomous` through `AbstractSciMLFunction.f`
- extend the FunctionProperties extension tests for wrapper methods and explicit qualified calls where exports overlap

Local verification:
- `timeout 3600 ~/.juliaup/bin/julia +1.10 --project=. -e 'using Pkg; Pkg.test(; test_args=["FunctionProperties extension"])'`\n  - observed `FunctionProperties extension | 8 pass`\n  - observed final `Testing SciMLBase tests passed`\n- `git diff --check`\n- `timeout 3600 ~/.juliaup/bin/julia +1.10 --startup-file=no -e 'using Pkg; Pkg.activate(; temp=true); Pkg.add("Runic"); using Runic; exit(Runic.main(["--check", "."]))'`\n\nNotes:\n- The `Pkg.test` invocation ran the broader SciMLBase suite despite the test argument; I let it complete and it passed.\n- This is intended as the small compatibility follow-up for #1418 rather than modifying the Dependabot branch directly.


Project.toml
```diff
@@ -92,7 +92,7 @@ DocStringExtensions = "0.9.5"
 EnumX = "1"
 Enzyme = "0.13"
 ForwardDiff = "0.10.36, 1"
-FunctionProperties = "0.1.7"
+FunctionProperties = "0.1.7, 1.1"
```
SebastianM-C, line 95: The compat extension is not correct because `islinear` was introduce in 1.x in https://github.com/SciML/FunctionProperties.jl/pull/66

The downstream tests involving SciMLSensitivity are broken https://github.com/SciML/SciMLBase.jl/actions/runs/28925741151/job/85813143959#step:11:2154 by this because SciMLSensitivity does not yet allow FunctionProperties v1

PR #1424, opened 2026-07-08
Guard FunctionProperties 1.1 predicate extensions

ChrisRackauckas-Claude:

## Summary

SciMLBase 3.32 added FunctionProperties extension methods for `islinear` and `isautonomous`, but the package still permits FunctionProperties 0.1.x for downstream compatibility. FunctionProperties 0.1.x has `hasbranching`, but does not define those newer predicate names, so the extension failed to precompile in resolver combinations such as registered SciMLSensitivity 7.113 + FunctionProperties 0.1.7.

This keeps the existing compat range and only defines the `islinear` / `isautonomous` forwarding methods when those names exist in FunctionProperties. The existing `hasbranching` forwarding remains active for the 0.1 series.

## Validation

- `timeout 3600 env GROUP=Core ~/.juliaup/bin/julia +1 --project=. -e 'using Pkg; Pkg.test(; julia_args=["--check-bounds=auto"])'`\n- `timeout 3600 ~/.juliaup/bin/julia +1 --project=/home/crackauc/sandbox/tmp_20260708_165650_11549/SciMLBase-fp01-check-env -e 'using Pkg; Pkg.develop(path="/home/crackauc/sandbox/tmp_20260708_165650_11549/SciMLBase.jl"); Pkg.add(Pkg.PackageSpec(name="FunctionProperties", version="0.1.7")); Pkg.precompile(); using SciMLBase, FunctionProperties; println("FunctionProperties.islinear defined: ", isdefined(FunctionProperties, :islinear)); println("SciMLBase/FunctionProperties 0.1.7 load succeeded")'`\n- `timeout 3600 ~/.juliaup/bin/julia +1 -e 'using Runic; exit(Runic.main(["--check", "."]))'`\n- `git diff --check`\n\nThe FunctionProperties 0.1.7 check precompiled `SciMLBaseFunctionPropertiesExt` successfully and printed `FunctionProperties.islinear defined: false`.


ChrisRackauckas-Claude:
Scratchpad update: local validation for the FunctionProperties guard branch passed against both FunctionProperties 1.1 and 0.1.7. Current red CI is blocked by separate infrastructure/base issues: QA fails on upstream master due `SciMLBase.EigenvalueTarget` being missing from the ExplicitImports EnumX allowlist (fixed in SciMLBase#1425), and several labeled downstream jobs are running the wrong/default groups because `downstream.yml@v1` does not export `inputs.group` as `GROUP` (fixed in SciML/.github#110). After those land and the `v1` workflow reference is advanced, this branch should be rerun before evaluating remaining downstream failures.

PR #1436, opened 2026-07-10
Add BasicInterpolation: runtime-switched Hermite/linear interpolation

ChrisRackauckas-Claude:
This draft PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

Adds `BasicInterpolation`, a runtime-switched fallback interpolation type, so any solver wrapper can produce solutions whose concrete type does not depend on whether the solve was dense.

Previously wrappers (e.g. Sundials.jl) chose `HermiteInterpolation` (dense) or `LinearInterpolation` (non-dense) by *type*, so the solution's concrete type changed with save arguments. Code holding a concretely-typed solution field and reassigning it across re-solves with different save settings (e.g. GaussAdjoint checkpointing) then broke. `BasicInterpolation` fixes this at the source: a single concrete type that stores `t`, `u`, `du`, a `dense::Bool`, and `sensitivitymode::Bool`, switching between Hermite (dense) and linear (non-dense) reconstruction at runtime via the `dense` field.

## Design: in-method dense branch over shared kernels (no wrapper objects)

Per review, `BasicInterpolation` does **not** construct `HermiteInterpolation`/`LinearInterpolation` objects at call time (that would be a type-unstable intermediate and a per-call allocation). Instead:

- The Hermite and linear formulas are factored into free kernel functions — `hermite_interpolant`/`hermite_interpolant!` and `linear_interpolant`/`linear_interpolant!` — with the formulas moved verbatim (no duplication, no numerical change).
- The legacy `HermiteInterpolation`/`LinearInterpolation` `interpolant`/`interpolant!` methods are now thin delegations to those kernels, so legacy behavior is bit-identical.
- The `interpolation`/`interpolation!` methods branch on `id.dense` **inside** the method for `BasicInterpolation` and call the kernels directly. Both branches return the same type (determined by the `t`/`u` containers, not the flavor), so every call is type-stable end to end. Only the dense branch reads `du`.

**Type-invariance contract:** preserving the invariance requires `du` to always be the *same container type* regardless of `dense` — pass an empty container of the dense `du` type (not `nothing`) when non-dense. The non-dense path never reads `du`. This is documented in the docstring.

## Zero allocations on the scalar in-place path

The scalar in-place interpolation path is now **0 bytes post-warmup** — for `BasicInterpolation` in both modes AND for the legacy `HermiteInterpolation` (was 96 B) and `LinearInterpolation` (was 80 B), `Val{0}` and `Val{1}`, forward and reversed time direction; confirmed by both `@allocated` behind a specialized function barrier and `Profile.Allocs` (0 recorded allocations). The pre-existing allocations came from two shared-machinery sources (found with `Profile.Allocs sample_rate=1`):

1. **64 B — dynamic dispatch at the callable boundary.** The callables took `deriv` in an unannotated slot; `deriv` is a `Type` (e.g. `Val{1}`), and Julia's Type-argument non-specialization heuristic compiled the callables with abstract `deriv`, making the inner `interpolation`/`interpolation!` call a runtime dispatch (argument box + tuple). Fixed by forcing specialization with `deriv::D ... where {D}` on all eight callables.
2. **16 B — runtime-`rev` sorted search.** `searchsortedfirst(t, tval; rev = tdir < 0)` with a runtime `Bool` makes `Base.Order.ord`'s ordering value-dependent (boxed union). Fixed with a `tdir_searchsortedfirst` helper that branches on the direction and calls `searchsortedfirst` with the concrete const `Forward`/`Reverse` ordering; also applied to the vector paths' per-iteration searches.

Through a Sundials solution (branch of SciML/Sundials.jl#547 dev'd), both the bare `interpolation!` call and the `sol(out, t)` sugar measure 0 bytes in dense and saveat modes.

## Interface hooks mirrored

- callable methods (out-of-place `(tvals, idxs, deriv, p, continuity)` and in-place `(val, tvals, idxs, deriv, p, continuity)`, scalar and vector `tvals`)
- `interp_summary`
- `enable_interpolation_sensitivitymode`
- `strip_interpolation`

## Public API / docs

- Made `public` via `@public` alongside the other interpolation types.
- Docstring documents fields, the type-invariance purpose, and the same-container-`du` requirement.
- Added to the rendered docs (`docs/src/interfaces/Solutions.md`, Interpolation Types `@docs` block).
- Additive public API → minor version bump `3.35.0` → `3.36.0`.

## Tests

New `test/interpolation_tests.jl` (registered in the core group), all passing locally on Julia 1.10 (370 assertions):

- correctness equal to `HermiteInterpolation` (dense) / `LinearInterpolation` (non-dense) on the same data — scalar and vector `tvals`, `idxs` ∈ {nothing, 1, 2, [1], [1,2]}, `deriv` ∈ {Val{0}, Val{1}}, both `continuity` values, in-place and out-of-place
- type-invariance (`typeof` identical for dense vs non-dense given same containers)
- `@inferred` for the callable in BOTH modes: scalar `tvals` with `idxs ∈ {nothing, Number, Vector}`, `Val{0}`/`Val{1}`, out-of-place and in-place — all pass exactly
- inference parity for the vector-`tvals` path: the `DiffEqArray` returned there has two `SymbolCache` metadata type parameters that are not inferred for ANY interpolation type on master (pre-existing, constructor-level; data parameters are fully inferred) — the tests pin that `BasicInterpolation`'s inferred return type is exactly equal to the legacy types' and `<: AbstractDiffEqArray`
- **allocation regression: `@allocated == 0`** on the scalar in-place path for all four (basic dense/non-dense, Hermite, linear) × `Val{0}`/`Val{1}` behind a specialized function barrier
- `interp_summary`, `strip_interpolation`, `sensitivitymode` toggling

```
Type invariance across dense/non-dense   |    3      3
interp_summary                           |    4      4
Out-of-place correctness vs delegates    |  280    280
In-place scalar correctness vs delegates |   48     48
In-place vector correctness vs delegates |    8      8
strip_interpolation                      |    2      2
Inference and allocations                |   16     16
sensitivitymode toggling                 |    9      9
```

Legacy regression checks with the branch dev'd: `solution_interface.jl` 35/35, `display.jl` 4/4, `public_interface.jl` 155/155.

## Downstream

Sundials.jl PR SciML/Sundials.jl#547 is reworked to consume this type in place of its local `SundialsInterpolation`. Merge order: this PR → SciMLBase release → 547. With this branch + the Sundials branch + SciMLSensitivity#1541 all dev'd together, the full `sundials_adjoint.jl` regression suite passes including the GaussAdjoint checkpointing testset.

https://claude.ai/code/session_01UrfZy7xBqQLgTqok5zGAdY

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.35.0"
+version = "3.36.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/docs/src/interfaces/Solutions.md
+++ b/docs/src/interfaces/Solutions.md
@@ -223,6 +223,7 @@ SciMLBase.AbstractDiffEqInterpolation
 SciMLBase.ConstantInterpolation
 SciMLBase.LinearInterpolation
 SciMLBase.HermiteInterpolation
+SciMLBase.BasicInterpolation
 SciMLBase.SensitivityInterpolation
 SciMLBase.interp_summary
 ```
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -2188,7 +2188,7 @@ export ODEAliasSpecifier, LinearAliasSpecifier
 
 # Interpolation types
 @public AbstractDiffEqInterpolation, ConstantInterpolation, LinearInterpolation,
-    HermiteInterpolation, SensitivityInterpolation
+    HermiteInterpolation, BasicInterpolation, SensitivityInterpolation
 
 # Interpolation / symbolic / solution interface
 @public interp_summary, getindepsym, getindepsym_defaultt,
--- a/src/interpolation.jl
+++ b/src/interpolation.jl
@@ -114,6 +114,69 @@ function enable_interpolation_sensitivitymode(interp::ConstantInterpolation)
 end
 
 """
+$(TYPEDEF)
+
+Runtime-switched fallback interpolation for time-series SciML solutions.
+
+`BasicInterpolation` is a single concrete type that covers both dense and
+non-dense solutions, choosing the reconstruction at *runtime* through the
+`dense` field rather than by the object's type: when `dense` is `true` it
+reconstructs values with third-order Hermite interpolation (using the stored
+derivatives `du`), and when `dense` is `false` it uses piecewise linear
+interpolation. It stores saved independent-variable values `t`, saved solution
+values `u`, and saved derivatives `du`, and it is callable through the standard
+solution interpolation interface.
+
+The point of this type is *type invariance*: the concrete type of a
+`BasicInterpolation` does not depend on whether the solve was dense. A solver
+wrapper that emits `BasicInterpolation` for both dense and non-dense solves
+produces solutions whose concrete type is identical across those save settings,
+so downstream code can hold a concretely-typed solution field and reassign it
+across re-solves with different save arguments (for example checkpointing in
+adjoint sensitivity analysis) without a type change.
+
+Preserving that invariance requires that `du` always be the **same container
+type** regardless of `dense`. When constructing a non-dense `BasicInterpolation`,
+pass an empty container of the same type that a dense solve would use (e.g. an
+empty `Vector` of the derivative element type) rather than `nothing`, so that
+`typeof` is stable across dense and non-dense solutions of the same problem. The
+non-dense path never reads `du`, so its contents are irrelevant when
+`dense = false`; only its type matters.
+
+The `sensitivitymode` flag records whether sensitivity-aware behavior has been
+enabled; as with the other interpolation types it is normally set indirectly
+through solver or sensitivity-algorithm options rather than manually.
+
+The actual numerics are delegated to [`HermiteInterpolation`](@ref) and
+[`LinearInterpolation`](@ref), so results are identical to using those types
+directly in the corresponding mode.
+"""
+struct BasicInterpolation{tType, uType, duType} <: AbstractDiffEqInterpolation
+    t::tType
+    u::uType
+    du::duType
+    dense::Bool
+    sensitivitymode::Bool
+end
+
+function BasicInterpolation(t, u, du, dense; sensitivitymode = false)
+    return BasicInterpolation(t, u, du, dense, sensitivitymode)
+end
+
+function enable_interpolation_sensitivitymode(interp::BasicInterpolation)
+    return BasicInterpolation(interp.t, interp.u, interp.du, interp.dense, true)
+end
+
+# Delegate the actual reconstruction to the Hermite/Linear interpolants so the
+# numerics are identical to those types. The temporary interpolant only wraps
+# references to the already-stored arrays (no data copy).
+@inline function as_hermite_or_linear(id::BasicInterpolation)
+    return id.dense ?
+        HermiteInterpolation(id.t, id.u, id.du; sensitivitymode = id.sensitivitymode) :
+        LinearInterpolation(id.t, id.u; sensitivitymode = id.sensitivitymode)
+end
+
+"""
     interp_summary(interp)
 
 Return a short human-readable `String` describing the interpolation used by a solution,
@@ -125,6 +188,7 @@ interp_summary(::AbstractDiffEqInterpolation) = "Unknown"
 interp_summary(::HermiteInterpolation) = "3rd order Hermite"
 interp_summary(::LinearInterpolation) = "1st order linear"
 interp_summary(::ConstantInterpolation) = "Piecewise constant interpolation"
+interp_summary(id::BasicInterpolation) = id.dense ? "3rd order Hermite" : "1st order linear"
 interp_summary(::Nothing) = "No interpolation"
 interp_summary(sol::AbstractSciMLSolution) = interp_summary(sol.interp)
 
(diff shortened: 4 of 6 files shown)


ChrisRackauckas-Claude:
## Rework: dense branch moved inside the interpolation methods (commit aaef92b)

Addressing the review point that the `id.dense ? HermiteInterpolation(...) : LinearInterpolation(...)` delegation constructed a different type per branch (type-unstable intermediate + per-call wrapper allocation). It's gone. New structure:

1. **Kernels.** The Hermite and linear formulas now live in free kernel functions — `hermite_interpolant`/`hermite_interpolant!` and `linear_interpolant`/`linear_interpolant!` — moved verbatim from the old `interpolant(Θ, id::HermiteInterpolation, ...)` / `interpolant(Θ, id::LinearInterpolation, ...)` methods (the `id` argument was dispatch-only; the bodies never read it). No formulas are duplicated.
2. **Legacy types unchanged.** `HermiteInterpolation`/`LinearInterpolation`'s `interpolant`/`interpolant!` methods are now thin `where {D}` delegations into the kernels — same numerics, same dispatch surface.
3. **`BasicInterpolation` branches in-method.** The four `interpolation`/`interpolation!` methods check `id isa BasicInterpolation` and then branch on `id.dense` at the kernel call site, calling `hermite_interpolant*` or `linear_interpolant*` directly. No wrapper object is ever constructed; only the dense branch reads `du`; both branches return the same type (fixed by the `t`/`u` containers), so calls are type-stable end to end and it does "different numerical things, not different type things."

Verification (Julia 1.10, real runs):
- `@inferred` passes exactly for the `BasicInterpolation` callable in both modes on all scalar-`tvals` paths: `idxs ∈ {nothing, Number, Vector}` × `Val{0}/Val{1}` × out-of-place/in-place.
- Vector-`tvals` calls return a `DiffEqArray` whose two `SymbolCache` metadata parameters are not inferred for ANY interpolation type on master (pre-existing, constructor-level; the data parameters are fully inferred) — verified identical for `HermiteInterpolation`/`LinearInterpolation` before the change. The tests pin inference parity: `Base.return_types` for `BasicInterpolation` equals the legacy types' exactly and is `<: AbstractDiffEqArray`.
- Allocation: the scalar in-place path allocates 80 bytes for `BasicInterpolation` in both modes vs 96 (Hermite) / 80 (linear) for the legacy types — the ~80-byte floor is shared machinery on master (e.g. `searchsortedfirst(...; rev = tdir < 0)` with a runtime `Bool`). The test asserts `BasicInterpolation <= legacy` on the identical call, which locks out per-call wrapper construction.
- All 344 pre-existing correctness assertions unchanged and passing (byte-identical numerics vs the legacy types); new totals 364/364.
- Legacy regression: `solution_interface.jl` 35/35, `display.jl` 4/4, `public_interface.jl` 155/155.
- Downstream re-verified with this branch dev'd: Sundials#547 interpolation tests (123 + `@inferred` through `sol(t)`/`sol(t, Val{1})` in dense and saveat modes), `common_interface/cvode` 33/33, `common_interface/ida` 24 pass/6 pre-existing broken, and the SciMLSensitivity#1541 `sundials_adjoint.jl` suite including `GaussAdjoint with CVODE_BDF`.

https://claude.ai/code/session_01UrfZy7xBqQLgTqok5zGAdY

ChrisRackauckas-Claude:
## Why were there any allocations? Found, fixed, now zero (commit 462a0e8)

Profiled the scalar in-place path with `Profile.Allocs.@profile sample_rate=1` — exactly 3 allocations = 80 B, two root causes, both pre-existing in the shared machinery (they hit the legacy `HermiteInterpolation`/`LinearInterpolation` identically on master):

**1. 64 B (a 16 B box + a 48 B arg tuple): dynamic dispatch at the callable boundary.** All the interpolation callables took `deriv` in an unannotated argument slot. `deriv` is a `Type` (`Val{0}`/`Val{1}`), and Julia heuristically does *not* specialize methods on `Type`-valued arguments in unannotated slots — so the callables were compiled with `deriv::Type` abstract, and the inner `interpolation!` call was a runtime dispatch boxing its arguments on every call. The profile stacktrace shows it directly: `(::BasicInterpolation)(val, tvals, idxs, deriv::Type, p, continuity)`. Fix: `deriv::D, ... ) where {D}` on all eight callables (Hermite, Linear, Constant, Basic) to force specialization.

**2. 16 B: runtime-`rev` ordering box in the sorted search.** `searchsortedfirst(t, tval; rev = tdir < 0)` passes a runtime `Bool` into `Base.Order.ord`, whose return type is then value-dependent (`ForwardOrdering` vs `ReverseOrdering` union) — one small box per call. Fix: `tdir_searchsortedfirst(t, tval, tdir)` branches on the direction and calls `searchsortedfirst` with the concrete const `Base.Order.Forward`/`Base.Order.Reverse` in each branch; also applied to the vector paths' per-iteration searches.

(Hermite's extra 16 B over the 80 was the same dispatch mechanism with a differently-sized box — same root cause, gone with the same fix.)

**Result (Julia 1.10, `@allocated` behind a specialized function barrier, post-warmup, and `Profile.Allocs` reporting 0 recorded allocations):**

| scalar in-place call | before | after |
|---|---|---|
| `BasicInterpolation` dense | 80 B | **0** |
| `BasicInterpolation` non-dense | 80 B | **0** |
| `HermiteInterpolation` (legacy) | 96 B | **0** |
| `LinearInterpolation` (legacy) | 80 B | **0** |

Zero holds for `Val{0}` and `Val{1}` and for reversed time direction. Through a Sundials solution (Sundials#547 branch dev'd), the bare `interpolation!` call **and** the `sol(out, t)` sugar both measure 0 bytes in dense and saveat modes, with `@inferred` passing.

One measurement gotcha worth noting for posterity: the benchmark harness itself must declare `deriv::D where {D}`, otherwise the same non-specialization heuristic re-boxes in the *harness* and misattributes 32–48 B to the library.

The allocation regression test is upgraded from `basic <= legacy` parity to `@allocated == 0` for all four types × both derivative orders. Numerics are untouched: all 344 correctness assertions pass unchanged (370 total with the new checks); legacy regression files (`solution_interface.jl` 35/35, `display.jl` 4/4, `public_interface.jl` 155/155) pass; downstream re-verified — Sundials#547 interpolation + cvode/ida common-interface tests, and SciMLSensitivity#1541 `sundials_adjoint.jl` 32/32 incl. `GaussAdjoint with CVODE_BDF` 4/4, all with the three branches dev'd together.

https://claude.ai/code/session_01UrfZy7xBqQLgTqok5zGAdY

ChrisRackauckas-Claude:
Follow-up PR: #1437 — the zero-allocation fixes described above were pushed to this branch after the merge snapshot, so they missed the merge window. #1437 carries them, reworked per follow-up direction: the interval search is routed through FindFirstFunctions.jl (per the SciML/OrdinaryDiffEq.jl#3826 precedent) with a `Guesser` warm start on `BasicInterpolation`'s scalar lookups, plus an AllocCheck static QA test. Scalar in-place interpolation is 0 bytes post-warmup for `BasicInterpolation` (both modes) and both legacy types.

https://claude.ai/code/session_01UrfZy7xBqQLgTqok5zGAdY

PR #1451, opened 2026-07-17
[v2-backport] Use explicit Moshi singleton clock syntax

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- spell the `ContinuousClock` and `SolverStepClock` Moshi singleton variants with the explicit `()` syntax
- raise the Moshi compatibility floor to 0.3.11, the first release supporting that syntax
- add a regression test that reloads SciMLBase from source with `--depwarn=error`

## Root cause

Moshi 0.3.11 commit `f79750c` added explicit singleton syntax and a forced deprecation for the former bare syntax. SciMLBase's v2 branch still used bare singleton declarations while allowing all Moshi 0.3 releases, so loading SciMLBase with Moshi 0.3.11 or newer fails when deprecations are errors.

This surfaced in Integrals' Julia 1.10 minimum-dependency build, which resolved SciMLBase 2.104.0 and Moshi 0.3.12.

## Impact

The next SciMLBase v2 release will load without this deprecation on both Moshi 0.3.11 and current Moshi. This does not retroactively repair older registered SciMLBase releases; downstream minimum-version jobs resolving SciMLBase 2.104.0 will still need to raise their SciMLBase floor to the fixed v2 release or receive a registry compatibility cap.

## Validation

- clean SciMLBase v2.155.1 with Moshi 0.3.12: reproduced the `ContinuousClock` deprecation failure under Julia 1.10 and `--depwarn=error`
- version boundary: SciMLBase 2.104.0 loads with Moshi 0.3.10 and fails with Moshi 0.3.11
- patched branch: source loading with compiled modules disabled and `--depwarn=error` passes with exact Moshi 0.3.11 and with Moshi 0.3.12
- `GROUP=Core` Julia 1.10 `Pkg.test()` with bounds checks and `--depwarn=error`: passed; the clock testset reports 38 passes and its pre-existing broken assertion, and Remake reports 4414 passes
- Runic 1.7.0 repository-wide check: all 115 Julia files passed
- `git diff --check`: passed


ChrisRackauckas-Claude:
Initial CI review: the spell job's two findings are unchanged lines on `v2-backport` from #1328 (`mis-treating` in `SciMLBaseChainRulesCoreExt.jl` and `SciMLBaseMooncakeExt.jl`). Neither file is changed by this PR. The remaining checks are still running; the branch-local Runic and Core tests pass locally as documented in the PR body.

ChrisRackauckas-Claude:
CI follow-up: the `StochasticDelayDiffEq.jl/All/1.10` failure was reproduced independently on comment-only #1452, confirming it is a clean `v2-backport` downstream failure unrelated to this Moshi clock-syntax patch. It precompiles current StochasticDelayDiffEq against incompatible current StochasticDiffEq APIs and errors at `handle_callback_modifiers!`. A separate clean-base investigation is queued.

The pre-existing Spell Check failure is isolated in #1452; this focused PR intentionally does not absorb that unrelated diff.

ChrisRackauckas-Claude:
CI exposed a second unrelated downstream failure in `ModelingToolkitStandardLibrary/Core`. It is identical on #1451 and #1452 and reproduces independently of either PR's changed lines.

The failing cases in `test/Core/rotational.jl` are `two inertias`, `two inertias with driving torque`, and `sensors` (7 pass / 3 error). ModelingToolkit initialization reports:

```
Expected an Initial parameter to exist for variable Differential(t, 2)(inertia2₊phi(t)), but did not find one.
```

This is being tracked as a separate clean-base downstream investigation. No workaround or unrelated fix will be added to this PR.

ChrisRackauckas-Claude:
The final `Catalyst/All` downstream job exposed another unrelated v2-base compatibility failure. It is identical on #1451 and #1452.

The primary failure is JumpProcesses precompilation at `src/SSA_stepper.jl:139`:

```
UndefVarError: derivative_discontinuity! not defined in SciMLBase
```

Because JumpProcesses cannot load, Catalyst's graph extensions also fail to load; the eventual Graph visualization result is 3 pass / 1 error with no applicable `plot_network(::ReactionSystem)` method. That latter error is a consequence of the extension-load failure, not evidence for changing either PR.

This is queued as a separate clean-v2-base downstream investigation. No unrelated compatibility workaround will be added here.

ChrisRackauckas:
Integrals.jl should bump its SciMLBase minimum to v3.

PR #1458, opened 2026-07-22
Add AutoDePSpecialize specialization level

ChrisRackauckas-Claude:
> [!NOTE]
> Draft PR — please ignore until reviewed by @ChrisRackauckas.

Adds `AutoDePSpecialize <: AbstractSpecialization`, a specialization level that extends `AutoSpecialize` by also **de-specializing the parameter object**: solver paths with an opaque-parameter strategy pack an isbits, non-`NullParameters` `p` into a fixed-type opaque container (e.g. `RespecializeParams.OpaqueParams`) and install wrappers whose signatures carry the container type in the `p` slot. The concretized problem type — and the solver compilation — becomes independent of the user's parameter struct type, so one precompiled solve covers all isbits parameter types, while the user's `f` still receives `p` fully typed via an allocation-free unpack.

SciMLBase only defines the **marker**: type + docstring, `@public`, `@docs` entry in the Specialization Levels section of `interfaces/Problems.md`, and `"AutoDePSpecialize"` added to the `SpecializationLevel` preference allowlist. No new dependencies — the wrapping is solver-owned, so any function family (ODE, nonlinear, …) can implement its own strategy.

Placing the level here (rather than in DiffEqBase, where an earlier revision defined it) is per @ChrisRackauckas's direction so that **NonlinearSolve.jl can implement the same level** for its residual wrappers. The ODE-stack implementation and benchmarks live in SciML/OrdinaryDiffEq.jl#3692 (measured there: first-solve TTFX for a never-seen parameter struct 12.1 s → 0.43 s for `DefaultODEAlgorithm`, runtime within ~2% of `AutoSpecialize`).

Construction through the generic `AbstractSciMLFunction` constructor branches already works for arbitrary specialization types, so no constructor changes are needed; the marker alone performs no wrapping in SciMLBase (fields stay concretely typed, `p` untouched), which the new tests in `test/problem_building_test.jl` assert, along with `specialization()` round-tripping and `unwrapped_f` preserving the marker.

**Version**: 3.36.1 → 3.37.0 (new public API).

**Tests run locally**: `test/problem_building_test.jl` with dev'd SciMLBase — all pass including the 8 new assertions (`AutoDePSpecialize specialization marker | 8 pass`). Runic applied to changed .jl files.

https://claude.ai/code/session_01UuAbA8hSnrirMrJjFqxwn5

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.36.1"
+version = "3.37.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/docs/src/interfaces/Problems.md
+++ b/docs/src/interfaces/Problems.md
@@ -55,6 +55,7 @@ Note that `iip` choice is required for specialization choices to be made.
 ```@docs
 SciMLBase.AbstractSpecialization
 SciMLBase.AutoSpecialize
+SciMLBase.AutoDePSpecialize
 SciMLBase.NoSpecialize
 SciMLBase.FunctionWrapperSpecialize
 SciMLBase.FullSpecialize
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -2230,7 +2230,7 @@ export ODEAliasSpecifier, LinearAliasSpecifier
 @public DECache
 
 # Specialization markers
-@public FullSpecialize, NoSpecialize, FunctionWrapperSpecialize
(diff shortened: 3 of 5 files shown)


ChrisRackauckas-Claude:
Added `should_opaque_p(p)` alongside the `AutoDePSpecialize` marker — the default policy for whether the level de-specializes a parameter (`isbits(p) && !(p isa NullParameters)`). It lives here because it references `NullParameters` and both consuming stacks (DiffEqBase, NonlinearSolve) dep SciMLBase, so they share one policy instead of each inlining the check. `@public` + documented (Problems.md `@docs`) + tested (`problem_building_test.jl`, 6 assertions). The mechanism it gates — containers, `OpaqueVoid`, the `wrap_void_opaque` installer — lives in RespecializeParams (SciML/RespecializeParams.jl#23).

PR #1462, opened 2026-07-23
Fix master QA: parameterless_type docstring + reexport allowlist

ChrisRackauckas-Claude:
**This PR should be ignored until reviewed by @ChrisRackauckas.**

## Summary

Master `Run Tests` / QA fails with two SciMLTesting checks:

1. **`public API has docstrings`** — `parameterless_type` was `@public` and listed in docs, but the docstring was attached to the internal `Base.@pure __parameterless_type` helper instead of the public methods.
2. **`No unapproved public reexports`** — intentional reexports (`@reexport using SciMLOperators`, CommonSolve verbs, `AllObserved`, Base `deleteat!` as the DEIntegrator verb) were not listed in `reexports_allow`.

## Changes

- Move the `parameterless_type` docstring onto the public methods in `src/utils.jl`.
- Pass `reexports_allow = _reexports_allow` in `test/qa/qa.jl` (shared with `rendered_ignore`, plus `:step!`, `:AllObserved`, `:deleteat!`).
- Add a mixed `Symbol`/`Int` all-states `save_idxs` regression next to the #1454 coverage (adjacent path to pure-symbolic all-vars).

## Local verification

```
GROUP=QA julia --project=. -e 'using Pkg; Pkg.test()'
# QA | 24 passed / 24 total
# Testing SciMLBase tests passed
```

## Related master CI notes

IntegrationTest / ReleaseTest failures on master are largely **downstream** of this package (not fixed here):

- OrdinaryDiffEq `DEVerbosity` resize MethodError → tracked by SciML/OrdinaryDiffEq.jl#3971
- Catalyst `UndefKeywordError: parent_sys` on ReleaseTest → Catalyst-side

PR #1464, opened 2026-07-23
Fix CheckInit MethodError when abstol is a vector (ODE #1214)

### LINKED ISSUE #1214: DABDF2: Initialization fails if abstol is a vector
When simulating a DAE problem (with DifferentialEquations v6.15.0) with integrator DABDF2 the following error occurs:
```
ERROR: LoadError: MethodError: no method matching isless(::Float64, ::Array{Float64,1})
Closest candidates are:
  isless(::Float64, ::Float64) at float.jl:465
  isless(::Missing, ::Any) at missing.jl:87
  isless(::PyCall.PyObject, ::Any) at D:\otter\home\.julia\packages\PyCall\zqDXB\src\pyoperators.jl:75
  ...
Stacktrace:
 [1] <(::Float64, ::Array{Float64,1}) at .\operators.jl:268
 [2] <=(::Float64, ::Array{Float64,1}) at .\operators.jl:317
 [3] _initialize_dae!(<..>) at ...\OrdinaryDiffEq\NsugH\src\initialize_dae.jl:193
```
The code at this place is:
```
 	integrator.opts.internalnorm(resid,t) <= integrator.opts.abstol && return
```
In my case, `abstol` is a vector and not a scalar and this seems the reason why this statement fails.

If this statement is changed to
```
integrator.opts.internalnorm(resid,t) <= maximum(integrator.opts.abstol) && return
```
the issue should be fixed.

(Note, in my applications, I do not want that the DAE is initialized by DifferentialEquations.jl, but could not yet figure out how to switch automatic initialization off, see SciML/DifferentialEquations.jl#549; if initialization could be switched off, the issue above would no longer occur).

ChrisRackauckas-Claude:
## Summary

- Fixes `MethodError: no method matching isless(::Vector{Float64}, ::Float64)` in `CheckInit` when `abstol` is a vector.
- Root cause: `normresid > abstol` is invalid when `abstol` is an array (default DAE path uses `CheckInit` via `DefaultInit`).
- Fix: weighted residual check for array `abstol` (`internalnorm(resid ./ abstol, t) > 1`), matching `check_dae_tolerance` in OrdinaryDiffEqNonlinearSolve.
- Adds regression test for vector `abstol` on `DAEProblem`.

Fixes SciML/OrdinaryDiffEq.jl#1214 (default initialization path).

## Test plan

- [x] Local: vector abstol + consistent `DAEProblem` with DABDF2/DFBDF/DImplicitEuler succeeds
- [x] Local: vector abstol still throws `CheckInitFailureError` when residual is large
- [ ] CI: `test/initialization.jl`

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -219,6 +219,20 @@ _vec(v::Number) = v
 _vec(v::SciMLOperators.AbstractSciMLScalarOperator) = v
 _vec(v::AbstractVector) = v
 
+# Check residual against abstol. Scalar abstol uses the usual scalar residual
+# norm; array abstol uses a weighted residual (componentwise / abstol) so vector
+# tolerances do not MethodError on `normresid > abstol` (OrdinaryDiffEq #1214).
+@inline function exceeds_checkinit_abstol(resid, abstol::Number, integrator, t)
+    normresid = isdefined(integrator.opts, :internalnorm) ?
+        integrator.opts.internalnorm(resid, t) : norm(resid)
+    return normresid > abstol
+end
+@inline function exceeds_checkinit_abstol(resid, abstol, integrator, t)
+    weighted = isdefined(integrator.opts, :internalnorm) ?
+        integrator.opts.internalnorm(resid ./ abstol, t) : norm(resid ./ abstol)
+    return weighted > 1
+end
+
 """
     $(TYPEDSIGNATURES)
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
## CI babysit note

### PR fix pushed

`test/initialization.jl` is **not** included by `runtests.jl` (only `test/downstream/initialization.jl` is, in the Downstream group). The original vector-abstol tests would never have run in CI.

Pushed `b0ab390` to wire the OrdinaryDiffEq #1214 / vector-abstol regression into `test/downstream/initialization.jl`.

### Downstream reds (not PR-caused)

Same master-side pattern as #1462/#1463: Catalyst `parent_sys`, DiffEqDevTools QA, `step_limiter` → #1466, DEVerbosity, etc.

Local unit tests for `exceeds_checkinit_abstol` (scalar + vector abstol) pass.

ChrisRackauckas-Claude:
@JuliaTagBot tag

ChrisRackauckas-Claude:
@JuliaRegistrator register

PR #1446, opened 2026-07-15
Propagate release downstream resolver failures

ChrisRackauckas-Claude:
## Summary

- remove the `ResolverError` catch from the release downstream job
- let dependency-resolution and Catalyst test failures propagate to GitHub Actions
- retain the existing local SciMLBase development and Catalyst test sequence

This is stacked on #1445; the extra concurrency-context commit will disappear from this diff when that PR merges.

Ignore this PR until reviewed by @ChrisRackauckas.

## Local verification

- `/home/crackauc/.juliaup/bin/julialauncher +1.12 --startup-file=no -m Runic --check .`
- reproduced the workflow's downstream environment locally with the branch version of SciMLBase and ran `Pkg.test("Catalyst"; coverage=true)`
- the command ran Catalyst's real test suite and exited 1 at the already-investigated `UndefKeywordError: keyword argument parent_sys not assigned` failure (20 passed, 1 errored in `Symbolic Stoichiometry`), rather than converting it to success

The Catalyst failure is pre-existing on the unmodified base and is tracked separately; this PR intentionally exposes it.

PR #1472, opened 2026-07-27
Make strict QA owner-contract clean

ChrisRackauckas-Claude:

This is a SciMLBase 4.0.0 public-contract cleanup. It removes blanket facade reexports, retains only narrow non-public compatibility bindings to public owner APIs for released solvers, and documents the versioned developer extension surface.

Highlights:
- `postamble!` is direct-definition documented, developer-public, rendered in the integrator interface, and covered by a generic dispatch contract test.
- Removes strict-QA issues from nonpublic imports, stale imports, reexports, and EnumX-generated source analysis.
- Replaces private static-array and Logging hooks with public-owner APIs.
- Enables doctests and bare `run_qa(SciMLBase)` with no repository exceptions.

Validation:
- `Pkg.test()` passed (including `Remake 4430/4430`).
- Bare strict QA against pending SciMLTesting v2.6: 20/20; AllocCheck 6/6.
- Runic and `git diff --check` passed.
- Docs/doctests passed before the v4 metadata transition. The final v4 docs resolver is correctly blocked by released OrdinaryDiffEq 7.2 capping SciMLBase at 6.6.6; this PR needs the accompanying OrdinaryDiffEq compatibility stack before its docs CI can resolve.

Depends on https://github.com/SciML/SciMLTesting.jl/pull/36 for generated-enum ExplicitImports handling.


src/utils.jl
```diff
@@ -422,9 +422,14 @@ macro def(name, definition)
...
-using Base: typename
+function _unwrap_unionall(T)
+    while T isa UnionAll
+        T = T.body
+    end
+    return T
+end
 
-Base.@pure __parameterless_type(T) = typename(T).wrapper
+__parameterless_type(T) = T.name.wrapper
```
ChrisRackauckas: This should just be the withoutargs? It needs to be compile time, and it's that same generated function

ChrisRackauckas (changes requested):

src/ensemble/basic_ensemble_solve.jl
```diff
@@ -588,12 +570,7 @@ function solve_batch(
 end
 
 function tmap(f, args...)
-    batch_data = Vector{
-        Core.Compiler.return_type(f, Tuple{typeof.(getindex.(args, 1))...}),
-    }(
-        undef,
-        length(args[1])
-    )
+    batch_data = Vector{Any}(undef, length(args[1]))
```
ChrisRackauckas: No just use `Core.Compiler.return_type`, set it as allowed.

src/ensemble/basic_ensemble_solve.jl
```diff
@@ -524,14 +513,7 @@ function solve_batch(
...
 function responsible_map(f, II...)
-    batch_data = Vector{
-        Core.Compiler.return_type(
-            f, Tuple{ntuple(i -> typeof(II[i][1]), Val(length(II)))...}
-        ),
-    }(
-        undef,
-        length(II[1])
-    )
+    batch_data = Vector{Any}(undef, length(II[1]))
```
ChrisRackauckas: No just use Core.Compiler.return_type, set it as allowed.

src/solutions/basic_solutions.jl
```diff
@@ -182,7 +182,7 @@ Return `sol` or wrap it in a higher-level SciML solution container.
 
 Solvers call `wrap_sol(sol)` after constructing a low-level solution. When
 `sol.prob` is an [`AbstractSciMLProblem`](@ref), the default implementation
-queries [`problem_type`](https://docs.sciml.ai/SciMLBase/stable/interfaces/Problem_Traits/) and dispatches to
+queries [`problem_type`](@ref) and dispatches to
```
ChrisRackauckas: No ref's here, since they won't work downstream for example when in diffeqdocs.

src/debug.jl
```diff
@@ -80,22 +80,5 @@ function _has_sciml_in_stacktrace()
...
-    Base.Experimental.register_error_hint(MethodError) do io, e, args, kwargs
-        if e isa MethodError && NullParameters in args
-            println(io, NO_PARAMETERS_ARITHMETIC_ERROR_MESSAGE)
-        end
-    end
-
-    return Base.Experimental.register_error_hint(FunctionWrappersWrappers.NoFunctionWrapperFoundError) do io,
-            e
-        println(io, FUNCTIONWRAPPERSWRAPPERS_MSG)
-    end
```
ChrisRackauckas, line 100: Keep and allow this.

src/solutions/solution_interface.jl
```diff
@@ -1,14 +1,3 @@
-### Abstract Interface
-"""
-    AllObserved
-
-Selector requesting every observed variable available through a solution's
-symbolic indexing interface. `AllObserved` is reexported from
-RecursiveArrayTools so SciML solution code can use the same selector without
-depending on its storage location.
-"""
-const AllObserved = RecursiveArrayTools.AllObserved
```
ChrisRackauckas, line 10: Why is this removed?

src/integrator_interface.jl
```diff
@@ -1110,7 +1121,6 @@ function Base.iterate(integrator::DEIntegrator, state = 0)
 end
 
 Base.eltype(::Type{T}) where {T <: DEIntegrator} = T
-Base.IteratorSize(::Type{<:DEIntegrator}) = Base.SizeUnknown()
```
ChrisRackauckas, line 1113: Why removed?

src/SciMLBase.jl
```diff
@@ -139,7 +135,7 @@ are:
   - `tspan`: the independent-variable interval for time-dependent problems.
   - `p`: parameters, defaulting to `NullParameters` when omitted.
   - `kwargs`: keyword arguments stored on the problem and forwarded to solves.
-  - construction-layout metadata available through [`problem_type`](https://docs.sciml.ai/SciMLBase/stable/interfaces/Problem_Traits/) when
+  - construction-layout metadata available through [`problem_type`](@ref) when
```
ChrisRackauckas: more ref issues.

src/solve.jl
```diff
@@ -38,7 +38,7 @@ promote_u0(::Nothing, p, t0) = nothing
 isdualtype(::Type{T}) where {T} = false
 
 has_kwargs(_prob::AbstractSciMLProblem) = has_kwargs(typeof(_prob))
-Base.@pure __has_kwargs(::Type{T}) where {T} = :kwargs ∈ fieldnames(T)
+__has_kwargs(::Type{T}) where {T} = :kwargs ∈ fieldnames(T)
```
ChrisRackauckas: does this infer?

src/utils.jl
```diff
@@ -330,7 +330,7 @@ function isinplace(
             # Find if there's a `f(args...)` dispatch
             # If so, no error
             _parameters = if methods(f).ms[1].sig isa UnionAll
-                Base.unwrap_unionall(methods(f).ms[1].sig).parameters
+                _unwrap_unionall(methods(f).ms[1].sig).parameters
```
ChrisRackauckas: does this infer?

ChrisRackauckas-Claude: Reverted to `Base.unwrap_unionall`. The hand-rolled loop inferred the same (both return `Any`), but there is no reason to carry a copy: SciMLTesting's import checks are owner-based, and `Base` owns `unwrap_unionall`, so it needs no exception. Same for `num_types_in_tuple(::UnionAll)`.

src/utils.jl
ChrisRackauckas-Claude: Done — `__parameterless_type` is now a generated function like `_without_args`.

Dropping `Base.@pure` here was worse than a missed fold: the replacement `T.name.wrapper` throws on a `UnionAll`, so `parameterless_type(ODEProblem)` errored outright. Generating from `Base.typename` handles both. Verified on 1.10/1.11/1.12/1.13-rc1 that `parameterless_type(p) === ODEProblem ? 1 : 1.0` infers `Int`, and `remaker_of` infers `Type{ODEProblem{true}}`.

src/solve.jl
ChrisRackauckas-Claude: It folded via effects analysis, but not by construction, so `__has_kwargs` is a generated function now to guarantee it. `has_kwargs(p) ? 1 : 1.0` infers `Int` on 1.10 through 1.13-rc1. Covered in `test/inference_tests.jl`.

src/ensemble/basic_ensemble_solve.jl
ChrisRackauckas-Claude: Restored. No allowance needed in the end — the qualified-access check is owner-based and `Core` owns `Core.Compiler.return_type`, so bare `run_qa` accepts it. `test/inference_tests.jl` asserts `responsible_map` returns `Vector{Int}` rather than `Vector{Any}`.

src/ensemble/basic_ensemble_solve.jl
ChrisRackauckas-Claude: Restored in `tmap` as well, same reasoning; asserted in `test/inference_tests.jl`.

src/debug.jl
ChrisRackauckas-Claude: Restored all three `register_error_hint` calls. No allowance needed (`Base` owns `Base.Experimental`).

`test/error_hint_tests.jl` did not catch the removal because it only asserted the base error text, never the hint body. It now asserts the hint contents, and I confirmed it fails when `__init__` is reverted.

src/solutions/basic_solutions.jl
ChrisRackauckas-Claude: Reverted to the full URL, along with the other 15 `problem_type` `@ref`s across `src/`.

src/SciMLBase.jl
ChrisRackauckas-Claude: Same — reverted to the docs URL.

src/solutions/solution_interface.jl
ChrisRackauckas-Claude: No good reason; it was collateral from dropping the reexport surface. Restored the const, the export, the docstring, and the `@docs` entry in `Symbolic_SaveIdxs.md`, plus the `public_interface.jl` binding assertion. This is the one name that still needs an exception: `run_qa(SciMLBase; reexports_allow = (:AllObserved,))`.

src/integrator_interface.jl
ChrisRackauckas-Claude: Same — collateral. Restored, and `test/inference_tests.jl` now asserts `Base.IteratorSize(SciMLBase.DEIntegrator) === Base.SizeUnknown()`.

ChrisRackauckas-Claude:
Addressed the review in 1221342f7. Summary of what changed and how it was checked.

**Restored as requested**

| Comment | Change |
| --- | --- |
| `Core.Compiler.return_type` (x2) | Back in `responsible_map` and `tmap`; ensemble batches allocate a concrete element type again instead of `Vector{Any}`. |
| Error hints | All three `register_error_hint` registrations back in `__init__`. |
| `AllObserved` | Const, export, docstring, `@docs` entry, and the `public_interface.jl` binding assertion restored. |
| `Base.IteratorSize` | `Base.SizeUnknown()` for `DEIntegrator` restored. |
| `problem_type` refs | All 16 `@ref`s reverted to the full docs URL. |
| `Base.unwrap_unionall` | Hand-rolled unwrapping loop removed. |

None of these needed a QA exception. SciMLTesting's import checks are owner-based, so `Base` and `Core` accesses pass unlisted. The only allowance left is `reexports_allow = (:AllObserved,)`.

**Compile-time traits**

`__parameterless_type` and `__has_kwargs` had lost `Base.@pure` and become ordinary functions. Both are generated functions now, matching `_without_args`, so they fold by construction rather than by effects analysis.

The `__parameterless_type` rewrite to `T.name.wrapper` was also a plain bug, not just a missed fold — a `UnionAll` has no `name` field:

```
[2] __parameterless_type(T::Type) @ SciMLBase src/utils.jl:432
[3] parameterless_type(::Type{ODEProblem}) @ SciMLBase src/utils.jl:443
```

so `parameterless_type(ODEProblem)` threw on the previous commit. Generating from `Base.typename` handles concrete types and `UnionAll`s alike.

**Also reverted, not flagged in review but the same class of regression**

- `Base.Experimental.@max_methods 1` at the top of `SciMLBase.jl`.
- `Base.promote_typejoin` -> `typejoin` in `SavedSubsystem`. Not equivalent: `typejoin` widens small unions to an abstract supertype where `promote_typejoin` keeps the `Union`.
- The `StaticArray` `__sum` specialization in `solve.jl`.

Happy to drop any of these three if you would rather keep them out of this PR.

**Verification**

- Full `Pkg.test()` on 1.10: passed, including `Remake 4430/4430`.
- `run_qa(SciMLBase; reexports_allow = (:AllObserved,))`: 18/18 against SciMLTesting master (needs SciML/SciMLTesting.jl#36, merged today but not yet registered; on the released 2.5.0 the generated-enum modules still error).
- `test/inference_tests.jl` (new, in `Core`): asserts `has_kwargs` and `parameterless_type` fold to `Int` at a branch site, `remaker_of` infers `Type{ODEProblem{true}}`, `responsible_map`/`tmap` return `Vector{Int}`, `parameterless_type` accepts `UnionAll`s, and `IteratorSize` is `SizeUnknown`. Green on 1.10, 1.11, 1.12, and 1.13-rc1.
- `test/error_hint_tests.jl` asserted only base error text, so it passed with the hints deleted. It now asserts the hint bodies; confirmed it fails when `__init__` is reverted.
- Runic and `git diff --check` clean.

Still ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
Diagnosed the CI failures. One was mine; the rest are the expected consequences of a breaking version bump, plus a stale pin in this repo.

**1. QA (Julia 1) — my regression, fixed in 1cf9c38bf**

`all_qualified_accesses_are_public` flagged all eleven names I restored: `@max_methods`, `Experimental`, `Compiler`/`return_type`, `typename`, `unwrap_unionall`, `promote_typejoin`, `SizeUnknown`, `register_error_hint`, `NoFunctionWrapperFoundError`, `StaticArray`.

My "no allowances needed" claim in the previous comment was wrong, and the reason is worth recording: SciMLTesting skips the two `_are_public` checks on Julia < 1.11, because `Base.ispublic` does not exist there. I verified only on the 1.10 LTS, so those two checks never ran and I read 18/18 as a full pass. The `_via_owners` checks do pass unlisted; the `_are_public` ones do not. So `set it as allowed` was right, and the ignore list is back.

Verified with the real harness on both QA lanes:

- `GROUP=QA` on Julia 1.12: 35/35
- `GROUP=QA` on Julia 1.10: 33/33 (two fewer, the checks skipped on LTS)

**2. Core — green**

Passes on lts, 1, and pre. No downstream-independent failures.

**3. Downstream / DownstreamAD / Python / SymbolicIndexingInterface — stale pin in this repo, now widened**

All four failed with `empty intersection between SciMLBase@4.0.0 and project compatibility 2 - 3`. `test/downstream/Project.toml` and `test/python/Project.toml` still capped `SciMLBase` at `"2, 3"`, so they could not resolve against the 4.0.0 dev version at all. Widened both to `"2, 3, 4"`. This does not turn the lanes green: they now fail on the real constraint, which is that ModelingToolkit and OrdinaryDiffEq still cap SciMLBase at 3. That is the compatibility stack the PR description already flags.

**4. Documentation — blocked on the same stack**

`Unsatisfiable requirements`: `docs/Project.toml` allows SciMLBase 4, but ModelingToolkit 10-11 caps it at 3.

**5. ReleaseTest — not a code failure**

This went green on a91749780 and red on 1221342f7 without any change from me. The cause is upstream: 7b806c50a ("Propagate release resolver failures", #1446) landed on master between the two runs and removed the `catch err; err isa Pkg.Resolve.ResolverError || rethrow()` that used to log `Not compatible with this release. No problem.` and `exit(0)`. Since `pull_request` runs use the merged workflow, my run picked up the stricter version. Both runs hit the identical Catalyst/MTK resolver error; only the handling changed. So a breaking bump now fails this job by design, and it will stay red until Catalyst allows SciMLBase 4.

**Also in this commit**

SciMLTesting 2.6.0 is registered now, so the compat entry moves from `"2.4"` to `"2.6"`. The generated-enum handling this qa.jl depends on is not in 2.5, and the dependency note in the PR description is resolved.

Still ignore until reviewed by @ChrisRackauckas.

PR #1474, opened 2026-07-28
Make has_sys public API

ChrisRackauckas-Claude:
> **Ignore this PR until it has been reviewed by @ChrisRackauckas.**

## Summary

`has_sys` is the only member of the `AbstractSciMLFunction` field-existence trait family that is neither public nor documented. Its siblings — `has_jac`, `has_jvp`, `has_vjp`, `has_tgrad`, `has_analytic`, `has_initialization_data` — are all in the `@public` block and all render in `docs/src/interfaces/SciMLFunctions.md`.

Concretely, NonlinearSolveBase's `AutoDePSpecialize` path needs it: it declines parameter de-specialization for problems carrying symbolic metadata, since `p` must stay concrete for symbolic indexing and the initialization pipeline. Today that access trips `ExplicitImports`' non-public-qualified-access check.

This PR:

- adds a docstring to `has_sys` at its definition site
- adds it to the `@public` trait block
- renders it in the SciMLFunctions trait docs
- adds `has_sys` coverage to `test/existence_functions.jl`, including the no-`sys`-field short-circuit
- bumps to 3.42.0 (new public API is a minor bump)

## Rebase note

This PR originally also made `ImmutableNonlinearProblem` public. That half has been **dropped from this PR** because #1482 landed on master first and already does it — declares it `@public`, gives it a docstring, and adds the `@docs` entry next to `NonlinearProblem`. Nothing is lost; only the `has_sys` half remains here.

The version bump also moved: the original target 3.41.0 was taken by #1482 and is registered in General, so this is now 3.42.0.

## Validation

Run locally on the rebased branch (master @ 79ed05e):

```
$ julia --project=. test/existence_functions.jl
Test Summary: | Pass  Total  Time
has_sys       |    3      3  0.2s

$ GROUP=QA julia --project=. -e 'using Pkg; Pkg.test()'
Test Summary: | Pass  Total     Time
QA            |   51     51  2m57.7s
```

The QA group runs `api_docs_kwargs = (; rendered = true, ...)`, so it is what verifies the new public name actually carries a docstring and appears in the rendered docs.

Runic `--check` on the changed source files exits 0.

Follow-up: SciML/NonlinearSolve.jl will consume this to drop its internal `has_sys` access.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.41.0"
+version = "3.42.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/docs/src/interfaces/SciMLFunctions.md
+++ b/docs/src/interfaces/SciMLFunctions.md
@@ -133,6 +133,7 @@ SciMLBase.has_jvp
 SciMLBase.has_vjp
 SciMLBase.has_tgrad
 SciMLBase.has_initialization_data
+SciMLBase.has_sys
 ```
 
 ## AbstractSciMLFunction API
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -2112,7 +2112,7 @@ export ODEAliasSpecifier, LinearAliasSpecifier
 
 # SciMLFunction derivative traits
 @public has_jac, has_jvp, has_vjp, has_tgrad, has_analytic, has_reinit,
-    has_initialization_data, has_stats
+    has_initialization_data, has_stats, has_sys
 
 # Function-argument validation errors
 @public FunctionArgumentsError, TooFewArgumentsError, TooManyArgumentsError
--- a/src/scimlfunctions.jl
+++ b/src/scimlfunctions.jl
@@ -5591,6 +5591,17 @@ has_Wfact(f::AbstractSciMLFunction) = __has_Wfact(f) && f.Wfact !== nothing
 has_Wfact_t(f::AbstractSciMLFunction) = __has_Wfact_t(f) && f.Wfact_t !== nothing
 has_paramjac(f::AbstractSciMLFunction) = __has_paramjac(f) && f.paramjac !== nothing
 has_vjp_p(f::AbstractSciMLFunction) = __has_vjp_p(f) && f.vjp_p !== nothing
+"""
+    has_sys(f::AbstractSciMLFunction)
+
+Return whether `f` carries a non-`nothing` symbolic system.
+
+The system is the symbolic description the function was generated from, such as
+a ModelingToolkit system or a `SymbolicIndexingInterface.SymbolCache`. It backs
+symbolic indexing and the initialization pipeline, so code that must keep a
+problem's `u0`/`p` in a form those paths can read should query this trait rather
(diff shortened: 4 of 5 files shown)


ChrisRackauckas-Claude:
Rebased onto current `master` (`e0733ef`, picking up #1472 / #1475 / #1477). No conflicts, and #1475's QA changes do not touch either of these two names.

Re-verified after the rebase:

```
$ GROUP=QA julia --project=. -e 'using Pkg; Pkg.test()'
QA            |   35     35  1m47.2s

$ julia --project=. -e 'include("test/existence_functions.jl")'
has_sys       |    3      3  0.2s
existence_functions OK
```

Runic 1.7.0 `--check` on the four changed source files exits 0.

**Note on why this PR previously showed no checks at all:** the workflows only ever ran against my first commit (`2ed1c64`); nothing was triggered for the second one, so the status rollup was empty. The rebase re-triggered the full matrix.

**Note on the failures that did appear on `2ed1c64`** — Documentation, IntegrationTest and ReleaseTest — all three fail on `master` too, so none are caused by this PR:

| workflow | last 3 runs on `master` |
|---|---|
| Documentation | failure, failure, failure |
| IntegrationTest | failure, failure, failure |
| ReleaseTest | failure, failure, failure |

The Documentation one is a `linkcheck` 403 against `https://docs.sciml.ai/SciMLBase/stable/interfaces/Problem_Traits/`, a URL that several pre-existing docstrings already reference. `Run Tests` passed.

On versioning: `master` is at 3.40.0 and the registry's newest is 3.39.1, so 3.40.0 is itself unreleased. I have kept the bump to 3.41.0 rather than folding into 3.40.0 deliberately — if 3.40.0 were registered from a commit predating this one, a downstream `SciMLBase = "3.40"` floor would resolve to a version without `has_sys`. Consuming PR SciML/NonlinearSolve.jl#1124 floors on `"3.41"` for that reason.

ChrisRackauckas-Claude:
CI update after the rebase. Everything that has finished passes except one downstream job, and that one is not caused by this PR.

Passing so far: `Downgrade` (7m32s), `Runic Format Check`, `Runic Suggestions`, `Spell Check`. The rest of the matrix is still queued.

**`OrdinaryDiffEq.jl/Core / Downstream Tests - Core` — fails, inherited from `master`.**

The error is in DiffEqBase's test suite:

```
Plot Vars: Error During Test at SafeTestsets.jl:30
  LoadError: UndefVarError: `solve` not defined
  in expression starting at downstream/lib/DiffEqBase/test/plot_vars.jl:18
Test Summary: | Error  Total  Time
Plot Vars     |     1      1  1.0s
```

`solve` is undefined because `master` no longer exports it. b9c3507 (#1472, "Make strict QA owner-contract clean") dropped these two lines from `src/SciMLBase.jl`:

```diff
-export solve, solve!, init, discretize, symbolic_discretize
-export step!, deleteat!, addat!, get_tmp_cache,
```

`master` still does `import CommonSolve: solve, init, step!`, but importing is not exporting, so downstream code doing `using SciMLBase` loses the verbs.

This PR does not touch any of that — `git diff origin/master -- src/SciMLBase.jl` contains no line mentioning `solve`. This branch simply sits on top of `e0733ef`, which includes b9c3507, and inherits the breakage.

#1476 is the fix-forward and restores exactly those exports:

```diff
+export solve, solve!, init, step!, discretize, symbolic_discretize
```

So this job should go green here once #1476 lands, with no change needed on this branch.

PR #1476, opened 2026-07-28
Export the CommonSolve verbs again, and drop the undocumented reexports

ChrisRackauckas-Claude:

Stacked on https://github.com/SciML/SciMLBase.jl/pull/1475 — review that one first; this branch contains its commit.

## What this keeps

#1472 removed `solve`, `solve!`, `init`, and `step!` from the export surface. Those four are the SciML solve interface as users and solver packages write it, and they are documented here, so this exports them again and allow-lists them in QA rather than dropping them.

## What stays removed

Everything else #1472 dropped stays dropped, because none of it is documented by SciMLBase. Checked rather than assumed:

- **The SciMLOperators reexport** (`MatrixOperator`, `FunctionOperator`, `ScalarOperator`, `IdentityOperator`, `NullOperator`, `AffineOperator`, `DiagonalOperator`, `TensorProductOperator`, `WOperator`, `AddVector`, `BlockDiagonalOperator`, `InvertibleOperator`, `StaticWOperator`, `TensorSumOperator`, `cache_operator`, `concretize`, `isconstant`, `iscached`, `islinear`, `issquare`, `isconvertible`, `kronsum`, `has_adjoint`, `has_concretization`, `has_exp`, `has_expmv`(`!`), `has_ldiv`(`!`), `has_mul`(`!`), `update_coefficients`(`!`), `SciMLOperators`) — zero `@docs` entries anywhere in `docs/src`. These are documented at SciMLOperators, which owns them.

- **`deleteat!`** — documented only as a method on the Base function, `Base.deleteat!(::SciMLBase.DEIntegrator, ::Any)` in `Init_Solve.md`. The export was redundant: `deleteat!` resolves through Base in any scope, so dropping it changes nothing for callers. Verified both halves — in a `using SciMLBase` module `deleteat!` still resolves to `Base.deleteat!`, and SciMLBase still contributes its `DEIntegrator` method.

- **`AllObserved`** — this one *was* documented, at `Symbolic_SaveIdxs.md`. But it is a const alias for the RecursiveArrayTools selector, so the docstring and rendered entry were documenting a name SciMLBase does not own. The const, its docstring, the `@docs` entry, and the export all go.

## Net effect on the export surface

Measured with `names(SciMLBase)` against both baselines:

- vs. current master: `init`, `solve`, `solve!`, `step!` restored; `AllObserved` removed.
- vs. pre-#1472 (3.40.0): 36 names still removed, being the 34 SciMLOperators names plus `AllObserved` and `deleteat!`.

So this is still a reduction in the export surface relative to 3.40.0, and `AllObserved` is the one name that genuinely leaves user scope. Worth a deliberate call on whether 3.41.0 is the right version number for that, since the removals now stand on "not documented here" rather than on SemVer.

## Verification

- `GROUP=QA` on Julia 1.12: 35/35.
- Runic clean.
- Full `Pkg.test()` running.

Also carries the `SciMLTesting` compat move to `"2.6"` from the parent commit.

https://claude.ai/code/session_01DFVmGLPUsWbRgkLJckxUGy

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,4 @@
 [deps]
-CommonSolve = "38540f10-b2f7-11e9-35d8-d573e4eb0ff2"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
@@ -8,7 +7,6 @@ SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
 
 [compat]
-CommonSolve = "0.2"
 Documenter = "1"
 ModelingToolkit = "10, 11.0"
 OrdinaryDiffEq = "6, 7.0"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,4 +1,4 @@
-using CommonSolve, Documenter, SciMLBase
+using Documenter, SciMLBase
 
 cp("./docs/Manifest.toml", "./docs/src/assets/Manifest.toml", force = true)
 cp("./docs/Project.toml", "./docs/src/assets/Project.toml", force = true)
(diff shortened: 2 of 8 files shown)


ChrisRackauckas-Claude:
Independent corroboration that this is needed, from an unrelated PR's CI.

#1474 is rebased on `e0733ef` (which includes b9c3507) and touches nothing related to exports — its `src/SciMLBase.jl` diff contains no line mentioning `solve`. Its `OrdinaryDiffEq.jl/Core` downstream job nonetheless fails:

```
Plot Vars: Error During Test at SafeTestsets.jl:30
  LoadError: UndefVarError: `solve` not defined
  in expression starting at downstream/lib/DiffEqBase/test/plot_vars.jl:18
```

So the dropped `export solve, solve!, init, …` from #1472 is breaking downstream `using SciMLBase` consumers in CI right now, not just in principle. This PR's `+export solve, solve!, init, step!, discretize, symbolic_discretize` is exactly what those consumers need.

PR #1484, opened 2026-08-02
Fix downstream rendering of integrator docstrings

ChrisRackauckas-Claude:
Ignore this PR until reviewed by @ChrisRackauckas.

## Summary

- replace local `@ref` links in the `last_step_failed`, `check_error`, and `check_error!` docstrings with links to the owner package documentation
- add a regression test that checks these downstream-rendered docstrings contain owner-documentation links and no local `@ref` targets

## Why

OrdinaryDiffEq renders these SciMLBase docstrings in its own manual, but it does not render the corresponding SciMLBase bindings as local targets. Documenter therefore cannot resolve the local `check_error` reference and terminates during cross-reference validation.

The first failing OrdinaryDiffEq docs environment used SciMLBase 3.40.0; the preceding successful environment used 3.39.1. The link was introduced by SciMLBase #1472. This follows the same downstream-rendering pattern already fixed for another SciMLBase docstring in #1469.

Failing job: https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/30746151318/job/91491989362

## Local validation

- `Runic --check --diff .`
- `GROUP=Core julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`
- exact OrdinaryDiffEq documentation build at `a7080c6`, with this SciMLBase checkout developed into its docs environment

The first two commands passed, and the downstream OrdinaryDiffEq build completed with exit code 0. SciMLBase's own documentation passed cross-reference checks locally, then its full build was blocked by an unrelated `Problem_Traits` linkcheck returning HTTP 403; that same linkcheck failure reproduces on an untouched upstream-master worktree.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -933,8 +933,9 @@ end
 Return whether the preceding attempted solver step failed to converge.
 
 Concrete differential-equation integrators may specialize this hook when their
-step controller tracks a recoverable failed attempt. [`check_error`](@ref) uses
-it to convert a non-adaptive repeated failure into
+step controller tracks a recoverable failed attempt.
+[`check_error`](https://docs.sciml.ai/SciMLBase/stable/interfaces/Init_Solve/#SciMLBase.check_error)
+uses it to convert a non-adaptive repeated failure into
 `ReturnCode.ConvergenceFailure`. The default is `false`.
 
 !!! warning "Developer API, not user API"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Investigation scratchpad:

1. Reproduced the OrdinaryDiffEq docs failure in a detached checkout at `a7080c6` with the same SciMLBase 3.41.0 / Documenter 1.17.0 environment as Actions.
2. Compared Actions environments: SciMLBase 3.39.1 passed; 3.40.0 first showed this exact cross-reference failure.
3. Mapped the 3.40.0 tree and blamed the new link to `b9c3507a` (#1472). The owner docstring was being rendered downstream with a target that existed only in the SciMLBase manual.
4. Applied the owner-documentation-link pattern already used by #1469, covered all three mutually linked integrator hooks, and added the regression test.
5. Verified whole-repository Runic, the Core test group, and the complete OrdinaryDiffEq docs build against commit `185512d15`.

PR #1487, opened 2026-08-02
Publish linear and eigenvalue solution construction hooks

ChrisRackauckas-Claude:
## Summary
- mark `build_linear_solution` and `build_eigenvalue_solution` as versioned SciMLBase developer APIs
- document both hooks at their definition sites with SciMLStyle arguments, keywords, returns, and doctested examples
- render the hooks in the Developer API and verify a downstream-style module imports only the public construction surface

## Verification
- `Pkg.test()`
- docs build with doctests
- Runic and `git diff --check`

### DIFF
--- a/docs/src/interfaces/Developer_API.md
+++ b/docs/src/interfaces/Developer_API.md
@@ -27,6 +27,17 @@ SciMLBase.value
 SciMLBase.unitfulvalue
 ```
 
+## Solution Construction Hooks
+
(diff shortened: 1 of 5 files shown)


ChrisRackauckas-Claude:
CI update: the `GROUP=Downstream` resolver failure reproduces unchanged on clean current `master` at `b6e131a37` before any downstream test executes. It is tracked separately in #1488; this PR only publishes/document existing solution-construction hooks.

PR #1492, opened 2026-08-03
Publish solver solution construction hooks

ChrisRackauckas-Claude:
## Summary
- make the LinearSolve solution-construction hooks source-owned public developer API
- document the extension contract and add a generic interface test

## Validation
- `julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`\n- Runic non-mutating check on changed Julia files\n\nIgnore until reviewed by @ChrisRackauckas.


ChrisRackauckas-Claude:
Superseded by the already-merged #1487, which contains the same public solution-construction API and interface test.

PR #1496, opened 2026-08-05
Keep SciML function trait docs at definition sites

ChrisRackauckas-Claude:
## Summary
- remove duplicate `@doc` propagation for SciML function traits
- retain the SciMLStyle docstrings immediately above each defining method
- assert the affected trait bindings remain public and documented

## Validation
- `test/existence_functions.jl` (`26/26` passed)
- Runic on touched files

### DIFF
--- a/test/existence_functions.jl
+++ b/test/existence_functions.jl
@@ -35,6 +35,18 @@ end
     @test !has_observed(missing)
 end
 
+@testset "Documented function trait interface" begin
+    if isdefined(Base, :ispublic)
+        for name in (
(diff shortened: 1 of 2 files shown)


test/existence_functions.jl
```diff
@@ -35,6 +35,18 @@ end
...
+        for name in (
+                :has_analytic, :has_jac, :has_jvp, :has_vjp, :has_tgrad,
+                :has_initialization_data,
+            )
+            @test Base.ispublic(SciMLBase, name)
+            @test Base.Docs.hasdoc(SciMLBase, name)
+        end
+    end
+end
+
```
ChrisRackauckas:
```suggestion
```

PR #1504, opened 2026-08-07
Move error logging from SciMLBase to DiffEqBase

Shreyas-Ekanathan:
## Checklist

## Additional context

A follow-up PR will be opened to OrdinaryDiffEq for the second half of this work.


src/integrator_interface.jl
```diff
@@ -900,6 +900,25 @@ has_mtk_sys(integrator) = false
...
+`check_error` only detects failures; describing them belongs to the solver stack,
+which owns both the wording and the verbosity settings that gate it. DiffEqBase.jl
+implements this for `DEIntegrator`s. The default is a no-op, so a stack that does not
+implement it still gets correct return codes, just no diagnostics.
+
+Implementations must not affect control flow and their return value is ignored. A
+`reason` configured at `SciMLLogging.ErrorLevel` throws instead of returning, which is
+the intended behaviour of that level.
+"""
+@inline report_integrator_failure(integrator, ::Val) = nothing
```
ChrisRackauckas, line 920: Who consumes this that isn't DiffEqBase?

src/integrator_interface.jl
```diff
@@ -962,26 +981,25 @@ mutate `integrator.sol.retcode`; use
 [`check_error!`](https://docs.sciml.ai/SciMLBase/stable/interfaces/Init_Solve/#SciMLBase.check_error!)
 when the solution must be updated. Concrete integrators may specialize the
 checks while preserving the return-code contract.
+
+Diagnostics for detected failures are emitted through
+[`report_integrator_failure`](@ref), which the solver stack implements; this
+function only performs detection.
 """
 function check_error(integrator::DEIntegrator)
```
ChrisRackauckas, line 989: Who consumes this that isn't DiffEqBase?

Shreyas-Ekanathan: So the answer is that its only DiffEqBase, but when I dropped check_error entirely from SciMLBase, there didn't seem to be a clean way to merge everything in one go without breaking one of DiffEqBase or SciMLBase (or a relevant dep), so the best solution I could think of was a more gradual approach, so we leave this small bit for now, and gradually migrate everything to make sure that we do this right.

src/SciMLBase.jl
```diff
@@ -2367,4 +2365,7 @@ export ODEAliasSpecifier, LinearAliasSpecifier
 # Problem alias specifiers
 @public RODEAliasSpecifier, SDEAliasSpecifier
 
+# Integrator diagnostic hooks
+@public report_integrator_failure
```
ChrisRackauckas: why would this be public if it's being removed?

ChrisRackauckas:
```suggestion
```

PR #1521, opened 2026-08-13
Raise SciMLTesting floor to 2.8

ChrisRackauckas-Claude:


ChrisRackauckas-Claude:
Closing: current SciMLBase floors are already 2.6 (root) and 2.4 (QA), so this is an unnecessary rebump under the fleet policy of requiring SciMLTesting >=2.4. The remaining work targets only package entries below 2.4.

PR #419, opened 2023-03-15
make types safe to print when verbose stacktraces are on

KristofferC:
The first commit invokes the show method for general `Type`s in order for it to be safe. So at least people that hit https://github.com/SciML/TruncatedStacktraces.jl/issues/18 can now disable it and get working code:

```julia
julia> using SnoopCompileCore; invalidations = @snoopr(using SciMLBase); using SnoopCompile


julia> trees = invalidation_trees(invalidations)
3-element Vector{SnoopCompile.MethodInvalidations}:
Error showing value of type Vector{SnoopCompile.MethodInvalidations}:
ERROR: UndefVarError: `T` not defined
Stacktrace:
  [1] show(io::IOContext{IOBuffer}, t::
SYSTEM (REPL): showing an error caused an error
ERROR: UndefVarError: `T` not defined
Stacktrace:
  [1] show(io::IOContext{IOBuffer}, t::
SYSTEM (REPL): caught exception of type UndefVarError while trying to handle a nested exception; giving up

julia> SciMLBase.TruncatedStacktraces.VERBOSE[] = true
true

julia> trees = invalidation_trees(invalidations)
3-element Vector{SnoopCompile.MethodInvalidations}:
 inserting print_matrix_row(io::IO, X::Union{FillArrays.AbstractFill{<:Any, 1}, FillArrays.AbstractFill{<:Any, 2}, FillArrays.RectDiagonal, LinearAlgebra.AbstractTriangular{<:Any, <:FillArrays.AbstractFill{<:Any, 2}}, LinearAlgebra.Diagonal{<:Any, <:FillArrays.AbstractFill{<:Any, 1}}}, A::Vector, i::Integer, cols::AbstractVector, sep::AbstractString, idxlast::Integer) @ FillArrays ~/.julia/packages/FillArrays/xqile/src/FillArrays.jl:688 invalidated:
   backedges: 1: superseding print_matrix_row(io::IO, X::AbstractVecOrMat, A::Vector, i::Integer, cols::AbstractVector, sep::AbstractString, idxlast::Integer) @ Base arrayshow.jl:97 with MethodInstance for Base.print_matrix_row(::IOContext{Base.TTY}, ::AbstractVecOrMat, ::Vector{Tuple{Int64, Int64}}, ::Int64, ::Vector{Int64}, ::String, ::Int64) (13 children)
   8 mt_cache

[...]

PR #1550, opened 2026-08-23
Restore the SciMLOperators operator interface on `using SciMLBase`

ChrisRackauckas:
## What broke

`b9c3507a` ("Make strict QA owner-contract clean", #1472) removed
`@reexport using SciMLOperators` from `src/SciMLBase.jl`.

Reexport exports `filter(x -> Base.isexported(m, x), names(m; all = true, imported = true))`,
so that one line had been putting **SciMLOperators' 33 exported names plus the
`SciMLOperators` module binding** into scope for every `using SciMLBase`. Deleting it
removed all 34 from SciMLBase's public surface. Nothing in SciMLBase's own test suite
imports them, so CI stayed green while `using SciMLBase; MatrixOperator(A)` began
failing with `UndefVarError` — for the whole ecosystem, since SciMLBase is the root of
the SciML dependency tree.

Measured: `length(names(SciMLBase))` is **466 on master**, **498 on this branch**.

This is the same class of breakage already confirmed elsewhere in the org — JumpProcesses
stripping its DiffEqBase reexport broke MomentClosure's `DiscreteProblem`/`EnsembleProblem`,
and NeuralPDE stripping `SciMLBase, ModelingToolkit` broke NeuralLyapunov.

## Evidence used to pick the list

1. **SciMLBase's own documentation.** `docs/src/interfaces/SciMLFunctions.md` documents
   `jac_prototype` as "an `AbstractSciMLOperator`" whose Jacobian update is built "using
   `update_coefficients!` for the in-place form and `update_coefficients` for the
   out-of-place form", and links to SciMLOperators' Premade Operators page.
   `SplitFunction`, `SplitODEProblem` and `IncrementingODEProblem` are specified as
   operator-valued. That documented use cannot work if `using SciMLBase` does not supply
   the operators.
2. **Reconstruction of the exact lost surface.** `SciMLOperators.jl` at `bb5437a`
   (2026-07-28, the state at the time of #1472) exported exactly 33 names, and had no
   `@public` declarations at all. Those 33 plus the module binding are what the reexport
   was providing.
3. **Downstream corroboration.** `OptimizationBase` records the same surface from the
   other side: its source comment says blanket-reexporting SciMLBase "put 256 names into
   every downstream namespace — `DynamicalSDEFunction`, `BlockDiagonalOperator`,
   `AddVector` and the rest of SciMLOperators". `DifferentialEquations.jl`'s
   `reexports_allow` list still carries this same operator set as approved public surface.

## Restored names (32)

`SciMLOperators` (the module binding), plus:

**Premade operator types (12)** — `AddVector`, `AffineOperator`, `BlockDiagonalOperator`,
`DiagonalOperator`, `FunctionOperator`, `IdentityOperator`, `InvertibleOperator`,
`MatrixOperator`, `NullOperator`, `ScalarOperator`, `TensorProductOperator`,
`TensorSumOperator`

**Operator interface and traits (19)** — `cache_operator`, `concretize`, `has_adjoint`,
`has_concretization`, `has_exp`, `has_expmv`, `has_expmv!`, `has_ldiv`, `has_ldiv!`,
`has_mul`, `has_mul!`, `isconstant`, `isconvertible`, `iscached`, `islinear`, `issquare`,
`kronsum`, `update_coefficients`, `update_coefficients!`

## Deliberately not restored

- **`WOperator`, `StaticWOperator`** — the stiff-solver `W = M - γJ` machinery. Grepping
  every SciML clone I have locally, the only live uses are in OrdinaryDiffEq and
  LinearSolve, both of which depend on SciMLOperators directly; everywhere else they
  appear only in comments. A `using SciMLBase` user never constructs one. This mirrors
  Sundials dropping problem types for equation classes it cannot solve.
- **`jacobian_stale`, `mark_jacobian_updated!`, `mark_jacobian_current!`** — SciMLOperators
  only began exporting these in its #419, *after* #1472, so they were never part of this
  surface.
- **`AbstractSciMLOperator`** — worth flagging, because it is the name most often assumed
  to have been lost here. It was not: SciMLOperators has never exported it, and its
  `@public` declaration landed in #418, also after #1472 — and Reexport filters on
  `Base.isexported` regardless of `public`. It stays reachable the documented way, as
  `SciMLOperators.AbstractSciMLOperator`, which is precisely why the `SciMLOperators`
  binding itself is back in the export list.

## Documenting the reexported API

New rendered page **`docs/src/interfaces/Reexported_API.md`**, wired into
`docs/pages.jl`, listing every name `using SciMLBase` provides that SciMLBase does not
define — so a reader can see at a glance what they get and which package owns each name.

- The 32 SciMLOperators names grouped **by role**: the `SciMLOperators` module binding;
  operator types (trivial / matrix-backed / scalar / matrix-free / structured and lazy);
  operator interface and traits (updating / caching and materialization / structural
  traits / supported-operation traits / lazy algebra).
- Stated explicitly as **owned and documented by
  [SciMLOperators.jl](https://docs.sciml.ai/SciMLOperators/stable/)**, linking to its
  [Premade Operators](https://docs.sciml.ai/SciMLOperators/stable/premade_operators/) and
  [interface](https://docs.sciml.ai/SciMLOperators/stable/interface/) pages — SciMLBase
  only re-exports them.
- A **boundary** section: *anything else from SciMLOperators must be imported from
  SciMLOperators directly*, with both deliberate exclusions spelled out so they do not
  read as oversights — `AbstractSciMLOperator` and the other public-but-unexported
  abstract/lazy-algebra types (reachable qualified, which is what the re-exported
  `SciMLOperators` binding is for), and the `WOperator`/`StaticWOperator` +
  `jacobian_stale`/`mark_jacobian_updated!`/`mark_jacobian_current!` W-matrix machinery.
- The CommonSolve verbs (`solve`, `solve!`, `init`, `step!`) already exported here get
  the same treatment on the same page.
- `SciMLFunctions.md`, which already documents `jac_prototype` in terms of
  `AbstractSciMLOperator` and `update_coefficients!`, now points at the new page.

The list is now in three places — the docs page, the `export` block in
`src/SciMLBase.jl`, and `_scimloperators_reexports` in `test/qa/qa.jl`. I verified
programmatically that **all three parse to the identical 32-name set**, each code copy
cross-references the other two, and the "Reexport surface" testset holds the code copies
to the real export surface.

## Implementation notes

- `const isconstant = SciMLOperators.isconstant` becomes an ordinary
  `using SciMLOperators: isconstant`, matching what `solve!` already does with CommonSolve.
  The alias made SciMLBase the binding owner, which would have failed `run_api_docs`'
  docstring check once the name was exported; imported, the docstring resolves to
  SciMLOperators.
- `test/qa/qa.jl` declares the list through `reexports_allow`, and adds a "Reexport
  surface" testset asserting every approved name is both in `names(SciMLBase)` and
  actually in scope from a bare `using SciMLBase`. That check runs in its own
  single-`using` module so the extension trigger packages `qa.jl` loads cannot mask a
  missing export. The allow-list cannot drift from the exports in either direction.

## Compat floor: SciMLOperators 1.3 → 1.19

**This is a genuinely tighter floor than SciMLBase carries today, and it is a direct
consequence of the restoration — not unrelated churn.** Downgrade CI caught it:

```
ERROR: LoadError: UndefVarError: `BlockDiagonalOperator` not defined
ERROR: LoadError: Failed to precompile SciMLBase
```

The job resolved SciMLOperators to the declared floor, 1.3.0, where the new
`using SciMLOperators: ...` line cannot even load. Until now nothing in SciMLBase
referenced these names, so the floor was never tested against them.

Walking every registered 1.x tag of SciMLOperators and extracting its `export` blocks
gives the introduction point of each restored name:

| introduced in | names |
|---|---|
| 1.0.0 | the other 27 |
| 1.13.0 | `has_concretization` |
| 1.19.0 | `BlockDiagonalOperator`, `TensorSumOperator`, `kronsum` |

So **1.19.0** is the first registered version exporting all 32, and every release from
1.19.0 through 1.28.0 (and master's 1.29.0) still does. 1.19.0 is not yanked, and its
registered `git-tree-sha1` matches the `v1.19.0` tag. 1.18.0 is missing three of the
names, so the floor is *minimal*, not merely sufficient.

No other compat entry needs moving. Julia floors match (both 1.10), and SciMLOperators
1.19.0's own bounds are already satisfied by SciMLBase's — `ArrayInterface` 7.19 exactly,
`Adapt`/`DocStringExtensions`/`StaticArraysCore` above — except `Accessors`, where
SciMLOperators 1.19.0 wants ≥ 0.1.42 and SciMLBase declares ≥ 0.1.39. That needs no change
here: `julia-downgrade-compat@v2` resolves a jointly minimal version set (Resolver.jl
`--min-resolve`) rather than pinning each direct dependency, so it lifts Accessors to
satisfy the transitive bound, and SciMLBase's own 0.1.39 floor stays accurate for its own
use.

## Why this is not breaking

Every name here used to be exported by this package; this only restores what #1472
removed, and adds nothing that was not previously in scope. Per SemVer that is a minor
(here: patch, 0.x-style) change. **No version bump in this PR** — a separate release pass
handles versions.

## What I verified vs. inspected

**Ran locally** (Julia 1.11.8, macOS):

- `Pkg.instantiate()` + `using SciMLBase` on this branch — loads clean.
- `length(names(SciMLBase))`: 466 on master → 498 on this branch, i.e. exactly the 32
  intended names and nothing else.
- All 32 names present in `names(SciMLBase)` **and** `isdefined` in a module whose only
  `using` is SciMLBase.
- Functional smoke test: `MatrixOperator(rand(3,3))`, `isconstant`, and
  `update_coefficients` round-trip through the reexported names.
- `SciMLTesting.public_reexports(SciMLBase; allow = <new list>)` → `Symbol[]`, i.e. the
  allow-list is exact, with no unapproved reexports left over.
- `SciMLTesting.run_api_docs(SciMLBase)` → 2/2 pass (docstring and rendered-docs checks
  both green; reexported names are exempt from local rendering and inherit SciMLOperators'
  docstrings).
- `ExplicitImports.improper_explicit_imports` stale-import check → no stale imports
  introduced by the new `using SciMLOperators: ...` block.
- A scoped `Documenter.makedocs` build over the new page plus the two pages it links to:
  the page renders, and all three new cross-references resolve to real anchors
  (`Reexported_API → SciMLFunctions#scimlfunctions`,
  `Reexported_API → Init_Solve#The-SciML-init-and-solve-Functions`,
  `SciMLFunctions → Reexported_API#Reexported-API`). All three upstream SciMLOperators
  doc URLs return HTTP 200 (`linkcheck = true` is on in this repo's `make.jl`).
- A script parsing all three copies of the list (docs page, `export` block, QA tuple) and
  asserting set equality → identical, 32 names each.
- **The compat floor**, two ways: statically, by extracting the `export` blocks of every
  registered SciMLOperators 1.x tag to find each name's introduction point (1.19.0 is the
  max; 1.18.0 is missing three); and dynamically, by resolving SciMLOperators to exactly
  1.19.0 against this branch — SciMLBase precompiles, `names(SciMLBase)` is 498, all 32
  names in scope, `MatrixOperator`/`isconstant` smoke test passes. Also confirmed 1.19.0
  is not yanked in the General registry and its registered tree hash matches the tag.
- `Runic --check --diff` on the changed `.jl` and `.md` files → clean.

**Not run**: the full `Pkg.test("SciMLBase")` suite, the complete `run_qa` entry point, and
the full `docs/make.jl` build. The QA environment pulls Enzyme, Makie, Mooncake, PythonCall
and the rest of the weakdep closure, and the docs environment pulls ModelingToolkit,
OrdinaryDiffEq, StochasticDiffEq and Plots with `doctest = true, linkcheck = true`; this
machine is heavily loaded by other work. Instead I exercised the three `run_qa` sub-checks
this change can actually affect (`public_reexports`, `run_api_docs`, ExplicitImports'
stale-import check) directly against the built package, and did the scoped Documenter build
described above. CI covers the rest.

https://claude.ai/code/session_01Rmh6B9eoW3N8oCbuuVPVxJ


ChrisRackauckas-Claude:
OpenAI Codex continuation (2026-08-25): the earlier “Not run” limitation in the PR body is superseded by full local validation.

- `GROUP=Core ~/.juliaup/bin/julia +release --project=. -e 'using Pkg; Pkg.test()'` → `Testing SciMLBase tests passed` (including Remake 4430/4430).
- `GROUP=QA JULIA_CONDAPKG_BACKEND=Null JULIA_PYTHONCALL_EXE=/usr/bin/python3 ... Pkg.test()` → QA 123/123 and `Testing SciMLBase tests passed`.
- `~/.juliaup/bin/julia +release --project=docs docs/make.jl` → exit 0 after doctests, cross-references, checks, and HTML rendering.
- Runic and typos over the complete parent diff → exit 0.
- Exact floor probe: SciMLOperators 1.18.0 lacks `BlockDiagonalOperator`, `TensorSumOperator`, and `kronsum`; 1.19.0 has all three.

The PR’s Documentation failure is the incorrect `JuliaLang/CommonSolve.jl` owner URL. The exact one-line correction is available in stacked draft https://github.com/SciML/SciMLBase.jl/pull/1555; the corrected URL returns HTTP 200 and the full docs build passes with it.

The `DownstreamAD` failures reproduce unchanged on master in https://github.com/SciML/SciMLBase.jl/actions/runs/32635237444 (Julia 1: https://github.com/SciML/SciMLBase.jl/actions/runs/32635237444/job/97183904080; LTS: https://github.com/SciML/SciMLBase.jl/actions/runs/32635237444/job/97183904057).

AI continuation: OpenAI Codex session ID 01a03a11-47c2-77e0-858b-167004f0b79c.

PR #1553, opened 2026-08-25
Add explicit ODE function respecialization

### LINKED ISSUE #1551: ODEFunction kwarg constructors cannot produce the natively-built AutoSpecialize parameterization
## Summary

A natively-constructed `AutoSpecialize` problem (e.g.
`ODEProblem(sys, [], tspan)` on an MTK system, where `AutoSpecialize` is the
default) carries an `ODEFunction` whose last two type parameters are **boxed**:

- `initialization_data :: Union{Nothing, SciMLBase.OverrideInitData}`
- `nlstep_data :: Union{Nothing, SciMLBase.ODENLStepData}`

But the documented way to (re)build a function at a given specialization level,

```julia
ODEFunction{iip, SciMLBase.AutoSpecialize}(f.f;
    initialization_data = f.initialization_data, sys = f.sys, ...)
```

always uses `typeof(initialization_data)` / `typeof(nlstep_data)` for those
parameters (`scimlfunctions.jl`, the generic `else` branch of the kwarg
constructor; only `NoSpecialize` widens). The result is an
**`AutoSpecialize`-labeled function with the `FullSpecialize` layout** — the
concrete `OverrideInitData{…}` aggregate inlined in the type. There is no
public API that produces the boxed parameterization for an existing function;
it only arises through whole-problem construction. The same pattern is
present in `unwrapped_f` and in `ODEInputFunction`.

Confirmed still present as of `SciMLBase` master `d3a30e5d` (2026-08-23,
v3.49.2).

## Why it matters (measured consequences)

The boxed vs inline layout is behaviorally significant under Enzyme
(measured on a DAE calibration-style workload with an SCC initialization
system, Julia 1.12.5, SciMLBase 3.34.0, Enzyme 0.13.181; MTK 11.31.1):

- **Compile time:** first Enzyme gradient of a `remake`+`solve` loss: 358.8 s
  with the boxed parameterization vs **1272.0 s with the label-only rebuild**
  (and 1222.9 s for plain `FullSpecialize`) — identical gradients and results.
  Enzyme's Type Analysis walks the inline aggregate; the `Union` boxing keeps
  it behind a pointer.
- **Correctness-adjacent:** the inline layout is the enabling ingredient for an
  order-dependent `EnzymeNoTypeError` in the dynamic `remake` thunk
  (`maybe_eager_initialize_problem`): a trivially-initialized problem with the
  inline layout fails to differentiate when any other Enzyme solve-gradient was
  compiled earlier in the same process; the boxed layout is immune. See
  SciML/SciMLSensitivity.jl#1583 for the full Enzyme-level reproducer of that
  half.

So "respecialize via the kwarg constructor" silently keeps the vulnerable/slow
layout while claiming `AutoSpecialize` in the type — a footgun for any
downstream code that converts specialization levels after construction.

## Reproducer (type-level, no Enzyme needed)

Verified 2026-08-24 against registry SciMLBase 3.49.2 / ModelingToolkit
11.39.1 / Julia 1.12.7:

```julia
using ModelingToolkit, SciMLBase
using ModelingToolkit: D_nounits as D, t_nounits as t

@parameters m=1.5 d=9.0
@variables s(t) v(t)
sys = mtkcompile(System([D(s) ~ v, m * D(v) ~ 1 - d * v], t; name = :model,
    initialization_eqs = [s ~ 0, v ~ 0]))

native = ODEProblem(sys, [], (0.0, 1.0))            # AutoSpecialize by default
f = native.f
rebuilt = ODEFunction{true, SciMLBase.AutoSpecialize}(f.f;
    mass_matrix = f.mass_matrix, observed = f.observed, sys = f.sys,
    initialization_data = f.initialization_data)

ps(T) = collect(Base.unwrap_unionall(T).parameters)
ps(typeof(f))[end-1]        # Union{Nothing, SciMLBase.OverrideInitData}
ps(typeof(rebuilt))[end-1]  # SciMLBase.OverrideInitData{…} (concrete!)
```

## Current workaround

Explicit type surgery + inner constructor. Verified: `remake(prob; f = newf)`
preserves the widened parameters, so either `remake` or a Setfield
`@set prob.f = newf` works for the problem-level swap:

```julia
params = collect(Base.unwrap_unionall(typeof(f)).parameters)
params[2] = SciMLBase.AutoSpecialize
params[end-1] = Union{Nothing, SciMLBase.OverrideInitData}   # initialization_data
params[end]   = Union{Nothing, SciMLBase.ODENLStepData}      # nlstep_data
newf = ODEFunction{params...}(ntuple(i -> getfield(f, i), fieldcount(typeof(f)))...)
```

This depends on the type-parameter order, i.e. it is version-coupled.

## Ask

One of:
1. Make the kwarg constructors honor the specialization level for
   `initialization_data`/`nlstep_data` (widen for `AutoSpecialize` the way the
   `NoSpecialize` branch already widens everything), or
2. provide an official `respecialize(f, AutoSpecialize)` /
   `SciMLBase.remake(f; specialization = AutoSpecialize)` API, or
3. at minimum document that kwarg-constructor "respecialization" does not
   reproduce the native `AutoSpecialize` parameterization.

Also worth deciding deliberately: where the boxing SHOULD happen. I could not
find the widening in ModelingToolkitBase or in the `ODEFunction` constructor
branches, so it presumably comes from the whole-problem construction path —
making it explicit would prevent the two layouts from drifting further apart.

## Related ask: does this also undermine `AutoDespecialize`/`AutoRespecialize`?

SciMLBase recently added `AutoDespecialize` and `AutoRespecialize` (née
`AutoDePSpecialize`), whose whole point is to make a single compiled solve
reusable across differently-typed `p` objects by boxing `p` into a
stable-outer-type container. But the kwarg constructor's `else` branch — the
one that mishandles `AutoSpecialize` above — is the *same* branch these two
markers fall into; there is no dedicated case for either in
`scimlfunctions.jl`. That means `initialization_data`/`nlstep_data` still get
inlined via `typeof(initdata)`/`typeof(nlstep_data)` under
`AutoDespecialize`/`AutoRespecialize` too.

If `OverrideInitData`'s concrete type varies across the "differently-typed"
problems these markers are meant to unify (e.g. different
`initializeprob`/`update_initializeprob!` closures per system), the
`ODEFunction` type — and therefore the compiled solve — still differs
per-instance, silently defeating the compilation-reuse guarantee
`AutoRespecialize`/`AutoDespecialize` advertise. Worth clarifying whether the
fix for `AutoSpecialize` (ask #1 above) should also cover these two markers,
or whether they need their own explicit widening branch, since their
documented value proposition depends on it more directly than
`AutoSpecialize`'s does.

(Concrete evidence this isn't hypothetical: `rebuilt`'s printed type in the
reproducer above includes a nested `NonlinearFunction{true,
SciMLBase.AutoDespecialize, ...}` for the initialization subproblem —
`AutoDespecialize`-tagged types are already flowing through exactly this
code path today.)

## Versions

Confirmed against `~/.julia/dev/SciMLBase` at `d3a30e5d` (2026-08-23,
v3.49.2, up to date with upstream master), and independently against
registry SciMLBase 3.49.2 / ModelingToolkit 11.39.1 / Julia 1.12.7.
Originally observed on SciMLBase 3.34.0 / ModelingToolkit 11.31.1 /
ModelingToolkitBase 1.51.0 / Julia 1.12.5 (measurement numbers above are
from that run).

ChrisRackauckas:
## Summary

- add the public, unexported `SciMLBase.respecialize` API for converting `ODEFunction` and `ODEInputFunction` values to `AutoSpecialize`, `AutoDespecialize`, or `AutoRespecialize`
- widen initialization and nonlinear-step metadata only during explicit respecialization, while preserving every stored field and avoiding a second `prepare_function` pass
- preserve widened metadata layouts through `unwrapped_f`, including custom `ODEInputFunction` metadata
- document the API, test its inference and layout behavior, and bump the minor version to 3.50.0

## Rationale

The symbolic construction path already widens bounded metadata so separately generated models can share a function type, while direct `AutoSpecialize` construction keeps concrete metadata. Changing all direct constructors would repeat the inference regression that caused #1242 to revert the earlier blanket widening behavior.

This PR therefore makes the tradeoff explicit. Callers that need compilation reuse can use `SciMLBase.respecialize(f, SciMLBase.AutoSpecialize)` instead of manually reconstructing type parameters. The same layout rule applies to `AutoDespecialize` and `AutoRespecialize`, where model-specific metadata would likewise defeat reuse.

## Tests

- `julia --project=. -e 'using Pkg; Pkg.test()'`
- `julia --project=. -e 'using Runic; exit(Runic.main(["--check", "--diff", "."]))'`

Closes #1551.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.49.2"
+version = "3.50.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/docs/src/interfaces/SciMLFunctions.md
+++ b/docs/src/interfaces/SciMLFunctions.md
@@ -102,6 +102,16 @@ which designates how the compiler should specialize on the model function `f`. F
 more details on specialization choices, see the [SciMLProblems](@ref scimlproblems)
 page.
 
+Use `SciMLBase.respecialize(f, specialization)` to rebuild an `ODEFunction` or
+`ODEInputFunction` at an automatic specialization level while retaining its field values.
+For `AutoSpecialize`, `AutoDespecialize`, and `AutoRespecialize`, this also widens
+initialization metadata so model-specific initialization types do not defeat compilation
+reuse.
+
+```@docs
+SciMLBase.respecialize
+```
+
 ### Specifying Jacobian Types
 
 The `jac` field of an inplace style `SciMLFunction` has the signature `jac(J,u,p,t)`,
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -2191,7 +2191,7 @@ export ODEAliasSpecifier, LinearAliasSpecifier
 
 # Core functions
 @public build_solution, build_linear_solution, build_eigenvalue_solution, numargs,
-    unwrap_parameters, invoke_with_despecialized_parameters
+    unwrap_parameters, invoke_with_despecialized_parameters, respecialize
 
 # Solver problem-concretization extension API
 @public get_concrete_p, get_concrete_u0, isconcreteu0, promote_u0,
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -241,6 +241,66 @@ SciMLBase.isinplace(widened)
 end
 
 """
+    respecialize(f, specialize)
+
+Reconstruct an [`ODEFunction`](@ref) or [`ODEInputFunction`](@ref) at the requested
+automatic specialization level while preserving its field values and in-place convention.
+
+`AutoSpecialize`, `AutoDespecialize`, and `AutoRespecialize` widen the initialization
+metadata type (and, for `ODEFunction`, the nonlinear-step metadata type). This matches
+the symbolic-system layout for the first two policies and prevents model-specific
+metadata from defeating the compilation reuse promised by `AutoRespecialize`.
+The direct function constructors retain concrete metadata types because widening can
+reduce inference through code that reads or remakes those fields; `respecialize` makes
+that tradeoff explicit.
+
+For an `ODEInputFunction`, a custom initialization payload that is not
+`OverrideInitData` is widened to `Any`, since the payload type is otherwise unbounded.
+
+# Example
+
+```julia
+f = ODEFunction{false, SciMLBase.FullSpecialize}((u, p, t) -> u)
+f_auto = SciMLBase.respecialize(f, SciMLBase.AutoSpecialize)
+SciMLBase.specialization(f_auto) === SciMLBase.AutoSpecialize
+```
+"""
+@generated function respecialize(
+        f::F, ::Type{S}
+    ) where {
+        F <: Union{ODEFunction, ODEInputFunction},
+        S <: Union{AutoSpecialize, AutoDespecialize, AutoRespecialize},
+    }
+    wrapper = F.name.wrapper
+    field_names = fieldnames(F)
+    length(F.parameters) == length(field_names) + 2 ||
+        error("SciMLFunction type parameters do not match its fields")
+
+    parameter_expressions = Any[F.parameters[1], S]
+    field_expressions = Any[]
+    for (i, field_name) in enumerate(field_names)
+        field_expression = :(getfield(f, $i))
+        push!(field_expressions, field_expression)
+        if field_name === :initialization_data
+            metadata_type = if F <: ODEFunction ||
+                    fieldtype(F, i) <: Union{Nothing, OverrideInitData}
+                Union{Nothing, OverrideInitData}
+            else
+                Any
+            end
+            push!(parameter_expressions, metadata_type)
+        elseif field_name === :nlstep_data
+            push!(parameter_expressions, Union{Nothing, ODENLStepData})
+        else
+            push!(parameter_expressions, Expr(:call, :typeof, field_expression))
+        end
+    end
+
+    new_type = Expr(:curly, wrapper, parameter_expressions...)
+    return Expr(:call, new_type, field_expressions...)
+end
+
+"""
     $(TYPEDSIGNATURES)
 
 A utility function which merges two `NamedTuple`s `a` and `b`, assuming that the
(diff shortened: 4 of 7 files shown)


ChrisRackauckas:
Implementation log for `5e91960c`:

1. Reproduced #1551 and traced the differing layouts to ModelingToolkit's explicit metadata widening versus SciMLBase's concrete direct constructors.
2. Reviewed prior work in SciMLBase #1241, #1242, and #1245, ModelingToolkit #4335 and #4919, and the downstream manual reconstruction in DyadModelOptimizer.
3. Kept direct constructors concrete because blanket widening previously regressed `remake` and `solve` inference.
4. Added the qualified public API `SciMLBase.respecialize` for the three automatic specialization policies. Its direct reconstruction preserves stored values without rerunning extension preparation hooks.
5. Preserved widened layouts through `unwrapped_f`; standard `ODEInputFunction` initialization metadata uses the bounded union, while accepted custom payloads use `Any`.
6. Added layout, field-identity, inference, remake, preparation-hook, custom-payload, and public-interface regressions.

Local verification:

- full `Pkg.test()` passed, including Function Building Error Messages 496/496 and Remake 4430/4430
- repository-wide Runic check passed
- independent final diff review found no actionable issues

ChrisRackauckas:
CI audit complete:

- 53 checks succeeded
- 17 checks failed
- 0 checks remain pending

No failed check is attributable to this PR. Every red shard matches current `master`, the exact base commit, or contemporaneous dependency drift reproduced by an unrelated PR:

- Run Tests: both DownstreamAD failures match the base/master `promote_type_with_nothing` and AutoMooncake failures; all three SymbolicIndexingInterface failures are the same `GradientDescent` import drift reproduced concurrently on unrelated run [32887491788](https://github.com/SciML/SciMLBase.jl/actions/runs/32887491788).
- IntegrationTest: the Catalyst, JumpProcesses, OrdinaryDiffEq InterfaceV, BoundaryValueDiffEq, ModelingToolkit InterfaceI, DiffEqFlux, and SciMLSensitivity Core1/4/5/6/8 failures match the corresponding shards on master run [32777957741](https://github.com/SciML/SciMLBase.jl/actions/runs/32777957741) or exact-base run [32635237444](https://github.com/SciML/SciMLBase.jl/actions/runs/32635237444).
- ReleaseTest: Catalyst's stale `ModelingToolkitBase.SymbolCache` reference matches master run [32777957173](https://github.com/SciML/SciMLBase.jl/actions/runs/32777957173).

The changed regression suite passed locally, the complete local `Pkg.test()` passed, and the repository-wide Runic check passed. No branch change or blind CI rerun is warranted for these baseline failures.

PR #1567, opened 2026-08-30
Handle despecialized parameter cotangents in ODE solution AD

ChrisRackauckas-Claude:
## What changed

ChainRulesCore returns `Tangent{DespecializedParameters}` for the parameter cotangent produced by the symbolic ODESolution pullback. Unwrap this cotangent before calling `remake`, and add a regression test at the pullback boundary. This is intentionally separate from https://github.com/SciML/SciMLBase.jl/pull/1564.

Please ignore this draft until reviewed by @ChrisRackauckas.

## Failing-before / passing-after

Environment: Julia 1.10.12 (commit d93beab124c), ModelingToolkit 11.40.0, ModelingToolkitBase 1.68.1, isolated `JULIA_DEPOT_PATH`.

Before the fix, exact command:
```bash
env JULIA_DEPOT_PATH=/home/crackauc/sandbox/tmp_20260829_113159_30815/SciMLBase-downstreamad-master/.julia_depot TMPDIR=/home/crackauc/tmp /home/crackauc/.juliaup/bin/julia +1.10.12 --project=downstream -e 'include("downstream/remake_autodiff.jl")'
```
failed (exit 1) with:
```text
MethodError: no method matching promote_type_with_nothing(::Type{Float64}, ::ChainRulesCore.Tangent{SciMLBase.DespecializedParameters, @NamedTuple{params::Vector{Float64}}})
Test Summary: | Pass Error Total
`split = false` | 4 1 5
```

After the fix, the same command passed (exit 0):
```text
`split = false` | 7 | 7
`split = true` | None | 0
Consistent gradients | Broken 1
```

## Other validation

- `GROUP=QA julia +1.10.12 --project -e 'using Pkg; Pkg.test()'`: `QA | 121 | 121`, `SciMLBase tests passed`.
- Runic 1.7.0 format check: both changed files clean.
- `typos ext/SciMLBaseChainRulesCoreExt.jl test/downstream/remake_autodiff.jl`: exit 0.
- The broader `GROUP=DownstreamAD` run still reports unrelated existing failures in `observables_autodiff.jl` (Mooncake type assertion and AutoZygote parameter equality); the focused `remake_autodiff.jl` test is green.

🤖 Generated with Codex CLI (model: gpt-5-codex). Agent session: `/home/crackauc/sandbox/tmp_20260829_113159_30815` (local session path).

PR #1581, opened 2026-09-06
Make the integrator plot recipe observed-aware

ChrisRackauckas-Claude:
## What changed and why

`plot(integrator)` was broken for any system that carries symbolic metadata, and it could
not resolve observed equations at all. Reported in
https://github.com/SciML/ModelingToolkit.jl/issues/5006#issuecomment-5491174422: the
integrator recipe is not observed-aware, so a live plot cannot be updated from the
integrator the way it can from a solution.

The recipe is rewritten to follow the solution recipe: symbolic indexes are resolved
through `SymbolicIndexingInterface` instead of by indexing `integrator.u`, so observed
equations and symbolic expressions work; a scalar `idxs` is wrapped the way
`plot(sol; idxs = x)` wraps it; the label string buffer is reset per series; the plot
function in an `idxs` tuple is applied; and `dims` uses the same convention as the
solution recipe so labels come out identical. A stray `@show plott` on the plotting path
is removed.

The value lookup is exposed as `SciMLBase.symbolic_interpolation(integrator, t, idxs)`,
the integrator analogue of `sol(t; idxs)`, together with the predicate
`SciMLBase.has_symbolic_idxs(idxs)`. Both are public and documented so that every solver
package owning an integrator call operator can route `integrator(t; idxs)` without reaching
into a SciMLBase internal or duplicating the predicate. The consumers are
https://github.com/SciML/OrdinaryDiffEq.jl/pull/4469 (OrdinaryDiffEqCore and DelayDiffEq)
and the Sundials and ODEInterfaceDiffEq PRs that follow it. StochasticDiffEq needs no change
of its own: `SDEIntegrator` is an alias for `ODEIntegrator{<:SDEAlgTypes}`.

## Verification

`plot(integrator)` and `integrator(t; idxs)` against the Lorenz system from the issue
thread, with `w ~ x + y + z` as an observed equation.

Before (SciMLBase master):

```
FAIL  plot(integrator)                                BoundsError: attempt to access 3-element Vector{SymbolicUtils...
FAIL  plot(integrator; idxs = x)                      AssertionError: typeof(vars) <: AbstractArray
FAIL  plot(integrator; idxs = w)  # observed          AssertionError: typeof(vars) <: AbstractArray
FAIL  plot(integrator; idxs = [x, w])                 BoundsError: attempt to access 3-element Vector{SymbolicUtils...
FAIL  plot(integrator; idxs = (x, y))                 ArgumentError: invalid index: x(t) of type SymbolicUtils...
FAIL  plot(integrator; denseplot = false, idxs = w)   AssertionError: typeof(vars) <: AbstractArray
```

After (this branch):

```
PASS  plot(integrator)
PASS  plot(integrator; idxs = x)
PASS  plot(integrator; idxs = w)  # observed
PASS  plot(integrator; idxs = [x, w])
PASS  plot(integrator; idxs = (x, y))
PASS  plot(integrator; denseplot = false, idxs = w)
```

The new testsets in `test/downstream/plot_lines.jl` cannot run at all on master — the file
fails to load at `using SciMLBase: symbolic_interpolation` with
`UndefVarError: symbolic_interpolation not defined` — so the failing-before evidence is the
table above, which exercises the same recipe paths the tests assert on.

`test/downstream/plot_lines.jl`, whole file, on this branch:

```
Test Summary:                           | Pass  Total  Time
Plots recipe on multi-state ODESolution |    3      3  5.5s
Test Summary:                                      | Pass  Total   Time
Makie convert_arguments on multi-state ODESolution |    1      1  33.8s
Test Summary:                  | Pass  Total  Time
tspan crops the plotted series |    8      8  1.2s
Test Summary:                                      | Pass  Total  Time
tspan crops solutions integrated backwards in time |    2      2  2.3s
Test Summary:                               | Pass  Total  Time
Integrator recipe without symbolic metadata |    9      9  3.7s
Test Summary:                                 | Pass  Total  Time
Integrator recipe resolves observed equations |   22     22  8.1s
Test Summary:                           | Pass  Total  Time
symbolic_interpolation on an integrator |    7      7  0.3s
```

The recipe now produces the same series labels as the solution recipe for the same `idxs`:

```
idxs            integrator            solution
nothing         ["z", "y", "x"]       ["z", "y", "x"]
[x, w]          ["x", "w"]            ["x", "w"]
(x, w)          ["(x,w)"]             ["(x,w)"]
w               ["w"]                 ["w"]
```

`GROUP=QA`:

```
Test Summary: | Pass  Total     Time
QA            |  123    123  2m49.9s
```

(The first QA attempt on this machine aborted at `using PythonCall` because the local
CondaPkg environment is broken; rerunning with `JULIA_CONDAPKG_BACKEND=Null` and a system
Python gives the run above. Nothing in that failure touches this change.)

Docs build (`julia --project=docs docs/make.jl`, which runs with `doctest = true` and
`linkcheck = true`) exits 0. It caught a real problem first: `is_independent_variable_index`
carried a docstring with no `@docs` entry, which `makedocs` treats as fatal
(`missing_docs`). It is an internal predicate, so it is listed in `Internal_API.md` under
Interpolation rather than promoted alongside `symbolic_interpolation`. The only remaining
output is the pre-existing `size_threshold_warn` warnings on the large interface pages.

Also run: Runic `--check`, clean on every file this PR touches
(`src/ensemble/ensemble_solutions.jl` and `src/solutions/solution_interface.jl` are
unformatted on master and are deliberately left alone), and `typos`.

## Things a reviewer should push back on

- **`getindepsym(integrator::DEIntegrator)` now returns the first independent variable
  rather than the whole vector.** Its own docstring already documents the singular
  contract, and every other method (`getindepsym(::AbstractSciMLProblem)`,
  `getindepsym(::AbstractSciMLSolution)`) returns `syms[1]`. Without this,
  `isequal(idx, getindepsym_defaultt(integrator))` never matches, so naming the
  independent variable symbolically in `idxs` is not recognised as the time axis and the
  x-label renders as a vector. `getindepsym` is `@public`, so this is a visible change even
  though it aligns the method with its documentation.
- **`plot_analytic = true` on an integrator now throws `ArgumentError`.** The previous
  branch referenced `u_n`, `plot_timeseries` and `sol`, none of which exist in SciMLBase,
  so it could only ever have raised `UndefVarError`. Erroring explicitly and pointing at
  `integrator.sol` seemed better than reimplementing a path with no users, but that is a
  judgement call.
- `symbolic_interpolation` and `has_symbolic_idxs` are new public API, so this is a minor
  version bump (3.51.0 → 3.52.0). Both are added to `@public` and to
  `docs/src/interfaces/Init_Solve.md` in this PR.

## Not verified

Downstream packages other than OrdinaryDiffEq, the GPU and Python groups, and the full
`GROUP=Downstream` suite (only `test/downstream/plot_lines.jl` was run). CI covers those.

---

https://claude.ai/code/session_019wKpyZHy9u4QhR4ePiVmam

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLBase"
 uuid = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-version = "3.51.0"
+version = "3.52.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]
--- a/docs/src/interfaces/Init_Solve.md
+++ b/docs/src/interfaces/Init_Solve.md
@@ -93,6 +93,7 @@ SciMLBase.AbstractDAEIntegrator
 SciMLBase.AbstractSDDEIntegrator
 SciMLBase.DECache
 SciMLBase.step!
+SciMLBase.symbolic_interpolation
 Base.resize!(::SciMLBase.DEIntegrator, ::Int)
 Base.deleteat!(::SciMLBase.DEIntegrator, ::Any)
 SciMLBase.addat!
--- a/docs/src/interfaces/Internal_API.md
+++ b/docs/src/interfaces/Internal_API.md
@@ -52,6 +52,7 @@ SciMLBase.linear_interpolant
 SciMLBase.linear_interpolant!
 SciMLBase.hermite_interpolant
 SciMLBase.hermite_interpolant!
+SciMLBase.is_independent_variable_index
 ```
 
 ## Symbolic Save Selection
--- a/src/SciMLBase.jl
+++ b/src/SciMLBase.jl
@@ -2325,7 +2325,7 @@ export cache_operator, concretize, has_adjoint, has_concretization, has_exp, has
 
 # Interpolation / symbolic / solution interface
 @public interp_summary, getindepsym, getindepsym_defaultt,
-    calculate_solution_errors!, initialize_dae!
+    calculate_solution_errors!, initialize_dae!, symbolic_interpolation
 @public get_saved_subsystem, SavedSubsystem, get_saved_state_idxs,
     SavedSubsystemWithFallback, get_save_idxs_and_saved_subsystem,
     create_parameter_timeseries_collection, get_saveable_values, save_discretes!,
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -785,7 +785,7 @@ function getindepsym(integrator::DEIntegrator)
     if isempty(syms)
         return nothing
     end
-    return syms
+    return syms[1]
 end
 
 function getparamsyms(integrator::DEIntegrator)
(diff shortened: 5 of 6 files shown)


ChrisRackauckas-Claude:
The two red `DownstreamAD` checks here are a pre-existing master failure, not something this PR introduces, and they have a fix in https://github.com/SciML/SciMLBase.jl/pull/1582.

`test/downstream/remake_autodiff.jl` has used `ChainRulesCore` since #1567 without it being added to `test/downstream/Project.toml`, so the group aborts on its first testset:

```
Autodiff Remake: Error During Test
  LoadError: ArgumentError: Package ChainRulesCore not found in current path.
```

The same failure appears on master's own run at `2f74543d`, before this branch existed: https://github.com/SciML/SciMLBase.jl/actions/runs/34027316086. `DownstreamAD` is `continue_on_error = true` in `test/test_groups.toml`, which is why master's workflow still reports green overall.

With #1582 applied the whole group passes locally, so once that merges these two checks should go green here on a re-run.

https://claude.ai/code/session_019wKpyZHy9u4QhR4ePiVmam

PR #1613, opened 2026-09-23
Fix QA: avoid qualified access to non-public Base.Callable

ChrisRackauckas-Claude:

## What changed and why

`GROUP=QA` has been red on master since PR #1603 (commit c13c04c7, merged 2026-09-14): `NonlinearProblem(::SteadyStateProblem)` uses `Base.Callable` to check whether a stored `lowered_problem` should be invoked, but `Base.Callable` has never been exported or declared `public` by `Base`. ExplicitImports' `all_qualified_accesses_are_public` check correctly flags this. `Base.Callable` is simply `Union{Function, Type}`, so this replaces the qualified access with that literal union instead of adding an ignore-list entry for an internal name that has a trivial public equivalent.

## Bisect

- First red on master: run https://github.com/SciML/SciMLBase.jl/actions/runs/34869948784 (commit 969bdea, right after c13c04c7/#1603 merged 2026-09-14). QA (julia 1) failed there; SymbolicIndexingInterface passed in that same run (that's a separate, unrelated failure — see below).
- Still red on the latest master run: https://github.com/SciML/SciMLBase.jl/actions/runs/35839187919
- Root: https://github.com/SciML/SciMLBase.jl/commit/c13c04c7ce36488579a56e9b2f88a64dd472ba4b (PR #1603)

## Verification

Confirmed `Base.ispublic(Base, :Callable) == false` on Julia 1.12, and `Base.Callable === Union{Function, Type}`.

The QA group can't run to completion on this machine (PythonCall's bundled Python fails to start locally: `no codec search functions registered`), so the specific ExplicitImports check was reproduced standalone instead:

Before the fix (unmodified master, `git stash` to revert):
```
ERROR: LoadError: NonPublicQualifiedAccessException
Module `SciMLBase` has explicit imports of names from modules in which they are not public (i.e. exported or declared public in Julia 1.11+):
- `Callable` is not public in `Base` but it was imported from `Base` at `.../src/problems/nonlinear_problems.jl:256:24`
```

After the fix:
```
RESULT: nothing
```
(`ExplicitImports.check_all_qualified_accesses_are_public` returns `nothing` on success — matches the ignore list used in `test/qa/qa.jl`.)

`GROUP=Core` passes locally in full (including `Problem building tests | 170 170`), confirming no behavior change:
```
Test Summary:          | Pass  Total   Time
Problem building tests |  170    170  16.1s
```

Runic `--check` and `typos` are both clean on the changed file.

## What I did not verify

Did not run `GROUP=QA` end-to-end locally (PythonCall/CondaPkg failure unrelated to this change, present on this machine for all branches). Relying on the standalone ExplicitImports reproduction above plus CI.

## Related

A second, unrelated failure (`GROUP=SymbolicIndexingInterface`) is red in the same CI run; that is fixed in a separate PR since it has a different root cause (a latent bug in the generic `NonlinearProblem(::AbstractNonlinearProblem)` fallback, exposed by an unrelated upstream ModelingToolkit.jl change).

https://claude.ai/code/session_01NB3oFTNzGW79UtDt8AyjsM


ChrisRackauckas-Claude:
Closing as a duplicate of https://github.com/SciML/SciMLBase.jl/pull/1608, which fixes the same line using SciMLBase's own `Callable` alias (consistent with `discrete_problems.jl`) and already has QA green. Reviewed: the two are `===`-identical in behaviour.

PR #1614, opened 2026-09-23
Raise a clear ArgumentError converting an SCCNonlinearProblem to NonlinearProblem

ChrisRackauckas-Claude:

## What changed and why

`GROUP=SymbolicIndexingInterface` is (or was) red on master (CI run https://github.com/SciML/SciMLBase.jl/actions/runs/35839187919): `init(ssprob, DynamicSS(Tsit5()))` on an MTK-lowered SCC steady-state problem throws `FieldError: type SCCNonlinearProblem has no field u0`.

Root cause: `NonlinearProblem(prob::AbstractNonlinearProblem) = NonlinearProblem{isinplace(prob)}(prob.f, prob.u0, prob.p)` assumes every `AbstractNonlinearProblem` subtype has a `u0` field. `SCCNonlinearProblem` never has one (its state is the concatenation of its sub-problems' states), so calling `NonlinearProblem(x)` on an `SCCNonlinearProblem` has always thrown this `FieldError`, standalone, with no other package involved:

```julia
using SciMLBase
f1 = NonlinearFunction((du, u, p) -> (du .= u .^ 2 .- p))
prob1 = NonlinearProblem(f1, [1.0], 4.0)
sccprob = SciMLBase.SCCNonlinearProblem((prob1,), (Returns(nothing),))
NonlinearProblem(sccprob)   # FieldError on master
```

## Revision history on this PR

An earlier version of this PR made `NonlinearProblem(prob::SCCNonlinearProblem)` an identity (`= prob`), matching how a stored `NonlinearProblem`/`LinearProblem` `lowered_problem` is already handled verbatim elsewhere (`"stored problem is used verbatim"` in `test/problem_building_test.jl`). Review caught two problems with that:

1. It doesn't fix the reported CI failure at all — the real cause is `SteadyStateDiffEq.jl` having no `__init` for `DynamicSS`/`SICNM`, which is fixed independently and sufficiently in **https://github.com/SciML/SteadyStateDiffEq.jl/pull/173** (confirmed: that fix alone, against today's *unmodified* SciMLBase, resolves the reported failure).
2. Worse, the identity is itself a footgun and a hazard: a `NonlinearProblem` constructor returning something that is not a `NonlinearProblem` breaks any caller that assumes `.u0`/`.lb`/`.ub`/etc., and it is exactly what let `NonlinearSolve.jl`'s default-algorithm converter (`src/default.jl`, which does `nlprob = NonlinearProblem(prob); return __init(nlprob, nothing, args...)`) recurse forever instead of terminating, once it received `nlprob === prob`. Reproduced:
   ```
   Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
   ERROR: LoadError: StackOverflowError:
    [1] __init(prob::SCCNonlinearProblem{...}, ::Nothing, args::DynamicSS{...}; ...) (repeats 39992 times)
   ```

## What this PR does now

`SCCNonlinearProblem` solves as an ordered sequence of block solves (see `SciMLBase.solve(::SCCNonlinearProblem, ::Union{DynamicSS,SICNM}, ...)` and `SSRootfind`'s handling in `SteadyStateDiffEq.jl`), not a single residual — there is no faithful `NonlinearProblem` to construct from it. So `NonlinearProblem(prob::SCCNonlinearProblem)` now raises a clear `ArgumentError`:

```
ArgumentError: an SCCNonlinearProblem cannot be converted to a NonlinearProblem; solve it directly or use its component problems.
```

An error cannot recurse, so this also removes the recursion hazard above without needing a companion guard in `NonlinearSolve.jl` (the earlier companion PR, https://github.com/SciML/NonlinearSolve.jl/pull/1327, has been closed as unnecessary: `NonlinearProblem(prob)` never returns `prob` itself now, so the `nlprob === prob` case it guarded against cannot occur).

## Verification

Standalone reproducer, fail-before/pass-after:

Before (unmodified master):
```
Caught: FieldError -- FieldError: type SCCNonlinearProblem has no field `u0`, available fields: `probs`, `explicitfuns!`, `f`, `p`, `parameters_alias`
```

After (this PR):
```
Caught: ArgumentError -- ArgumentError: an SCCNonlinearProblem cannot be converted to a NonlinearProblem; solve it directly or use its component problems.
```

Regression test updated to `@test_throws ArgumentError NonlinearProblem(sccprob)` in `test/problem_building_test.jl` (still discriminates: throws `FieldError` on unmodified master, so `@test_throws ArgumentError` fails there and passes here).

`GROUP=Core` passes locally in full, including the updated test (`Problem building tests | 182 182`):
```
```

Runic `--check` and `typos` are clean on both changed files.

## What I did not verify

- Did not run the actual downstream `GROUP=SymbolicIndexingInterface` test in isolation against this PR — see https://github.com/SciML/SteadyStateDiffEq.jl/pull/173's body for why that test doesn't need this PR at all (it is fixed independently there).
- Did not run `GROUP=QA` locally (PythonCall/CondaPkg fails to start on this machine, unrelated to this change).
- Did not audit every existing caller of `NonlinearProblem` on an `AbstractSteadyStateProblem`-typed value for reliance on the old identity behavior; grepped `SteadyStateDiffEq.jl`'s source and test suite (including its new tests in PR #173) and found none that call `NonlinearProblem` directly on an already-built `SCCNonlinearProblem` — all existing callers only convert a `SteadyStateProblem` (which correctly returns the stored `SCCNonlinearProblem` verbatim; that path is untouched by this change).

## Links

- Original CI failure: https://github.com/SciML/SciMLBase.jl/actions/runs/35839187919
- Real fix for the reported failure (independent of this PR): https://github.com/SciML/SteadyStateDiffEq.jl/pull/173
- Closed companion (no longer needed): https://github.com/SciML/NonlinearSolve.jl/pull/1327

https://claude.ai/code/session_01NB3oFTNzGW79UtDt8AyjsM


SebastianM-C:
https://github.com/SciML/NonlinearSolve.jl/pull/1319 is also related to the same error

